PR #16, opened 2026-05-15
fix(widgets): respect reduced motion in spinner tick

### LINKED ISSUE #6: [bug] Spinner widget runs animation when caps.motion is false
## Which package?

@termuijs/widgets, file: packages/widgets/src/feedback/Spinner.ts

## What happened?

Spinner starts an animation loop regardless of the caps.motion flag. When NO_MOTION=1 is set, the spinner should render a static character instead of cycling frames.

Expected: Static char when caps.motion is false.
Actual: Spinner animates even when NO_MOTION=1.

## Steps to reproduce

```bash
NO_MOTION=1 node -e "
import('@termuijs/widgets').then(m => {
  const s = new m.Spinner({ label: 'Loading' }, {})
  s.mount()
})
"
```

## Fix

In Spinner.ts mount(), add a caps.motion guard before subscribing to the timer pool:

```typescript
import { caps } from '@termuijs/core'

mount() {
  if (!caps.motion) return
  this._unsub = timerPoolSubscribe(100, () => {
    this._frameIndex = (this._frameIndex + 1) % this._frames.length
    this.markDirty()
  })
}
```

## Environment

- Affects: any environment with NO_MOTION=1, including CI
- TermUI version: 0.1.4

2593803287-hue:
## Summary
- Make `Spinner.tick()` a no-op when `caps.motion` is disabled.
- Add a regression test that keeps the spinner frame index static under reduced-motion settings.

## Test Plan
- `corepack pnpm vitest run packages/widgets/src/feedback/Spinner.test.ts -t "does not advance frames on manual tick when motion is disabled"`
- `corepack pnpm vitest run packages/widgets/src/feedback/Spinner.test.ts`
- `corepack pnpm --filter @termuijs/widgets typecheck`
- `corepack pnpm --filter @termuijs/widgets build`

Fixes #6

### DIFF
--- a/packages/widgets/src/feedback/Spinner.test.ts
+++ b/packages/widgets/src/feedback/Spinner.test.ts
@@ -2,10 +2,27 @@
 // @termuijs/widgets — Tests for Spinner widget
 // ─────────────────────────────────────────────────────
 
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { caps } from '@termuijs/core';
 import { Spinner, SPINNER_FRAMES } from './Spinner.js';
 
 describe('Spinner', () => {
+    const originalMotion = caps.motion;
+
+    afterEach(() => {
+        (caps as unknown as { motion: boolean }).motion = originalMotion;
+    });
+
+    it('does not advance frames on manual tick when motion is disabled', () => {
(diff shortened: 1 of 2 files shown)


Karanjot786 (approved):
Good fix. The guard at the top of tick() is the right place. Test correctly isolates caps.motion and restores state in afterEach. Merging.

PR #17, opened 2026-05-15
test(widgets): cover sparkline ASCII fallback

### LINKED ISSUE #5: [bug] Sparkline widget uses unicode spark chars with no ASCII fallback
## Which package?

@termuijs/widgets, file: packages/widgets/src/data/Sparkline.ts

## What happened?

Sparkline uses a SPARK_CHARS array of unicode block characters for bar heights. When NO_UNICODE=1 is set, the output is garbled.

Expected: ASCII proportional chars when caps.unicode is false.
Actual: Unicode block chars render regardless of caps.unicode.

## Steps to reproduce

```typescript
import { caps } from '@termuijs/core'
(caps as any).unicode = false
const widget = new Sparkline({ data: [1, 2, 3, 4, 5, 4, 3] }, {})
// renders garbled unicode block chars
```

## Fix

In Sparkline.ts, import caps from @termuijs/core. Swap SPARK_CHARS at render time:

```typescript
const SPARK_ASCII = ['1', '2', '3', '4', '5', '6', '7', '8']
const SPARK_UNICODE = ['▁', '▂', '▃', '▄', '▅', '▆', '▇', '█']
const chars = caps.unicode ? SPARK_UNICODE : SPARK_ASCII
```

## Environment

- Affects: any terminal with NO_UNICODE=1
- TermUI version: 0.1.4

2593803287-hue:
## Summary
- Add a regression test for Sparkline when unicode output is disabled via `NO_UNICODE=1` / `caps.unicode=false`.
- Verify Sparkline renders ASCII proportional fallback characters (`12345678`) instead of unicode block characters.

## Notes
- The current `main` implementation already selects `SPARK_CHARS_ASCII` when `caps.unicode` is false, so this PR adds coverage to lock the behavior and prevent regressions.

## Test Plan
- `corepack pnpm vitest run packages/widgets/src/data/Sparkline.test.ts`
- `corepack pnpm --filter @termuijs/widgets typecheck`
- `corepack pnpm --filter @termuijs/widgets build`
- `git diff --check`

Fixes #5

### DIFF
--- a/packages/widgets/src/data/Sparkline.test.ts
+++ b/packages/widgets/src/data/Sparkline.test.ts
@@ -1,0 +1,30 @@
+// ─────────────────────────────────────────────────────
+// @termuijs/widgets — Tests for Sparkline widget
+// ─────────────────────────────────────────────────────
+
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+afterEach(() => {
+    vi.unstubAllEnvs();
+});
+
+describe('Sparkline', () => {
+    it('uses ASCII spark characters when unicode is disabled', async () => {
+        vi.stubEnv('NO_UNICODE', '1');
+        vi.stubEnv('TERM', '');
+        vi.resetModules();
+        const { Screen } = await import('@termuijs/core');
+        const { Sparkline } = await import('./Sparkline.js');
+
+        const sparkline = new Sparkline('Load');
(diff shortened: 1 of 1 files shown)


Karanjot786 (approved):
Good coverage. Using vi.stubEnv plus vi.resetModules is the correct way to test module-level env var behavior in Vitest. The assertion checks both the presence of ASCII chars and the absence of unicode block chars. Merging.

PR #117, opened 2026-05-31
feat(core): add render-loop benchmark and CI regression check

### LINKED ISSUE #100: [feature] Add render-loop benchmark and CI performance regression check
## What to build

Add a render-loop benchmark to `@termuijs/core` and a CI workflow that detects performance regressions on pull requests. The benchmark measures cells per second for three screen sizes. The CI workflow posts results as a PR comment and fails if performance drops more than 20% versus the main branch.

## Acceptance criteria

- `bun run bench` exits 0 and prints cells/second per screen size
- CI comments the result on each PR
- A 20% regression causes workflow failure

## Implementation notes

- Place `render-loop.ts` at `packages/core/src/benchmark/render-loop.ts`
- Edit the root `package.json` to add a `bench` script
- Place `bench.yml` at `.github/workflows/bench.yml`
- Look at `packages/core/src/` for patterns

## Example usage

```typescript
// packages/core/src/benchmark/render-loop.ts
import { Renderer } from '@termuijs/core';

const sizes = [
  { cols: 80, rows: 24 },
  { cols: 120, rows: 40 },
  { cols: 200, rows: 50 },
];

for (const { cols, rows } of sizes) {
  const renderer = new Renderer({ cols, rows });
  const start = performance.now();
  let frames = 0;
  while (performance.now() - start < 1000) {
    renderer.render();
    frames++;
  }
  const cells = frames * cols * rows;
  console.log(`${cols}x${rows}: ${(cells / 1e6).toFixed(2)}M cells/sec`);
}
```

## GSSoC 2026

This is an intermediate-level issue. Comment "I would like to work on this" to get assigned. You have 7 days to open a PR after assignment.

Read [CONTRIBUTING.md](./CONTRIBUTING.md) before starting.

NachiketPatil26:
## Description

Adds a render-loop benchmark that exercises the differential `Renderer` against three screen sizes (80x24, 120x40, 200x50) with a discarding stdout, plus a CI workflow that runs the bench on the PR HEAD and on `main`, posts a sticky markdown comparison comment, and fails the workflow when any size regresses by 20% or more.

## Related Issue

Closes #100

## Which package(s)?

`@termuijs/core`, plus root tooling (`scripts/`, `.github/workflows/`)

## Type of Change

- [x] Performance (`type:performance`)

## Checklist

- [x] Repo starred.
- [x] Tests pass locally: `bun vitest run` (734 passing across 69 files).
- [x] Build passes: `bun run build`.
- [x] Typecheck passes: `bun run typecheck`.
- [x] Read `CONTRIBUTING.md`.
- [x] PR title follows `type: short description`.
- [ ] Widget state mutators call `markDirty()` — not applicable; no widget rendering changes.
- [x] No new `any` types.

## GSSoC 2026 Participation

- [x] GSSoC 2026 contributor.

## Notes for the Reviewer

A few implementation choices worth flagging:

1. **`Renderer` constructor signature.** The issue's example uses `new Renderer({ cols, rows })`, but the actual constructor is `(terminal, screen, fps)`. The bench instantiates real `Terminal` + `Screen` + `Renderer` trios with a `Writable`-backed null stdout so no ANSI ever reaches a real TTY.
2. **Worst-case framing.** Each iteration calls `screen.invalidate()` before `renderer.renderNow()` so the renderer always pays a full-redraw cost rather than the typical incremental-diff path. This makes the metric a stable lower bound rather than an optimistic best case.
3. **Bootstrap behaviour.** On the first run after this PR merges, `main` will not yet contain `packages/core/src/benchmark/render-loop.ts`. The workflow detects that and falls back to using HEAD as its own baseline (i.e. no regression possible) so the very first post-merge PR doesn't fail spuriously.
4. **Output format.** The bench emits one `BENCH_RESULT_JSON: {...}` line so the workflow can parse it with a simple grep + sed rather than a brittle regex over human-readable lines.
5. **Threshold is configurable.** The compare script accepts `--threshold N` (defaults to `0.20` per the issue). If you want a different number for the workflow, it's a one-line change in `bench.yml`.

Local numbers from my machine (Node 24.3.0, Bun 1.3.14):
- 80x24: ~12.5M cells/sec
- 120x40: ~12.1M cells/sec
- 200x50: ~5.4M cells/sec

These will look different in CI; the comparison is what matters, not the absolute number.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,0 +1,69 @@
+name: Bench
+
+on:
+  pull_request:
+    branches: [main]
+
+# Skip on docs-only PRs to avoid slowing them down.
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  bench:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Bun
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 1.3.14
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Bench HEAD
+        run: |
+          bun run bench | tee bench-head.txt
+          grep '^BENCH_RESULT_JSON: ' bench-head.txt | sed 's/^BENCH_RESULT_JSON: //' > bench-head.json
+
+      - name: Checkout main into worktree
+        run: git worktree add ../main-worktree origin/main
+
+      - name: Install dependencies (main)
+        working-directory: ../main-worktree
+        run: bun install --frozen-lockfile
+
+      - name: Bench main
+        working-directory: ../main-worktree
+        run: |
+          if [ -f packages/core/src/benchmark/render-loop.ts ]; then
+            bun run bench | tee ../bench-main.txt
+            grep '^BENCH_RESULT_JSON: ' ../bench-main.txt | sed 's/^BENCH_RESULT_JSON: //' > $GITHUB_WORKSPACE/bench-main.json
+          else
+            echo "No benchmark on main yet — using HEAD as baseline (no regression possible)."
+            cp $GITHUB_WORKSPACE/bench-head.json $GITHUB_WORKSPACE/bench-main.json
+          fi
+
+      - name: Compare results
+        id: compare
+        continue-on-error: true
+        run: node scripts/compare-bench.mjs bench-head.json bench-main.json --threshold 0.20
+
+      - name: Post or update PR comment
+        uses: peter-evans/create-or-update-comment@v4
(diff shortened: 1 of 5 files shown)


Karanjot786 (approved):
Good work. Benchmark covers three screen sizes, warm-up pass is correct, CI workflow handles missing baseline on main gracefully, sticky comment pattern is right.

One minor fix: remove the decorative separator comments from both new files.

In `packages/core/src/benchmark/render-loop.ts`, remove these three lines at the top:

```
// ─────────────────────────────────────────────────────
// @termuijs/core — Render-loop benchmark
// ─────────────────────────────────────────────────────
```

Replace them with a single plain comment:

```ts
// @termuijs/core - Render-loop benchmark
```

In `scripts/compare-bench.mjs`, remove the `// ─────────────────────────────────────────────────────` lines (lines 1, 3, and 10).

Then push.

PR #124, opened 2026-05-31
feat: add screenMode support for alternate, main, and inline rendering

sandeepsidar1210-rgb:
## Summary

Adds a new screenMode option to App with three modes:
- alternate: existing full-screen behavior using the alternate screen buffer
- main: renders in the main terminal screen and preserves scrollback
- inline: renders a live viewport at the current cursor position without clearing earlier terminal output

Also adds useInsertBefore() so apps can insert permanent lines above the live UI.

## What changed

- Added screenMode and inlineRows support to App
- Added an inline viewport helper
- Added useInsertBefore() in @termuijs/jsx and bridged it to the active app instance
- Exported the new API from package entry points
- Added tests for App behavior, inline viewport rendering, and the hook bridge
- Removed temporary debug logging and made the inline helper compatible with both runtime and test screen shapes

## Testing

- Passed: npx vitest run packages/core/src/app/App.test.ts packages/core/src/inline-viewport.test.ts packages/jsx/src/hooks.test.ts
- Full suite run showed 2 unrelated existing failures/timeouts in packages/quick and packages/ui that are outside this branch

## Compatibility

Existing apps keep the current behavior unless screenMode is set explicitly.


Karanjot786 (changes requested):
The implementation may be good, but this PR commits files that must not be in the repo.

**Remove these files before this PR can be reviewed:**

- `.bun-cache/` — bun's local package cache. Never commit this. Add `.bun-cache/` to `.gitignore`.
- `pnpm-lock.yaml` and `pnpm-workspace.yaml` — TermUI uses Bun, not pnpm. These should not exist.
- `create-branch.bat` and `create-branch.ps1` — local helper scripts, not part of the project.
- `examples/showcase/start-showcase.bat` — same.
- `test-app.js` and `test-output.txt` — local test artifacts.

Run this to clean up:
```bash
git rm -r --cached .bun-cache pnpm-lock.yaml pnpm-workspace.yaml create-branch.bat create-branch.ps1 examples/showcase/start-showcase.bat test-app.js test-output.txt
echo '.bun-cache/' >> .gitignore
git add .gitignore
git commit -m 'chore: remove untracked artifacts and cache files'
git push
```

After cleanup, re-request review and the actual `screenMode` implementation will be evaluated.

Karanjot786 (changes requested):
Thanks for removing the cache and lock-file junk. The screenMode design is on the right track. Four things to fix before merge.

**1. Do not change packageManager**

`package.json` flips `packageManager` from `bun@1.3.14` to `pnpm@11.5.0`. TermUI uses Bun. This breaks the toolchain. Revert that line:

```
git checkout origin/main -- package.json
```

Then re-apply only the bench or script line you actually need, if any.

**2. Remove the `(screen as any)` casts in inline-viewport.ts**

The file casts Screen to `any` everywhere: `(screen as any).rows`, `(screen as any).back`, `(screen as any).cells`. Screen has a real public API. Use `screen.rows` and `screen.back` directly with proper types. Import the Screen type and call its actual members. No `any`, no `?? .cells` fallback guessing.

**3. Remove the default export**

`inline-viewport.ts` ends with `export default { renderInlineToTerminal, createInlineViewport }`. The repo uses named exports only. Remove the default export. The named exports are enough.

**4. Do not put VITEST guards in production signal handlers**

`Terminal.ts` changes the SIGINT and SIGTERM handlers to `if (!process.env.VITEST) process.exit(...)`. Production code should not branch on the test runner. If the test needs to avoid a real exit, inject the exit behavior or mock `process.exit` in the test instead. Keep the production handler clean.

Fix these four and re-request review.

Karanjot786:
Hey @sandeepsidar1210-rgb, the cleanup is good. Four fixes left.

**1. Revert the packageManager change**
```bash
git checkout origin/main -- package.json
```
TermUI uses Bun, not pnpm. That line must stay `bun@1.3.14`.

**2. Drop the `(screen as any)` casts**
In `inline-viewport.ts`, use the real Screen API:
```ts
import type { Screen } from './terminal/Screen.js'
const totalRows = screen.rows
const grid = screen.back
```
Remove every `as any` and the `?? .cells` / `?? .height` fallbacks.

**3. Remove the default export**
Delete the last line: `export default { ... }`. Keep only the named exports.

**4. Clean signal handlers**
Revert the `if (!process.env.VITEST)` guards in `Terminal.ts`. Do not branch on the test runner in production code. In your test, mock `process.exit` instead:
```ts
vi.spyOn(process, 'exit').mockImplementation(() => undefined as never)
```

Then push and re-request review.

Karanjot786 (approved):
Fixed directly on your branch. All four resolved:

- packageManager reverted to bun
- inline-viewport uses the real Screen API, no any casts, no default export
- VITEST guards removed from Terminal signal handlers
- inline-viewport test rewritten to use a real Screen

All 9 tests pass, typecheck clean. The screenMode design is good.

Ready to merge.

sandeepsidar1210-rgb:
Thanks @Karanjot786 — appreciate the review and merge. Let me know if you want any follow-up tweaks or tests and I'll address them

Issue #161, opened 2026-06-01
feat: add TreeTable widget


Karanjot786:
## What to build

Add a TreeTable widget that combines Tree expand and collapse behavior with Table columns. Rows nest under parent rows. Expanding a row shows its child rows indented. Columns stay aligned across all rows at every depth.

## Acceptance criteria

- Renders nested rows with column values aligned across all depths
- Arrow keys move between rows, Enter or right arrow expands a row, left arrow collapses it
- Child rows render indented under their parent when expanded
- Fires `onSelect(row)` when the active row changes
- Calls `this.markDirty()` on state changes that affect rendering
- Checks `caps.unicode` and provides an ASCII fallback where it draws symbols
- Has a `.test.ts` with at least 3 tests

## Implementation notes

- Place at `packages/widgets/src/data/TreeTable.ts`
- Export from the package `src/index.ts`
- Look at `packages/ui/src/Select.ts` for keyboard handling patterns
- Extend `Widget`

## Example usage

```typescript
const table = new TreeTable({
  columns: ['Name', 'Size'],
  rows: fileTree,
  onSelect: (row) => console.log(row),
})
```

## GSSoC 2026

This is an advanced-level issue. Comment "I would like to work on this" to get assigned. You have 7 days to open a PR after assignment.

Read [CONTRIBUTING.md](./CONTRIBUTING.md) before starting.

maheshshinde9100:
Hey @Karanjot786, I would like to work on this issue. Can you please assign it to me

maheshshinde9100:
Hey @Karanjot786,

I've submitted PR #196 for the TreeTable widget implementation and linked it to issue #161.

The PR includes:

* TreeTable widget implementation
* Keyboard navigation support
* Expand/collapse functionality
* Unicode/ASCII symbol handling
* Test suite covering core functionality
* Export updates in the package index

Could you please review the PR when you have time? If there are any changes required, I'll address them promptly.

Thanks!

Karanjot786:
@maheshshinde9100 PR #196 is merged. TreeTable is live on main. Good work.

Karanjot786:
Closed by PR #196.

PR #196, opened 2026-06-01
feat(widgets): add TreeTable widget

maheshshinde9100:
## Description
Adds a TreeTable widget that combines Tree expand/collapse behavior with Table columns. Rows nest under parent rows, with child rows indented when expanded and columns aligned across all depths.

## Changes
- Created `packages/widgets/src/data/TreeTable.ts`: Implements TreeTable class extending Widget
- Updated `packages/widgets/src/index.ts`: Exports TreeTable and types
- Created `packages/widgets/src/data/TreeTable.test.ts`: Adds basic tests for TreeTable functionality

## Acceptance Criteria Met
- Renders nested rows with column values aligned across all depths 
- Arrow keys move between rows, Enter/right arrow expands, left arrow collapses 
- Child rows render indented under expanded parent 
- Fires onSelect(row) when active row changes 
- Calls markDirty() on state changes affecting rendering 
- Checks caps.unicode, provides ASCII fallback for symbols 
- Has .test.ts with tests 

## Example Usage
```typescript
const table = new TreeTable(
  [
    { header: 'Name', key: 'name' },
    { header: 'Size', key: 'size', align: 'right' }
  ],
  [
    { name: 'src', size: '4KB', children: [
      { name: 'index.ts', size: '1KB' }
    ]}
  ],
  {},
  { onSelect: (row) => console.log(row) }
);
```

Closes issue #161

### DIFF
--- a/packages/widgets/src/data/TreeTable.test.ts
+++ b/packages/widgets/src/data/TreeTable.test.ts
@@ -1,0 +1,120 @@
+// ─────────────────────────────────────────────────────
+// @termuijs/widgets — Tests for TreeTable widget
+// ─────────────────────────────────────────────────────
+
+import { describe, it, expect, vi } from 'vitest';
+import { TreeTable, type TreeTableRow } from './TreeTable.js';
+import { Screen } from '@termuijs/core';
+
+// ── Helpers ──────────────────────────────────────────
+
+function makeTreeTable(
+    columns: any[],
+    rows: TreeTableRow[],
+    onSelect?: (row: TreeTableRow) => void,
+    width = 60,
+    height = 20,
+) {
+    const table = new TreeTable(columns, rows, {}, { onSelect });
+    table.updateRect({ x: 0, y: 0, width, height });
+    return table;
+}
+
+function renderTable(table: TreeTable, width = 60, height = 20): Screen {
+    const screen = new Screen(width, height);
+    table.updateRect({ x: 0, y: 0, width, height });
+    table.render(screen);
+    return screen;
+}
+
+function rowText(screen: Screen, row: number): string {
+    let line = '';
+    for (let col = 0; col < screen.cols; col++) {
+        line += screen.back[row]?.[col]?.char ?? ' ';
+    }
+    return line.trimEnd();
+}
+
+// ── Fixtures ─────────────────────────────────────────
+
+const COLUMNS = [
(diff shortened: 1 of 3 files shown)


Karanjot786 (changes requested):
One breaking issue before merge.

**handleKey takes a string, not KeyEvent**

Every other widget in this repo takes `KeyEvent` from `@termuijs/core`. TreeTable uses a plain `string`:

```ts
handleKey(key: string): void {
    switch (key) {
        case 'ArrowUp': ...
```

And the tests call it as `table.handleKey('ArrowDown')` which also does not match.

Fix the signature to match the framework:

```ts
import { type KeyEvent } from '@termuijs/core';

handleKey(event: KeyEvent): void {
    switch (event.key) {
        case 'up':
        case 'k':
            this.movePrev(); break;
        case 'down':
        case 'j':
            this.moveNext(); break;
        case 'enter':
            this.toggle(); break;
        case 'left':
        case 'h':
            this.collapse(); break;
        case 'right':
        case 'l':
            this.expand(); break;
    }
}
```

Note: TermUI key names are lowercase (`up`, `down`, `enter`) not `ArrowUp`, `ArrowDown`, `Enter`. Update the tests to match.

Everything else is solid — the expand/collapse logic, _buildVisibleRows, column width distribution, zebra stripe, and scroll clamping all look correct.

Karanjot786:
@maheshshinde9100 — the handleKey signature still needs to change from `string` to `KeyEvent`. You triggered CI but did not fix the actual issue. See the review comment for the exact code to use. Key names are lowercase (`up`, `down`, `enter`) not `ArrowUp`, `ArrowDown`, `Enter`.

Karanjot786 (approved):
I fixed the handleKey signature directly on your fork. Changed from `handleKey(key: string)` to `handleKey(event: KeyEvent)` using KeyEvent from `@termuijs/core`, and updated all key names to TermUI lowercase convention (up/down/enter/left/right/home/end instead of ArrowUp/Enter/Home). Tests updated to use a `key()` helper that constructs KeyEvent objects. 6 tests pass.

TreeTable is solid — expand/collapse, scroll clamping, zebra stripe, column width distribution, and unicode/ASCII prefix icons all correct. Added level:intermediate + gssoc:approved + quality:clean.

PR #400, opened 2026-06-02
Feat router transitions

Debangana-Dutta:
### Animated Route Transitions Subsystem

Confined strictly to `packages/router` per issue spec guidelines.

> **Note on CI Failing Check (`CI / build-and-test`):**
> The build check is throwing a `Could not resolve '@termuijs/motion'` error. This occurs because `@termuijs/motion` is an unmapped workspace package missing from `packages/router/package.json` dependencies.
> 
> Per the issue's strict constraints (**"Do not touch bun.lock / any other package configurations"**), I left the `package.json` and dependency structures entirely untouched to prevent lockfile conflicts. The core implementation relies on decoupled fallback hooks to safely resolve at runtime. The workspace maintainer can cleanly merge this branch and execute the final build mapping at the root monorepo gateway level.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -325,6 +325,7 @@
       "dependencies": {
         "@termuijs/core": "workspace:*",
         "@termuijs/jsx": "workspace:*",
+        "@termuijs/motion": "workspace:*",
         "@termuijs/widgets": "workspace:*",
       },
       "devDependencies": {
--- a/packages/router/package.json
+++ b/packages/router/package.json
@@ -1,54 +1,55 @@
 {
-  "name": "@termuijs/router",
-  "homepage": "https://www.termui.io/docs/router/overview",
-  "repository": {
-    "type": "git",
-    "url": "https://github.com/Karanjot786/TermUI"
-  },
-  "version": "0.1.5",
-  "description": "File-based screen routing for TermUI with typed params and navigation guards",
-  "type": "module",
-  "main": "./dist/index.cjs",
-  "module": "./dist/index.js",
-  "types": "./dist/index.d.ts",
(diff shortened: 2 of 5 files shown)


Debangana-Dutta:
hi @Karanjot786  The code implementation is complete, but the `CI / build-and-test` check is failing with `Could not resolve '@termuijs/motion'`. 

This is happening because `@termuijs/motion` is not listed in the dependencies inside `packages/router/package.json`. 

Per the strict issue description constraints (**"Do not touch bun.lock / any other package"**), I intentionally didn't modify the `package.json` to avoid violating the rules. Could you please let me know if I have permission to add `@termuijs/motion` to the router's dependencies so the build can pass, or if you'll handle it on merge? Thank you!

Karanjot786:
I fixed the build error — `transitions.ts` imported `@termuijs/motion` but `packages/router/package.json` didn't declare it (TS2688). Added the dependency and merged main; build passes now.

Two tests still fail, your transition logic:
```
× should capture navigate triggers and dispatch core happy-path transition frame updates
  → expected "spy" to be called 1 times, but got 0 times
× should run leave configurations cleanly when historical back steps occur
  → expected "spy" to be called 1 times, but got 0 times
```
The frame-update callback isn't firing on navigate/back. Check that the transition manager subscribes to the router's navigate event and actually drives the motion timeline (the spy should be called once per transition). Fix those two and CI goes green.

Debangana-Dutta:
@Karanjot786 i have  updated the `transitions.ts` file to use standard imports now that the dependency has been declared. Local testing and the CI pipelines are completely green and passing now. Ready for review and merge!

Issue #145, opened 2026-06-01
feat: add useLayoutEffect hook


Karanjot786:
## What to build

Add a `useLayoutEffect` hook. It runs synchronously after render and before the next paint. This differs from `useEffect`, which runs after paint. It uses the same deps-array semantics as `useEffect`.

## Acceptance criteria

- `useLayoutEffect(fn, deps)` runs the effect synchronously after the render commits
- The effect re-runs only when a value in the deps array changes
- Cleanup runs before the next effect and on unmount
- Has a `.test.ts` covering the behavior

## Implementation notes

- Place at `packages/jsx/src/hooks/useLayoutEffect.ts` (or add to `packages/jsx/src/hooks.ts` next to the existing hooks)
- Export from the package `src/index.ts`
- Look at `packages/jsx/src/hooks.ts` for the existing hook pattern (useState, useEffect, useMemo). Reuse the `useEffect` deps-compare logic and the `currentFiber()` plus `hookIndex` slot pattern.

## Example usage

```typescript
useLayoutEffect(() => {
  measureLayout()
}, [size])
```

## GSSoC 2026

This is an intermediate-level issue. Comment "I would like to work on this" to get assigned. You have 7 days to open a PR after assignment.

Read [CONTRIBUTING.md](./CONTRIBUTING.md) before starting.

CoderPrateek971:
Hi @Karanjot786, Kindly assign me this under GSSOC'26.

CoderPrateek971:
@Karanjot786 My PR is merged but this issue is not closed yet.

PR close #572

Karanjot786:
In a terminal renderer, `useLayoutEffect` runs its effect synchronously after the render tree is committed but before the screen buffer is flushed to stdout — so any changes appear in the same frame. `useEffect` queues after the flush. The implementation lives alongside the existing hooks in `packages/jsx/src/hooks.ts`: reuse the deps-compare logic from `useEffect` and invoke the callback inside the commit phase rather than post-flush. Export from `packages/jsx/src/index.ts` when done. Good luck!

jagnyaseni-panda:
Hi @Karanjot786  ,

I am a GSSoC'26 contributor.

I noticed that this issue is already assigned. If it becomes available, could you please consider assigning it to me?

I would like to work on this.

Thank you !

harsh0028-boop:
i would like to work on this

PR #572, opened 2026-06-03
type: Uselayout effect hook added

CoderPrateek971:
Implemented `useLayoutEffect` for the JSX package.

### Changes made

* Added `useLayoutEffect(fn, deps)` hook implementation.
* Reused the existing dependency comparison logic used by other hooks.
* Ensured effects run synchronously after commit.
* Added cleanup handling before re-runs and on unmount.
* Exported the hook from the package public API.
* Added test coverage for execution, dependency updates, and cleanup behavior.

This PR addresses Issue #145.

## New Features
* Added `useLayoutEffect` hook for managing layout-dependent side effects with automatic dependency tracking and cleanup handling.
* Layout effects execute during the render and reconciliation process at the appropriate time in the component lifecycle.
* Enhanced component teardown to properly clean up all associated layout effects.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -562,9 +562,6 @@
 
     "@termuijs/example-auth-flow": ["@termuijs/example-auth-flow@workspace:examples/auth-flow"],
 
-    "@termuijs/example-cli-wrapper-live": ["@termuijs/example-cli-wrapper-live@workspace:examples/cli-wrapper-live"],
-
-    "@termuijs/example-dashboard": ["@termuijs/example-dashboard@workspace:examples/dashboard"],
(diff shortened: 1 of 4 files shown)


Karanjot786:
Two fixes in `hooks.ts`:

**1. Stale cleanup after `runLayoutEffects`** (line 547-555): After calling the effect, if `cleanup` is not a function, explicitly set `record.cleanup = undefined`. Otherwise an old cleanup stays and gets called by `destroyFiber()` when it shouldn't.

**2. Bare type assertion** (line 255): Replace `prev.value as EffectRecord` with a runtime type guard that checks the shape before casting.

CoderPrateek971:
@Karanjot786 I have to make the changes or not as the branch is merged. Also is this counted under gssoc'26?

PR #696, opened 2026-06-04
feat(core): add visible focus ring support

### LINKED ISSUE #325: feat(core): add focus ring indicators
> For AI agents (Claude Code, Copilot, Cursor): read [AGENTS.md](https://github.com/Karanjot786/TermUI/blob/main/AGENTS.md) and [packages/core/AGENTS.md](https://github.com/Karanjot786/TermUI/blob/main/packages/core/AGENTS.md) before writing code. This issue is a complete spec. Follow it exactly.

## Objective

Add a visible focus ring (outline) to the focused widget in `@termuijs/core`, driven by the existing focus system.

## Package scope

`packages/core`. All work is confined here. Import shared types from `@termuijs/core`.

## Files to create or modify

```
Files to create or modify are described in the objective and reference file.
Follow the reference pattern and confine all changes to packages/core.
```

## Acceptance criteria

- [ ] the new feature works correctly in isolation
- [ ] existing tests in `packages/core` continue to pass
- [ ] `bun vitest run packages/core` passes
- [ ] `bun run typecheck` passes

## Test expectations

```
Run:  bun vitest run packages/core
```

Add tests for the core behavior. Use real objects from `@termuijs/core`. Use `vi.spyOn` for caps, never mutate it.

## Reference pattern to follow

Follow `packages/jsx/src/hooks/useFocus.ts` for structure and conventions.

## Not included (out of scope)

- No mouse support.
- No docs-site changes.
- No JSX intrinsic registration.

## Do not touch

- Packages outside `packages/core`
- `bun.lock`
- `.github/`

## Agent prompt (copy this into your AI agent)

```
Read AGENTS.md at the repo root and packages/core/AGENTS.md.
Read packages/jsx/src/hooks/useFocus.ts and copy its structure.
Task: implement the feature in this issue.
Plan first. Confine all changes to packages/core. Do not edit bun.lock. No new dependencies.
Tests use the real Screen and vi.spyOn for caps, never mutate caps.
When done run: bun vitest run packages/core && bun run typecheck. Both must pass.
```

## Before you open the PR

- [ ] I run `bun run build && bun vitest run && bun run typecheck` and all pass
- [ ] My change is confined to `packages/core`
- [ ] `bun.lock` has no unrelated changes
- [ ] I starred the repo

## GSSoC 2026

This is a intermediate-level issue. Comment "I would like to work on this" to get assigned. You have 7 days to open a PR after assignment.

Read [CONTRIBUTING.md](./CONTRIBUTING.md) before starting.

Pushtikansara:
## Description
This PR connects the existing focus system with widget focus state so that focused widgets correctly render their focus ring.
##Changes
-Added focus and blur event handling in App.
-Synchronized widget isFocused state with FocusManager.
-Triggered widget re-rendering when focus changes.
-Added tests covering focus propagation and synchronization behavior.

## Related Issue
Closes #325 

## Which package(s)?
@termuijs/core

## Type of Change

- [ ] 🐛 Bug fix (`type:bug`)
- [ ] 📝 Docs (`type:docs`)

## Checklist

## GSSoC 2026 Participation
- https://gssoc.girlscript.org/profile/adc9e69c-a9de-4469-ba44-ed6de335d366

N/A

## Notes for the Reviewer
This change synchronizes widget focus state with FocusManager events and ensures focused widgets are re-rendered when focus changes. Tests have been added to verify focus state propagation and synchronization behavior.

## Summary by CodeRabbit

* **New Features**
  * App-level focus tracking so focus-aware widgets visibly reflect focus, queue early focus events until widgets register, trigger redraws only on actual focus changes, and stop updates after unmount.
* **Tests**
  * End-to-end integration tests for focus transitions, dirty-state propagation, render counts, registration-before-mount synchronization, and unmount behavior.


Karanjot786:
Code review — feat(core): add visible focus ring support

The approach is correct: subscribe to FocusManager focus/blur events in App, propagate isFocused to widgets, and request a re-render. The bones are solid. There are specific issues below that need fixing before merge.

**Bug 1: Duplicate subscriptions on every mount() call (event listener leak)**

The constructor already sets `_unsubFocus` and `_unsubBlur`. The guard added in `mount()` checks `if (!this._unsubFocus)` — but because the constructor always sets them, that branch never runs. So the guards are dead code.

However, `App` is a single-use object: calling `mount()` twice is guarded by `if (this._mounted) return 0`. The real leak risk is the *constructor itself* always subscribing, even when the app is used only for fallback rendering (no unmount called). The `unmount()` path does call the unsub functions, so interactive use is fine. But a path that calls `mount()` into fallback and returns early before setting `_mounted = true` will never reach `unmount()`, and the subscriptions stay alive indefinitely.

Fix: move the subscription setup to the start of the interactive path in `mount()`, after the fallback check, and remove the constructor subscriptions.

**Bug 2: _handleFocusEvent fires before the widget map is populated**

`_widgetById` is populated inside `requestRender()` via `_buildWidgetMap()`. The focus subscriptions are established in the constructor. If a widget calls `app.focus.register(...)` before `mount()`, the FocusManager auto-focuses the first widget immediately (see FocusManager.register line 76), which emits a `focus` event, which calls `_handleFocusEvent`, which calls `_setWidgetFocused`, which calls `_widgetById.get(id)` — and gets `undefined` because `_buildWidgetMap` has not run yet.

The third test ("syncs focus state when focus events happen before the widget map exists") acknowledges this scenario but the assertions show `isFocused` remains `false` before `mount()`. That means the handler silently did nothing, which is exactly the bug: the initial focus event is lost. The comment in the test says it "syncs" but the widget stays unfocused until `mount()` runs and `requestRender()` calls `_walkWidget` which sets `isFocused` from `focus.currentId`. So the behavior works by coincidence through the walk, not through the event handler.

Fix: in `_handleFocusEvent`, if `_widgetById` is empty or the widget is not found, store the pending focused ID and apply it inside `_buildWidgetMap` after the map is rebuilt.

**Bug 3: blur event never clears the previously focused widget when cycling**

`_changeFocus` in FocusManager emits blur for the old widget, then emits focus for the new one. `_handleFocusEvent` calls `_setWidgetFocused(event.targetId, event.type === 'focus')`. For a blur event, `focused` is `false`. This works correctly IF `_widgetById` holds the old widget. Since `_buildWidgetMap` clears and rebuilds the map on every `requestRender()`, and the blur event fires before the render, the old widget should still be in the map from the previous render cycle. This is fragile: if `requestRender()` runs between blur and focus (which won't happen synchronously but shows the design is order-dependent).

No immediate code change required, but document this ordering dependency explicitly with a comment.

**Bug 4: requestRender() inside _handleFocusEvent is a no-op when not mounted**

`requestRender()` returns early if `!this._mounted`. When focus events fire before `mount()` (the scenario in test 3), `_handleFocusEvent` calls `requestRender()` which silently returns. The test passes only because `mount()` triggers `requestRender()` unconditionally after setup. This means the focus event path and the mount path both try to trigger a render, but only the mount path succeeds. The `changed = true` return from `_setWidgetFocused` is wasted. No bug in practice, but misleading.

**Bug 5: The "focus ring" title is misleading — no ring is rendered**

The PR title says "add visible focus ring support." The diff adds no rendering of any visual ring. It propagates `isFocused` to widgets. Rendering a focus ring is entirely up to each widget implementation. The PR should either render a ring inside App (it does not) or retitle itself as "propagate focus state to widgets." This matters for GSSoC point labeling — `level:advanced` on a feature that is half the claimed scope is over-credited.

**Bug 6: FocusAwareWidget interface uses `any` in _widgetById without documenting why**

`_widgetById` is typed as `Map<string, any>`. The PR adds `_isFocusAwareWidget` to narrow the type at the call site, which is fine, but the existing `any` map is never documented. Your checklist says "No new `any` types without an inline comment explaining why." The PR does not add a new `any` here, but if you are touching this map, add the inline comment the CONTRIBUTING rules require.

**Test gap: no test for the listener-leak path**

None of the three new tests verify that calling `unmount()` removes the focus/blur subscriptions (i.e., that a subsequent `focus.focusNext()` after unmount does not call `requestRender()` or mutate widget state). Add a test that mounts, unmounts, then calls `focus.focusNext()` and asserts no widget state changes.

**Minor: missing newline at end of original App.ts**

The diff shows `+}` without a trailing newline at the old end of file. The new file ends with `+}` and a newline. Good fix, but worth noting the original file was malformed.

**Summary of required changes before merge:**

- Move focus subscriptions out of the constructor and into the interactive mount path (after the fallback check) to prevent subscriptions from leaking in fallback-only usage.
- Document or fix the pre-mount focus event loss (test 3 currently asserts broken behavior as if it were correct).
- Add an unmount/unsubscribe test.
- Either render an actual focus ring in App or retitle the PR to match what it delivers, and adjust the GSSoC level label accordingly.

Pushtikansara:
Thanks for the review.

I've addressed the requested changes:

* Moved focus subscriptions into the mount lifecycle.
* Fixed focus synchronization before widget map creation.
* Added an unmount/unsubscribe test.
* Resolved merge conflicts with the latest main branch.

Validation completed successfully:

* bun vitest run packages/core ✅
* bun run typecheck ✅
* bun run build ✅

I'd appreciate another review when you have time.

Karanjot786 (commented):
Re-review: The focus ring implementation looks solid. However, CI shows `star-check` is failing — the contributor needs to star the repo per GSSoC rules. Please star the repo and re-request review.

Additionally, this modifies `packages/core/` which requires a linked issue per AGENTS.md rules.

Pushtikansara:
Thanks for the review.

I've already starred the upstream repository and pushed an update to rerun the checks. The PR is linked to issue #325.
I'd appreciate another review when you have time.

Karanjot786:
Fixed the build error directly.

**Root cause:** `_widgetById` was declared twice in the `App` class body — once at line 94 (with the explanatory comment) and again at line 98 (a duplicate introduced during a rebase/merge). This caused:
- esbuild warning: `[duplicate-class-member]`  
- TypeScript DTS error: `TS2300 Duplicate identifier '_widgetById'`
- `@termuijs/core` build failure

**Fix:** Removed the second (bare) declaration at line 98. The first declaration with its comment is the canonical one.

No action needed from you.

Karanjot786:
Fixed the remaining build errors and all test failures. Here is a full breakdown of everything that was repaired:

**`App.ts` — Build errors (DTS failure)**
- Removed duplicate `private _unsubPaste` field (lines 82 and 88 were both present)  
- Added missing `private _unsubFocus` and `private _unsubBlur` declarations — these were referenced in `unmount()` and `_subscribeFocusEvents()` but never declared, causing `TS2551`/`TS2339` errors in the DTS build

**`App.test.ts` — Test crash (vitest worker killed)**  
Four distinct bugs all interacted to kill the worker via `Terminal._unhandledRejectionHandler → process.exit(1)`:

1. **TypeScript parameter properties** (`readonly id: string` in constructor) — Node's strip-only TS mode doesn't support them; converted to explicit field declarations
2. **Missing `once()` on fake stdout** in `createInteractiveTestOptions()` — `Terminal._processWriteQueue()` calls `stdout.once('drain', ...)` when `write()` returns falsy. The stub returned `undefined` (falsy); fixed by returning `true` and adding `once() {}`
3. **`does not merge borders` test** — created `App` without fake streams, leaving real `process`-level signal/error handlers registered across tests; fixed with fake streams + `app.terminal.restore()` in `finally`
4. **Focus integration tests** — `focus.register()` was called after `mount()`, but `requestRender()` now defers via `setImmediate()`, so the widget map wasn't built yet when focus events arrived; fixed by moving `register()` before `mount()` and adding `await new Promise(r => setImmediate(r))` before render assertions

**Result: 27 test files, 322 tests, all passing. No action needed.**

PR #735, opened 2026-06-04
chore: explicitly define module distribution type in root package.json

### LINKED ISSUE #734: chore: explicitly define module distribution type in root package.json
The root package.json relies on default module loader resolution. Adding an explicit module type field improves configuration predictability for developers and bundlers interacting with the monorepo packages.

Rish-2006:
Closes Karanjot786/TermUI#734

## Summary by CodeRabbit

* **Chores**
  * Updated package configuration to use ES modules as the default module system.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "name": "termui",
   "version": "0.0.5",
   "private": true,
-  "description": "A TypeScript-first framework for building beautiful, interactive terminal user interfaces",
+    "type": "module",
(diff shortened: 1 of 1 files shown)


Karanjot786:
Thank you so much for your contribution, @Rish-2006!

Your PR #735 added an explicit module type to the root package.json. This removes ambiguity for tools that check the module format before processing files.

Keep contributing. We appreciate your work.

PR #770, opened 2026-06-05
fix(core): fix multi-byte UTF-8 input corruption by using code point iteration and raw Buffer for escape buffer

### LINKED ISSUE #753: [bug] Multi-byte UTF-8 Input Corruption — Non-ASCII Characters Split into Garbled Bytes
## Description
The `_processInput()` method converts raw stdin `Buffer` to a JS string via `data.toString('utf8')`, then iterates with `str[i]` (UTF-16 code units). This splits multi-byte UTF-8 characters (emoji, CJK, accented characters) into surrogate halves or individual bytes. Each byte fragment is treated as a separate "key press" and emitted as a distinct `KeyEvent`. The raw buffer reconstruction `Buffer.from(ch, 'utf8')` only captures a single code unit, so non-ASCII input is completely corrupted.

## Files Affected
- `packages/core/src/input/InputParser.ts` — `_processInput()` (line 105), `_tryParseEscape()` (line 227), raw buffer reconstruction (line 163)

## Expected Behavior
When a user types a multi-byte UTF-8 character (e.g., CJK ideograph, emoji, accented letter), the character should be emitted as a single `KeyEvent` with the correct character value.

## Actual Behavior
Multi-byte characters are split into individual continuation bytes. Each byte is emitted as a separate `KeyEvent` with garbled character values. Non-ASCII textual input is completely unusable.

## Steps to Reproduce
1. Create a TermUI application that listens for text input
2. Type a non-ASCII character such as `ñ` (U+00F1, 2 bytes in UTF-8) or `中` (U+4E2D, 3 bytes)
3. Observe that the input handler receives multiple key events instead of one
4. The reconstructed character is garbled

## Impact
Any user typing non-ASCII characters (CJK text via IME, emoji, accented Latin characters, math symbols) will see them split into garbled individual bytes. Terminal applications built with TermUI that accept text input from international users are completely broken.

## Fix Required
Rewrite the non-escape input path to process JavaScript code points via `for (const char of str)` or `str.codePointAt()` instead of indexed access. The escape sequence parsing path also needs to handle continuation bytes properly. Affects `_processInput()`, `_tryParseEscape()`, and the raw buffer reconstruction logic.

ionfwsrijan:
## Summary

Fixes garbled multi-byte character input (e.g., `é`, `ü`, CJK) caused by byte-level iteration treating each continuation byte as a separate character.

## Root Cause

`InputParser` used `string[index]` indexing on the escape buffer, which operates on UTF-16 code units — splitting multi-byte characters into invalid fragments.

## Changes

- **`packages/core/src/input/InputParser.ts`** — changed `_escapeBuffer` from `string` to `Buffer`; replaced index-based character access with code-point iteration via `Buffer.toString("utf-8")` + `for...of`

## Testing

- Send UTF-8 input via stdin (`echo "éüñ" | node ...`)
- Verify each multi-byte character is received as a single `key` event
- Verify non-printable sequences (arrows, Ctrl+C) remain unaffected

Closes #753 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of escape sequences and input buffering for more reliable keyboard and mouse input parsing, including better handling of partial/oversized sequences, Alt+key combos, and cursor/focus reports, reducing erroneous or missed inputs during rapid or mixed input streams.

### DIFF
--- a/packages/core/src/input/InputParser.ts
+++ b/packages/core/src/input/InputParser.ts
@@ -2,6 +2,7 @@
 // @termuijs/core — Input Parser
 // ─────────────────────────────────────────────────────
 
+import { Buffer } from 'node:buffer';
 import type { KeyEvent, MouseEvent } from '../events/types.js';
 import { createKeyEvent } from '../events/types.js';
 import { ESCAPE_SEQUENCES, CTRL_KEYS, SPECIAL_KEYS } from './KeyMap.js';
@@ -29,7 +30,7 @@ export class InputParser {
     private _stdin: NodeJS.ReadStream;
     private _handler: ((data: Buffer) => void) | null = null;
     private _escapeTimeout: ReturnType<typeof setTimeout> | null = null;
-    private _escapeBuffer = '';
+    private _escapeBuffer = Buffer.alloc(0);
     private _isPasting = false;
     private _pasteBuffer = '';
     private _cursorRequests: Array<{
@@ -96,7 +97,7 @@ export class InputParser {
             clearTimeout(this._escapeTimeout);
             this._escapeTimeout = null;
         }
-        this._escapeBuffer = '';
+        this._escapeBuffer = Buffer.alloc(0);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

Karanjot786:
Hi @ionfwsrijan,

This PR is approved for merge.

Switching `_escapeBuffer` from a UTF-16 string to a `Buffer` and using `for...of` code-point iteration for non-escape characters is the correct fix for multi-byte input corruption. Characters like `é`, `ü`, and CJK glyphs all need proper code-point handling, not index-based string access.

Adding test coverage for multi-byte input scenarios would strengthen this, but the code change itself is solid.

Thank you.

PR #778, opened 2026-06-05
fix(core): resolve mount() promise when unmount() is called directly

### LINKED ISSUE #776: [BUG] App.mount() promise never resolves when unmount() is called directly
## Bug Report

**Package:** `@termuijs/core`
**File:** `packages/core/src/app/App.ts`
**Priority:** High
**Branch:** `fix/core-app-unmount-resolve`

## Description

`App.mount()` returns a promise that only resolves when `exit()` is called internally.
Calling `unmount()` directly — which is common in tests, wrappers, and custom shutdown
flows — tears down input and renderers correctly, but leaves the `mount()` promise
pending forever, causing hangs.

## Steps to Reproduce

\`\`\`typescript
const app = new App({ ... })
const mountPromise = app.mount()
app.unmount() // tears down correctly
await mountPromise // ← hangs forever, never resolves
\`\`\`

## Expected Behavior

`mount()` should resolve (with exit code `0` by default) when `unmount()` is called.

## Actual Behavior

The promise returned by `mount()` is never resolved or rejected, causing `await app.mount()`
to hang indefinitely in tests and shutdown wrappers.

## Root Cause

`mount()` stores `_exitResolve` and only `exit()` calls it. `unmount()` does not invoke
`_exitResolve`, so any caller awaiting `mount()` is permanently blocked.

## Suggested Fix

In `unmount()`, after confirming the app is currently mounted, call `_exitResolve?.(0)` to
resolve the pending promise. This aligns `unmount()` behavior with `exit()` for the purpose
of the mount lifecycle.

## Tests to Add

In `App.test.ts`:
- Mount an app, call `unmount()`, assert the `mount()` promise resolves (with `Promise.race`
  against a timeout to fail fast if it hangs)

Pranaykarvi:
## Summary

Closes #776

## Problem

`App.mount()` returns a promise that only resolved when `exit()` was called internally.
Calling `unmount()` directly left `_exitResolve` never invoked, causing the `mount()`
promise to hang forever in tests and custom shutdown wrappers.

## Fix

In `unmount()`, after the existing teardown (`events.removeAll()`), added:

```ts
if (this._exitResolve) {
    this._exitResolve(0);
    this._exitResolve = null;
}
```

`exit()` is completely untouched — it still calls `unmount()` internally so
existing behavior is preserved.

## Files Changed

- `packages/core/src/app/App.ts` — resolve `_exitResolve` in `unmount()`
- `packages/core/src/app/App.test.ts` — new test covering the hang scenario

## Test Results

- `bun vitest run packages/core` — 155 passed
- `bun run typecheck` — passed
- `bun run build` — all packages built successfully

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown so unmounting now reliably resolves in-progress startup and propagates an explicit exit code.
* **Tests**
  * Added a test ensuring that forcing an unmount causes pending startup to complete promptly and honor the provided exit code.

### DIFF
--- a/packages/core/src/app/App.test.ts
+++ b/packages/core/src/app/App.test.ts
@@ -31,6 +31,42 @@ function createMockRootWidget(): RootWidget {
 }
 
 describe('App', () => {
+    describe('unmount()', () => {
+        it('mount() promise resolves when unmount() is called directly', async () => {
+            const root = createMockRootWidget();
+            const fakeStdout: any = {
+                writes: '',
+                columns: 80,
+                rows: 24,
+                isTTY: true,
+                write(s: string) { this.writes += s; },
+                on() {}, off() {},
+            };
+            const fakeStdin: any = { isTTY: true, setRawMode() {}, resume() {}, pause() {}, on() {}, off() {} };
+
+            const app = new App(root, {
+                forceFallback: false,
+                skipFallback: true,
+                screenMode: 'main',
+                stdout: fakeStdout,
+                stdin: fakeStdin,
+            } as AppOptions);
+
+            const mountPromise = app.mount();
+
+            await new Promise((r) => setTimeout(r, 10));
+
+            app.unmount();
+
+            const result = await Promise.race([
+                mountPromise.then(() => 'resolved'),
+                new Promise<string>((r) => setTimeout(() => r('timeout'), 500)),
+            ]);
+
+            expect(result).toBe('resolved');
+        });
+    });
+
     describe('exit()', () => {
         it('does NOT call process.exit when called before mount()', () => {
             const exitSpy = vi.spyOn(process, 'exit').mockImplementation(() => undefined as never);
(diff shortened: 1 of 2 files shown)


Pranaykarvi:
Hey @Karanjot786, 
The change is a single guard in `unmount()` that calls `_exitResolve(0)` 
after teardown — `exit()` is untouched. New test covers the hang scenario 
using `Promise.race` with a timeout. All 155 core tests pass.

Pranaykarvi:
@Karanjot786,
All CodeRabbit comments addressed:

1. **`any` types** — added inline justification comments on all three flagged lines in the test file.
2. **`exit(code)` semantics** — `unmount()` now accepts `exitCode: number = 0` and resolves with that value; `exit(code)` passes the code down via `this.unmount(code)`.

All 155 core tests pass. `typecheck` clean. Ready for review @Karanjot786 🚀

Karanjot786:
Hi @Pranaykarvi,

This PR is approved for merge.

The bug is real: calling `unmount()` directly left `mount()` permanently pending. Resolving `_exitResolve` inside `unmount()` is the correct fix. The 500ms timeout guard in the test prevents the suite from hanging if the promise never resolves.

One cleanup worth doing in this same PR: after your change, the `if (this._exitResolve)` block inside `exit()` is now dead code since `unmount()` always nulls it out first. Remove that block.

Thank you for the contribution.

Issue #880, opened 2026-06-06
[bug] Layout Cache Invalidation Is Completely Broken


ionfwsrijan:
## Description
The `_dirty` flag on `LayoutNode` is initialized to `true`, set to `false` after `computeLayout()`, but **never re-set to `true`** when a widget's layout-affecting style changes. Meanwhile, the Widget's `markDirty()` propagates upward independently but has zero connection to the layout node cache. This means every `requestRender()` call recomputes the **entire** layout tree from scratch, even when nothing changed. In deep widget trees, this is O(n) wasted work every frame (30+ fps). The `_dirty` field exists solely as dead weight — it adds complexity and occupies memory but provides zero performance benefit.

## Files Affected
| File | Lines | Role |
|------|-------|------|
| `packages/core/src/layout/LayoutEngine.ts` | Full | Sets `node._dirty = false` after layout, never checks it before computation. Must add dirty check, `invalidateLayout()` API, and skip recursion for clean subtrees. |
| `packages/widgets/src/base/Widget.ts` | Full | `getLayoutNode()` creates a **new** `LayoutNode` every call, discarding the old `_dirty` state. Must cache the `_layoutNode` across calls and call `invalidateLayout()` on `markDirty()`. |
| `packages/core/src/app/App.ts` | Full | `requestRender()` calls `getLayoutNode()` then `computeLayout()` unconditionally. Must only call `computeLayout()` when layout is dirty. |
| `packages/core/src/style/Style.ts` | Full | No mechanism to detect layout-relevant style changes. Must add `getLayoutFingerprint()` or change-tracking proxy for layout-affecting props. |
| `packages/core/src/layout/LayoutEngine.test.ts` | Full | Must add tests for layout skip on clean nodes, invalidation propagation, and performance regression. |
| `packages/widgets/src/data/markDirty.test.ts` | Full | Must extend to test that `markDirty()` invalidates layout cache. |
| `packages/widgets/src/base/Widget.test.ts` | Full | Must add tests: `setStyle` invalidates layout, `addChild`/`removeChild` invalidates layout. |

## Expected Behavior
- A widget subtree with no layout-affecting changes should completely skip `computeLayout()` for that subtree.
- When a widget's layout-relevant style changes (width, height, flexGrow, padding, margin, border, flexDirection, justifyContent, alignItems, gap, min/max dimensions), the layout cache for that branch must be invalidated.
[...]

PR #895, opened 2026-06-06
fix: make layout cache invalidation actually work

ionfwsrijan:
### Issue #880

Widgets rendered stale positions after JSX reconciliation because the layout cache was never
invalidated when the widget tree changed.

### Changes
- `packages/core/src/layout/LayoutEngine.ts`: Added `invalidateLayout()` method + dirty-skip
  to avoid redundant recalculations
- `packages/core/src/style/Style.ts`: Added `LAYOUT_PROPS` set, `hasLayoutChanges()` to
  detect style changes that require re-layout
- `packages/jsx/src/reconciler.ts`: Calls `invalidateLayout()` after every widget replacement
- `packages/widgets/src/base/Widget.ts`: `styleDidChange` no longer triggers full remount
- Added test coverage in `packages/core/src/layout/__tests__/layout-engine.test.ts`

### Testing
```
npx jest packages/core/src/layout/__tests__/layout-engine.test.ts
```

## Summary by CodeRabbit

* **Performance Improvements**
  * Skip unnecessary layout recomputation when the layout tree is clean
  * Reuse cached layout nodes to reduce widget recreation and speed re-renders

* **New APIs**
  * Add utilities to invalidate layout caches and detect style changes that affect layout

* **Tests**
  * Added tests covering layout cache invalidation and recomputation behavior


ionfwsrijan:

Karanjot786:
There is a resize regression in the layout cache implementation that needs fixing before this PR is ready.

The cache key uses only the widget's own width and height. When the terminal is resized, children get new absolute positions during layout, but if no widget is marked dirty, the cache returns the stale layout and children render at wrong positions.

To fix: include the terminal dimensions (or a full layout fingerprint that covers child positions) in the cache key, or invalidate the cache on any resize event. The existing resize tests in core pass, but a test like this will fail:

   resize terminal from 80x24 to 100x24
   assert child widget renders at the correct new x/y position

Add this test and fix the invalidation logic, then push an update.

Karanjot786:
Correction to my earlier review — the resize regression concern was wrong. The App.ts resize handler already calls markDirty() on the root widget (line 151), which now propagates to layoutNode._dirty through the updated markDirty() implementation. The layout cache invalidates correctly on resize. PR is approved and ready to merge.

ionfwsrijan:
@Karanjot786 Alright. Working on it sir.

PR #932, opened 2026-06-06
fix(router): resolve initialPath on startup via deferred _navigateTo

### LINKED ISSUE #930: [BUG] Router initialPath does not render initial screen on startup
## Bug Report

**Package:** `@termuijs/router`
**File:** `packages/router/src/router.ts`
**Priority:** High
**Branch:** `fix/router-initial-path-navigate`

## Description

The Router constructor pushes `initialPath` onto `_history` when provided
but never sets `_currentMatch`, calls `matchRoute()`, or emits a `navigate`
event. Any consumer reading `router.current` or listening for `navigate`
on startup sees `null` until an explicit `push()` is called manually.
This breaks every app that relies on the router rendering a screen on load.

## Steps to Reproduce

```typescript
const router = new Router({
    initialPath: '/dashboard',
    routes: [
        { path: '/dashboard', component: Dashboard },
    ]
})

console.log(router.current) // null — should be the dashboard route match
// navigate event never fires on startup
```

## Expected Behavior

On construction, if `initialPath` is provided, the router should resolve
the matching route, set `_currentMatch`, and emit a `navigate` event —
exactly the same as calling `push(initialPath)` manually.

## Actual Behavior

`router.current` is `null` on startup. No `navigate` event fires.
The initial screen never renders until `push()` is called explicitly.

## Root Cause

The constructor only pushes `initialPath` onto `_history` without calling
`matchRoute()` or emitting `navigate`. The navigation logic that runs in
`push()` is never triggered on initialization.

## Suggested Fix

After routes are registered in the constructor, call an internal
`_navigateTo(initialPath)` method that runs the same logic as `push()`
— matches the route, sets `_currentMatch`, and emits `navigate`.

## Tests to Add

- Construct a router with `initialPath` and assert `router.current` is
  not null immediately after construction
- Assert a `navigate` event fires on construction with the correct route
- Assert the matched route has the correct path and params

Pranaykarvi:
## Summary

Closes #930

## Problem

The Router constructor pushed `initialPath` onto `_history` but never
called `matchRoute()`, set `_currentMatch`, or emitted `navigate`.
Any consumer reading `router.current` on startup got `null`, and the
initial screen never rendered until `push()` was called manually.

## Fix

Since routes are registered after construction via `addRoute()`, a
deferred pattern was used:

- Added `_pendingInitialPath` field to store the path until a matching
  route exists
- Added `_applyInitialPathIfPending()` — called from the constructor
  and after each `addRoute()` — which matches the route, sets
  `_currentMatch`, and emits `navigate` once a match is found
- `push()` now delegates to `_navigateTo()` for consistency
- `replace()` and `back()` are untouched

## Files Changed

- `packages/router/src/router.ts` — deferred initialPath resolution
- `packages/router/src/router.test.ts` — 3 new tests covering startup behavior

## Test Results

- `bun vitest run packages/router` — 16 passed 
- `bun run typecheck` — passed 
- `bun run build` — all packages built successfully 

## Summary by CodeRabbit

* **Bug Fixes**
  * Initial path is now applied once routes are registered; parameterized paths resolve correctly and a single navigation event is emitted when matched.

* **Tests**
  * Added tests covering initial-path application, single navigation event emission, parameter resolution, and a negative case where no match leaves the router unchanged.

### DIFF
--- a/packages/router/src/router.test.ts
+++ b/packages/router/src/router.test.ts
@@ -191,4 +191,15 @@ describe('Router', () => {
         expect(r.current).not.toBeNull();
         expect(r.params.id).toBe('42');
     });
+
+    it('initialPath with no matching route never sets current or emits navigate', () => {
+        const navFn = vi.fn();
+        const r = new Router({ initialPath: '/never' });
+        r.events.on('navigate', navFn);
+        r.addRoute('/dashboard', () => 'Dashboard');
+        r.addRoute('/settings', () => 'Settings');
+
+        expect(r.current).toBeNull();
+        expect(navFn).not.toHaveBeenCalled();
+    });
 });
--- a/packages/router/src/router.test.ts
+++ b/packages/router/src/router.test.ts
@@ -164,4 +164,31 @@ describe('Router', () => {
         
         expect(spy).toHaveBeenCalled();
     });
+
+    it('initialPath sets current match once a matching route is registered', () => {
+        const r = new Router({ initialPath: '/dashboard' });
+        r.addRoute('/dashboard', () => 'DashboardScreen');
+
+        expect(r.current).not.toBeNull();
+        expect(r.currentPath).toBe('/dashboard');
+        expect(r.current?.route.path).toBe('/dashboard');
+    });
+
+    it('initialPath emits navigate when a matching route is registered', () => {
+        const navFn = vi.fn();
+        const r = new Router({ initialPath: '/home' });
+        r.events.on('navigate', navFn);
+        r.addRoute('/home', () => 'Home');
+
+        expect(navFn).toHaveBeenCalledOnce();
+        expect(navFn.mock.calls[0][0].match.route.path).toBe('/home');
+    });
+
+    it('initialPath resolves route params when a matching route is registered', () => {
+        const r = new Router({ initialPath: '/user/42' });
+        r.addRoute('/user/[id]', () => 'UserScreen');
+
+        expect(r.current).not.toBeNull();
+        expect(r.params.id).toBe('42');
+    });
 });
--- a/packages/router/src/router.ts
+++ b/packages/router/src/router.ts
@@ -42,14 +42,15 @@ export class Router {
     private _forwardStack: string[] = [];
     private _currentMatch: RouteMatch | null = null;
     private _maxHistory: number;
-
+    private _pendingInitialPath: string | null = null;
     readonly events = new EventEmitter<RouterEvents>();
 
     constructor(options: RouterOptions = {}) {
         this._maxHistory = options.maxHistory ?? 100;
 
         if (options.initialPath) {
-            this._history.push(options.initialPath);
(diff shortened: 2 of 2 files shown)


Pranaykarvi:
Hey @Karanjot786, 

Used a deferred pattern — `_pendingInitialPath` stores the path on
construction, then `_applyInitialPathIfPending()` fires once a matching
route is registered via `addRoute()`. Three new tests cover `router.current`,
the `navigate` event, and param resolution on startup. All 16 router tests pass

Pranaykarvi:
@Karanjot786 the CI failure is in `packages/core/src/layout/LayoutEngine.test.ts` 
— `clean node skips computation` — which is unrelated to this PR's changes 
(router package only). This test was already failing on upstream main before 
this branch was created. Happy to confirm if needed.

Pranaykarvi:
@Karanjot786 the CI failure is in `packages/core/src/layout/LayoutEngine.test.ts` 
(`clean node skips computation`) — this is a pre-existing failure on upstream main, 
unrelated to this PR which only touches the router package.

Verified by running the failing test locally on a clean pull of upstream main:
- `bun vitest run packages/core/src/layout/LayoutEngine.test.ts` → same failure
- All router tests pass: 24/24 ✅

Happy to rebase again once the upstream test is fixed, or this can be merged as-is 
since the failure predates this branch.

Karanjot786:
Good work identifying and fixing the root cause here. The deferred pattern is the right approach since routes register after construction.

Two small items to address before merge:

- The `_applyInitialPathIfPending()` call inside the constructor (line 58) is always a no-op. The guard `this._routes.length === 0` returns immediately every time because no routes exist yet at construction time. Remove this call to keep the code honest about what happens at each site.

- There is no test for the "no match ever found" edge case. Add one test that creates `Router({ initialPath: '/never' })`, registers only unrelated routes, and asserts `r.current` stays null and `navigate` is never called. This confirms the pending path does not cause stale state.

On the CI failure: the failing test (`layout cache invalidation > clean node skips computation`) is in `packages/core/src/layout/LayoutEngine.test.ts` and is pre-existing on the main branch. It is not caused by this PR.

The core logic of `_navigateTo`, `_applyInitialPathIfPending`, and the three new tests are all correct. The fix properly sets `_currentMatch`, emits `navigate`, resolves params, and respects `beforeEnter` guards on the initial navigation.

Pranaykarvi:
@Karanjot786 Both review items are done in 9abfe8d:

Removed the no-op _applyInitialPathIfPending() call from the constructor — deferred navigation still runs from addRoute().
Added a test for initialPath: '/never' with unrelated routes — asserts current stays null and navigate is never emitted.
bun vitest run packages/router — 45/45

Karanjot786:
Thank you, @Pranaykarvi, for fixing the router's initialPath resolution on startup using a deferred _navigateTo call. This is a clean and well-targeted fix that improves the reliability of routing from the start. Great work on tracking down this issue!

PR #944, opened 2026-06-06
feat(tss): add semantic theme derivation from Normal color pair

Sneha6657:
## Changes
- Created `packages/tss/src/theme/derive.ts` — derivation logic
- Created `packages/tss/src/theme/derive.test.ts` — unit tests
- Updated `AutoThemeProvider.ts` — calls deriveTheme at startup
- Exported `deriveTheme` from `packages/tss/src/index.ts`

## Tests
- bun run build ✅
- bun run test ✅  
- bun run typecheck ✅

## Summary by CodeRabbit

* **New Features**
  * Improved automatic dark/light detection in terminal environments (including enhanced background probing) to pick appropriate themes.
  * Added theme derivation that generates semantic color roles (Normal, Focus, Active, Disabled, Highlight) from a base color pair.
  * Re-exported theme utilities for direct use.

* **Tests**
  * Added tests covering theme derivation, normalization, and error handling.

### DIFF
--- a/packages/tss/src/AutoThemeProvider.test.ts
+++ b/packages/tss/src/AutoThemeProvider.test.ts
@@ -45,7 +45,7 @@ describe('AutoThemeProvider', () => {
         const vnode = AutoThemeProvider({});
         clearCurrentFiber();
 
-        expect(vnode.props.value).toBe(defaultDark);
+        expect(vnode.props.value).toMatchObject({ Normal: { fg: defaultDark.fg, bg: defaultDark.bg } });
         destroyFiber(fiber);
     });
 
@@ -58,7 +58,7 @@ describe('AutoThemeProvider', () => {
         const vnode = AutoThemeProvider({});
         clearCurrentFiber();
 
-        expect(vnode.props.value).toBe(defaultLight);
+        expect(vnode.props.value).toMatchObject({ Normal: { fg: defaultLight.fg, bg: defaultLight.bg } });
         destroyFiber(fiber);
     });
 
@@ -72,7 +72,7 @@ describe('AutoThemeProvider', () => {
         setCurrentFiber(fiber);
         let vnode = AutoThemeProvider({ darkTheme: customDark, lightTheme: customLight });
         clearCurrentFiber();
-        expect(vnode.props.value).toBe(customDark);
+        expect(vnode.props.value).toMatchObject({ Normal: { fg: customDark.fg, bg: customDark.bg } });
         destroyFiber(fiber);
 
         // Light mode
@@ -81,7 +81,7 @@ describe('AutoThemeProvider', () => {
         setCurrentFiber(fiber);
         vnode = AutoThemeProvider({ darkTheme: customDark, lightTheme: customLight });
         clearCurrentFiber();
-        expect(vnode.props.value).toBe(customLight);
+        expect(vnode.props.value).toMatchObject({ Normal: { fg: customLight.fg, bg: customLight.bg } });
         destroyFiber(fiber);
     });
 
--- a/packages/tss/src/AutoThemeProvider.ts
+++ b/packages/tss/src/AutoThemeProvider.ts
@@ -1,8 +1,9 @@
 import { createContext, useContext, useState, useEffect } from '@termuijs/jsx';
 import type { FC, VNode } from '@termuijs/jsx';
-import { caps } from '@termuijs/core';
+import { ansi, caps, parseColor, relativeLuminance } from '@termuijs/core';
 import { detectDark, defaultDark, defaultLight, systemTheme } from './tokens.js';
 import type { ThemeTokens } from './tokens.js';
+import { deriveTheme } from './theme/derive.js';
 
 /**
  * Context holding the current ThemeTokens.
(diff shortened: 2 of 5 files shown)


Karanjot786:
## Review

This PR adds semantic theme derivation from a Normal color pair and integrates OSC-based dark-mode detection into `AutoThemeProvider`.

### Issues to Fix

**1. Breaking change to `useTheme()` return type without a migration path**

`useTheme()` previously returned `ThemeTokens`. It now returns `ReturnType<typeof deriveTheme>` (`DerivedTheme`). Any consumer that calls `useTheme()` and reads fields that exist on `ThemeTokens` but not on `DerivedTheme` (e.g. accent colors, border colors, or any custom token) will get a TypeScript error or `undefined` at runtime. This is a breaking API change shipped without a deprecation notice, changelog entry, or adapter.

Fix: either keep returning `ThemeTokens` from `useTheme()` and map `DerivedTheme` into it, or document the breaking change and bump the major version, or provide a compatibility shim.

**2. `deriveTheme` is called redundantly — twice at init, twice more in the effect**

In `AutoThemeProvider`:

```typescript
// Called once in useState initializer:
deriveTheme({ Normal: { fg: dark.fg, bg: dark.bg } })

// Then called again inside useEffect:
const derivedDark = deriveTheme({ Normal: { fg: dark.fg, bg: dark.bg } });
const derivedLight = deriveTheme({ Normal: { fg: light.fg, bg: light.bg } });
```

These are the same computations. Hoist the derived values outside the component body or memoize them so derivation runs once per theme change, not twice per render.

**3. OSC query writes a malformed escape sequence**

```typescript
process.stdout.write(`${ansi.OSC}${code};?`);
```

A well-formed OSC query ends with `ST` (String Terminator: `\x1b\\` or `\x07`). Without the terminator, some terminals buffer the sequence indefinitely rather than responding. The write should be:

```typescript
process.stdout.write(`\x1b]${code};?\x07`);
```

Verify that `ansi.OSC` is the correct escape prefix and add the `BEL` or `ST` terminator.

**4. `DerivedThemeRole` interface is exported but never used in `DerivedTheme`**

`DerivedTheme` defines each role inline rather than using `DerivedThemeRole`. The exported interface is therefore dead code. Either use it consistently in `DerivedTheme` or remove it from the export surface.

**5. `brighten` and `dim` call `parseColor` without validating the result**

```typescript
function brighten(color: string, amount = 0.15): string {
    const [r, g, b] = colorToRgb(parseColor(color));
```

`parseColor` returns `{ type: 'none' }` for invalid input, but `brighten` passes it straight to `colorToRgb`. The result will be `NaN` channel values. Since `brighten` is only called with colors already validated by `toHex`, this is safe today — but fragile. These helpers should either accept a pre-parsed color object or guard against `type === 'none'`.

**6. Test coverage gaps**

- No test for `brighten` or `dim` in isolation.
- No test for the OSC-based detection path (`detectDarkViaOsc`, `queryOscColor`, `parseOscColorResponse`).
- The `normalizes named colors` test asserts `theme.Normal.fg` is `'#aaaaaa'` for the input `'white'` — this pin-tests an internal color mapping detail. If `parseColor`'s ANSI color table changes, this test breaks for unrelated reasons. Pin the test to a behavior you control (e.g. a known `#rrggbb` input normalizing correctly) or at least add a comment explaining the expected mapping.

### What Looks Good

- The `derive.ts` module is cleanly separated with well-defined interfaces and clear pure functions.
- Color normalization via `toHex` is solid: it validates, converts, and formats in one place.
- The cancellation flag (`cancelled`) in the async `updateTheme` path correctly prevents stale state updates after unmount.
- Conventional commit title format is correct.
- PR scope is tight — no unrelated changes bundled in.

Update your PR to address the issues above and request re-review.

Karanjot786 (commented):
Re-review: No `build-and-test` CI check ran for this PR — only `star-check` and `path-labels`. Please push a small commit or close/reopen the PR to trigger the build-and-test workflow. Cannot approve without CI verification.

Karanjot786:
Good work on the `deriveTheme` utility — the logic is clean and the test coverage is solid. However the `AutoThemeProvider.ts` changes introduce two blocking problems.

## BLOCKING

**1. OSC terminal color query code belongs in a separate PR**

The new `queryOscColor()` directly manipulates `process.stdin.setRawMode`, reads raw bytes from stdin, and writes OSC escape sequences to stdout at component mount time. This cannot run in non-TTY environments (CI/pipes/tests) and has no test coverage. It also mixes two unrelated features into one PR (theme derivation + OSC colour detection).

Per `AGENTS.md`: *Ask first: Changing an existing public API signature.*

Please remove `queryOscColor`, `parseOscColorResponse`, `rgbComponentFrom4Hex`, and `detectDarkViaOsc` from this PR and open a separate issue + PR for OSC detection.

**2. `ThemeContext` default is a breaking change**

```ts
// Before — returns ThemeTokens
export const ThemeContext = createContext<ThemeTokens>(systemTheme);

// After — returns DerivedTheme (incompatible type)
export const ThemeContext = createContext(
    deriveTheme({ Normal: { fg: systemTheme.fg, bg: systemTheme.bg } })
);
```

Existing consumers typed as `ThemeTokens` will break. This needs explicit maintainer sign-off before landing.

## MINOR

**3. Named-color test expectation looks wrong**

```ts
expect(theme.Normal.fg).toBe('#aaaaaa'); // 'white' → #aaaaaa?
```

Standard white is `#ffffff`. If the internal named-color table maps `'white'` to `#aaaaaa`, document why. Otherwise fix the expected value.

---

**Action needed:** Remove the `AutoThemeProvider.ts` + `ThemeContext` changes. The `derive.ts`, `derive.test.ts`, and `index.ts` export are clean and ready to merge as-is.

PR #982, opened 2026-06-07
feat(widgets): add PinInput widget for capturing discrete code sequences

### LINKED ISSUE #978: [feature] feat(widgets): add PinInput widget for capturing discrete code sequences

When building CLI tools that require multi-factor authentication (MFA) or single-use verification codes, users typically need to input a standard 4-digit or 6-digit PIN. Using a standard `TextInput` works, but it isn't visually distinct or clear to the user exactly how many digits are required. TermUI currently lacks a specialized input component for capturing these discrete code sequences.

## Proposed solution

I propose building a `PinInput` widget in `@termuijs/widgets`. 

It will render a sequence of separate, boxed cells (e.g., `[ 1 ] [ 2 ] [ _ ] [ _ ]`). As the user types, the cursor focus automatically advances to the next empty block. 

```typescript
// Example usage
import { PinInput } from '@termuijs/widgets';

const pin = new PinInput('Enter 2FA Code', {
  length: 6,
  masked: false, // If true, replaces inputs with '*' or '•'
  onComplete: (code) => {
    console.log(`Verifying: ${code}`);
  }
});
```
## Alternatives considered

I considered just using the existing `TextInput` (or potentially the `MaskedInput` in `@termuijs/ui`), but these don't provide the distinct visual "blocks" per character that help guide users when entering fixed-length numeric codes.

`@termuijs/widgets`

## GSSoC contributor?

- [x] Yes. You contribute under GSSoC 2026 and want to work on this.
- [ ] No. You are filing a suggestion.

Satvik-art-creator:
## Description

This PR introduces the new `PinInput` widget for capturing discrete code sequences (like OTPs or 2FA codes). It renders visually isolated character blocks, auto-advances the cursor as the user types, handles bounded `left`/`right` keyboard navigation, processes `backspace` clearing elegantly, and natively supports masked (e.g., `•`) characters.

## Related Issue

Closes #978

## Which package(s)?

`@termuijs/widgets`

## Type of Change

- [ ] 🐛 Bug fix (`type:bug`)
- [ ] 📝 Docs (`type:docs`)
- [x] 🎨 Design / UX (`type:design`)

## Checklist

## GSSoC 2026 Participation

- Your GSSoC profile: `https://gssoc.girlscript.org/profile/172e996c-3943-4a86-8249-1f01922d0f64`

*(Reviewer: You can test the widget rendering interactively via tests, or I can provide screenshots of the visual `[ • ] [   ] [   ] [   ]` states if required).*

## Notes for the Reviewer

The implementation handles boundary conditions firmly to prevent out-of-bounds cursor index crashes. The rendering logic ensures that the individual cells are properly isolated by one space from each other with brackets (`[ x ]`) to form distinct blocks, utilizing standard cell attributes (like `inverse`) to denote the focused cell cleanly.

## Summary by CodeRabbit

* **New Features**
  * Added a PinInput widget for fixed-length PIN/OTP entry with optional masking, keyboard cursor navigation, per-character entry, and onChange/onComplete callbacks.
  * Screen now provides a read-only cell retrieval API to inspect back-buffer cells by coordinates.

* **Tests**
  * Added tests covering initialization, character entry, cursor movement, deletion/backspace behavior, completion callback, and masked vs visible rendering.

### DIFF
--- a/packages/core/src/terminal/Screen.ts
+++ b/packages/core/src/terminal/Screen.ts
@@ -139,6 +139,12 @@ export class Screen {
         this.back = this._createGrid(cols, rows);
     }
 
+    /** Retrieve a read-only copy of the cell at (x, y) from the back buffer. */
+    getCell(x: number, y: number): Readonly<Cell> | undefined {
+        if (x < 0 || x >= this._cols || y < 0 || y >= this._rows) return undefined;
+        return this.back[y][x];
+    }
+
     /** Serialize a back-buffer row to a plain string (skips continuation cells). */
     getLine(row: number): string {
         if (row < 0 || row >= this._rows) return '';
--- a/packages/core/src/terminal/Screen.ts
+++ b/packages/core/src/terminal/Screen.ts
@@ -141,7 +141,9 @@ export class Screen {
 
     /** Retrieve a read-only copy of the cell at (x, y) from the back buffer. */
     getCell(x: number, y: number): Readonly<Cell> | undefined {
-        if (x < 0 || x >= this._cols || y < 0 || y >= this._rows) return undefined;
+        x = Math.floor(x);
+        y = Math.floor(y);
+        if (!(x >= 0 && x < this._cols && y >= 0 && y < this._rows)) return undefined;
         return this.back[y][x];
     }
 
--- a/packages/widgets/src/index.ts
+++ b/packages/widgets/src/index.ts
@@ -46,6 +46,8 @@ export { RangeInput } from './input/RangeInput.js';
 export type { RangeInputOptions } from './input/RangeInput.js';
 export { Slider } from "./input/Slider.js";
 export type { SliderOptions } from "./input/Slider.js";
+export { PinInput } from "./input/PinInput.js";
+export type { PinInputOptions } from "./input/PinInput.js";
 
 // ── Data Widgets ──────────────────────────────────────
 export { Table } from './data/Table.js';
--- a/packages/widgets/src/input/PinInput.test.ts
+++ b/packages/widgets/src/input/PinInput.test.ts
@@ -90,9 +90,7 @@ describe("PinInput", () => {
     const screen = new Screen(40, 1);
     pin.render(screen);
 
-    const rendered = screen.back[0]
-      .map((c: { char: string }) => c.char)
-      .join("");
+    const rendered = Array.from({ length: 40 }, (_, i) => screen.getCell(i, 0)?.char ?? " ").join("");
 
     expect(rendered).toContain("[ 1 ]");
     expect(rendered).toContain("[ 2 ]");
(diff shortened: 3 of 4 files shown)


Karanjot786:
## Review

Thanks for contributing the `PinInput` widget. The overall structure is solid: `focusable = true` is set, `markDirty()` is called on every state mutation, key names follow the project convention (`left`, `right`, `backspace`), and `onChange`/`onComplete` callbacks are present and wired correctly. Two issues need fixing before this can merge.

### Issues to Fix

1. **`value` getter produces incorrect length when slots have gaps**

   The `value` getter joins `this._value` using `Array.join("")`, which skips empty strings. If a user navigates left and skips a slot, the resulting string can have fewer characters than `_length` even though the last slot is filled. The `onComplete` check relies on `currentVal.length === this._length`, so it can either fire too early or not at all.

   Replace the completion guard with a check that every slot is non-empty:

   ```ts
   get value(): string {
     return this._value.join("");
   }

   private get _isFull(): boolean {
     return this._value.every((v) => v !== "");
   }
   ```

   Then in `handleKey`, replace:
   ```ts
   } else if (currentVal.length === this._length) {
     this._onComplete?.(currentVal);
   }
   ```
   with:
   ```ts
   } else if (this._isFull) {
     this._onComplete?.(this.value);
   }
   ```

2. **Render test accesses `screen.back`, a private internal field**

   The test reads `screen.back[0]` to inspect rendered output. Project convention requires testing through the public API, not private or internal fields. Use whatever public method `Screen` exposes for reading cell content (for example `screen.getCell(x, y)` or a similar accessor), or check the `Screen` source for the correct public surface.

   ```ts
   // Before (accesses internal state)
   const rendered = screen.back[0].map((c: { char: string }) => c.char).join("");

   // After (use public API)
   const rendered = Array.from({ length: 40 }, (_, i) => screen.getCell(i, 0).char).join("");
   ```

   If `Screen` has no public cell accessor, add one or raise it as a separate issue rather than using the private field.

### Minor Notes

- The `blockStr` indentation inside `_renderSelf` uses 4 spaces while the rest of the file uses 2 spaces. Align to 2 spaces for consistency.
- Consider documenting the constructor signature `(style, opts)` with a JSDoc comment since the argument order differs from typical widget usage patterns in the codebase.

Fix the two issues above and this will be ready to merge.

Karanjot786 (approved):
Re-reviewed — LGTM. CI is green. Clean PinInput widget implementation. Tests properly use `createKeyEvent` from core (not `as any`). Widget handles input, auto-advance, backspace, onComplete callback, and renders correctly with screen assertions.

Note: This PR adds `getCell()` to `Screen` in core. Since PR rules say to ask first before modifying core, please confirm this is tied to a linked issue.

PR #988, opened 2026-06-07
fix(dev-server): avoid theme watcher debounce collisions

### LINKED ISSUE #987: [bug] ThemeWatcher debounce collisions suppress valid theme change events
## Which package?

@termuijs/dev-server

## What happened?

ThemeWatcher debounces file change events using only the emitted filename as the debounce key.

When multiple watched directories contain files with the same basename (for example `index.tss`), change events from different directories collide in the debounce map. One event cancels the other, causing a legitimate theme change notification to be dropped.

Expected behavior:

* Theme changes from different watched directories should be debounced independently.

Actual behavior:

* Only one change event is emitted when two watched directories report the same filename within the debounce window.

## Steps to reproduce

Create a ThemeWatcher that watches multiple directories:

```typescript
const watcher = new ThemeWatcher({
  watchDirs: ['./themes', './more-themes']
});
```

Create the following files:

```text
./themes/index.tss
./more-themes/index.tss
```

Modify both files within the debounce window.

Expected:

* Two independent theme change events are emitted.

Actual:

* Only one change event is emitted.

A focused regression test reproduces the issue:

```typescript
it('does not collide when different watched dirs contain the same filename', () => {
  // two watched directories both emit index.tss
  // expected: two independent change events
  // actual: one event is dropped due to debounce key collision
});
```

## Environment

* Bun version: 1.x
* OS: Linux (WSL)
* Terminal emulator: Windows Terminal
* TermUI version: current main branch

## Screenshots or terminal output

Failing regression test before the fix:

```text
FAIL theme-watcher.test.ts > ThemeWatcher > does not collide when different watched dirs contain the same filename

AssertionError: expected "spy" to be called 2 times, but got 1 times
```

## GSSoC contributor?

* [x] Yes. You contribute under GSSoC 2026.
* [ ] No.

ZainabTravadi:
## Description

Fixes a debounce key collision in ThemeWatcher when multiple watched directories contain files with the same basename.

This PR uses a directory-aware debounce key internally while preserving the existing public API, ensuring valid theme change events are not dropped. It also adds regression coverage reproducing and preventing the issue.

## Related Issue

Closes #987 

## Which package(s)?

* @termuijs/dev-server

## Type of Change

* [x] 🐛 Bug fix (`type:bug`)
* [x] 🧪 Tests (`type:testing`)

## Checklist

* [x] ⭐ You starred the repo. The `needs-star` check blocks your merge otherwise.
* [x] Tests pass locally: `bun vitest run`
* [ ] Build passes: `bun run build`
* [ ] Typecheck passes: `bun run typecheck`
* [x] You read [`CONTRIBUTING.md`](./CONTRIBUTING.md).
* [x] Your PR title follows `type: short description`.
* [ ] Widget state mutators call `markDirty()` (if your change affects rendering).
* [x] No new `any` types without an inline comment explaining why.
* [x] No unrelated refactors bundled into this PR.

## GSSoC 2026 Participation

* [x] You are a GSSoC 2026 contributor.
* Your GSSoC profile: `https://gssoc.girlscript.org/profile/3652c85d-e890-41f5-a48f-584ed0a7f209`

N/A – bug fix and regression test coverage only.

## Notes for the Reviewer

* Root cause: debounce timers were keyed only by the basename reported by fs.watch, causing collisions when different watched directories emitted the same filename (for example, index.tss).
* The fix uses a directory-resolved path as the internal debounce key while continuing to emit the original `filename` to consumers.
* Added a focused regression test reproducing the collision scenario and verifying that independent change events are emitted.
* No public API changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved incorrect debouncing when watching multiple directories so identical filenames in different folders no longer interfere with each other; file change notifications are now per-path.

* **Tests**
  * Added test coverage for multi-directory watch scenarios to ensure change events from different directories are handled independently.

### DIFF
--- a/packages/dev-server/src/theme-watcher.test.ts
+++ b/packages/dev-server/src/theme-watcher.test.ts
@@ -52,4 +52,27 @@ describe('ThemeWatcher', () => {
         expect(arg).toHaveProperty('filename', 'styles.tss');
         expect(arg).toHaveProperty('timestamp');
     });
+
+    it('does not collide when different watched dirs contain the same filename', () => {
+        // Simulate two separate watchers (different directories) both reporting
+        // a change for the same basename. They should produce two onChange events.
+        const emitters = [new EventEmitter(), new EventEmitter()];
+        let callCount = 0;
+        vi.mocked(watch).mockImplementation(() => emitters[callCount++] as any);
+
+        const watcher = new ThemeWatcher({ watchDirs: ['./themes', './more-themes'] });
+        const changeSpy = vi.fn();
+
+        watcher.onChange(changeSpy);
+        watcher.start();
+
+        // Both directories emit a change for "index.tss"
+        emitters[0].emit('change', 'change', 'index.tss');
+        emitters[1].emit('change', 'change', 'index.tss');
+
+        vi.advanceTimersByTime(100);
+
+        // Expect two separate events (one per directory/file), not one.
+        expect(changeSpy).toHaveBeenCalledTimes(2);
+    });
 });
--- a/packages/dev-server/src/theme-watcher.test.ts
+++ b/packages/dev-server/src/theme-watcher.test.ts
@@ -58,7 +58,11 @@ describe('ThemeWatcher', () => {
         // a change for the same basename. They should produce two onChange events.
         const emitters = [new EventEmitter(), new EventEmitter()];
         let callCount = 0;
-        vi.mocked(watch).mockImplementation(() => emitters[callCount++] as any);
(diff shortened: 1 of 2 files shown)


ZainabTravadi:
Addressed.

Added an inline justification comment explaining why the `as any` cast is required in this test. The mock uses a simplified `EventEmitter` fixture that provides the `on`/`emit` behavior needed for the test, while `watch()` is typed to return `FSWatcher`.

No behavior changes were made. Re-ran the focused test and it continues to pass.

PR #995, opened 2026-06-07
fix(core): reject and clear pending cursor requests on InputParser.stop() (closes #993)

### LINKED ISSUE #993: [BUG] InputParser.stop() leaves pending cursor position requests hanging forever
## Description
`InputParser.stop()` in `packages/core/src/input/InputParser.ts` clears the data handler and escape buffer but never rejects or clears `_cursorRequests` or their associated timeouts. Any pending `requestCursorPosition()` promises hang until their individual 200ms timeout fires, and the timeout handles leak until then.

## Affected Package
`@termuijs/core`

## Affected File
`packages/core/src/input/InputParser.ts`

## Affected Function
`InputParser.stop()`

## Steps to Reproduce
```ts
const parser = new InputParser(terminal);
parser.start();
const pos = parser.requestCursorPosition(); // pending promise
parser.stop(); // tears down input but never rejects the promise
await pos; // hangs for 200ms then rejects via timeout — handle leaked until then
```

## Expected Behavior
`stop()` should immediately reject all pending `_cursorRequests` with a descriptive error and clear their timeout handles to avoid leaking resources.

## Suggested Fix
In `stop()`, iterate `_cursorRequests`, call `clearTimeout` on each entry's timeout handle, reject the promise, then reset the array:
```ts
for (const req of this._cursorRequests) {
    clearTimeout(req.timeout);
    req.reject(new Error('InputParser stopped'));
}
this._cursorRequests = [];
```

## Additional Context
`_cursorRequests` and their timeouts are set up in `requestCursorPosition()` but the cleanup path in `stop()` does not handle them, leaving dangling promises and timer handles on shutdown.

Pranaykarvi:
## Problem
`InputParser.stop()` cleared the data handler and escape buffer but never rejected or cleared `_cursorRequests` or their timeout handles. Pending `requestCursorPosition()` promises hung until their 200ms timeout fired, leaking timer handles on every shutdown.

## Fix
In `stop()`, after resetting the escape buffer, iterate `_cursorRequests`, call `clearTimeout` on each timeout handle, reject with `'InputParser stopped'`, and reset the array.

## Changes
- `packages/core/src/input/InputParser.ts` — cleanup loop added to `stop()`
- `packages/core/src/input/InputParser.test.ts` — test verifying immediate rejection with correct message when `stop()` is called

## Verification
- `bun vitest run packages/core` — 280 passed
- `bun run typecheck` — 27 tasks passed

Closes #993

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where pending cursor position requests were not properly terminated when stopping the input parser.

### DIFF
--- a/packages/core/src/input/InputParser.test.ts
+++ b/packages/core/src/input/InputParser.test.ts
@@ -134,6 +134,18 @@ describe('InputParser', () => {
         await expect(positionPromise).rejects.toThrow('Cursor position request timed out');
     });
 
+    it('rejects pending cursor position requests immediately when stop() is called', async () => {
+        vi.useFakeTimers();
+        const stdin = createMockStdin();
+        const parser = new InputParser(stdin);
+        parser.start();
+
+        const positionPromise = parser.requestCursorPosition(200);
+        parser.stop();
+
+        await expect(positionPromise).rejects.toThrow('InputParser stopped');
+    });
+
     it('does not emit a key event for cursor position reports', async () => {
         const stdin = createMockStdin();
         const parser = new InputParser(stdin);
--- a/packages/core/src/input/InputParser.ts
+++ b/packages/core/src/input/InputParser.ts
@@ -98,6 +98,11 @@ export class InputParser {
             this._escapeTimeout = null;
         }
         this._escapeBuffer = Buffer.alloc(0);
+        for (const req of this._cursorRequests) {
+            clearTimeout(req.timeout);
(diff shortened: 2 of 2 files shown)


Pranaykarvi:
Hi @Karanjot786, fixed issue #993 — stop() now iterates _cursorRequests, clears each timeout handle, and rejects the promise with 'InputParser stopped' before resetting the array. Only InputParser.ts and its test file were modified. 280 tests pass and typecheck is clean.

PR #997, opened 2026-06-07
test(core): improve App.mount() + unmount() test to assert exit code directly (closes #996)

### LINKED ISSUE #996: [BUG] App.mount() promise never resolves when unmount() is called directly
## Description
`App.mount()` in `packages/core/src/app/App.ts` stores `_exitResolve` and only calls it from `exit()`. Calling `unmount()` directly — which is common in tests, wrappers, and custom shutdown flows — tears down input and renderers but leaves the `mount()` promise pending forever.

## Affected Package
`@termuijs/core`

## Affected File
`packages/core/src/app/App.ts`

## Affected Function
`App.mount()`, `App.unmount()`

## Steps to Reproduce
```ts
const app = new App();
const mountPromise = app.mount(rootWidget);
app.unmount(); // tears down everything
await mountPromise; // hangs forever — never resolves or rejects
```

## Expected Behavior
`unmount()` should resolve the `mount()` promise (with exit code 0) so callers can safely `await app.mount(...)` and have it settle when the app shuts down regardless of whether `exit()` or `unmount()` was called.

## Impact
- Tests that `await app.mount()` hang forever unless they call `exit()` specifically
- Wrapper libraries that call `unmount()` on cleanup leak the pending promise
- Any `finally` block after `await mount()` never runs on `unmount()`-based shutdown

## Suggested Fix
In `unmount()`, resolve `_exitResolve` with code `0` if it exists:
```ts
unmount(): void {
    // existing teardown logic unchanged
    this._exitResolve?.(0);
    this._exitResolve = undefined;
}
```

## Additional Context
`exit()` already calls `_exitResolve` correctly. This fix brings `unmount()` into parity so both shutdown paths settle the `mount()` promise.

Pranaykarvi:
## Context
The underlying fix for issue #996 was already merged in PR #778 — `unmount()` correctly resolves the `mount()` promise with exit code 0.

However the existing test used a fragile `Promise.race` with a 500ms timeout to verify this behavior, which can produce false positives if the promise resolves for the wrong reason.

## Change
Replaced the `Promise.race` timeout pattern with a direct `await mountPromise` and an explicit `expect(code).toBe(0)` assertion, making the test deterministic and self-documenting.

## Verification
- `bun vitest run packages/core` — 279 passed
- `bun run typecheck` — 27 tasks passed

Closes #996

## Release Notes

This release contains no user-facing changes. Internal testing infrastructure improvements were made to streamline the application lifecycle verification process.

* **Tests**
  * Improved test reliability and simplified assertion strategy for app lifecycle verification.

### DIFF
--- a/packages/core/src/app/App.test.ts
+++ b/packages/core/src/app/App.test.ts
@@ -32,7 +32,7 @@ function createMockRootWidget(): RootWidget {
 
 describe('App', () => {
     describe('unmount()', () => {
-        it('mount() promise resolves when unmount() is called directly', async () => {
+        it('mount() resolves when unmount() is called directly', async () => {
             const root = createMockRootWidget();
             const fakeStdout: any = { // minimal stdout stub — full NodeJS.WriteStream type not required here
                 writes: '',
@@ -58,12 +58,8 @@ describe('App', () => {
 
             app.unmount();
 
-            const result = await Promise.race([
-                mountPromise.then(() => 'resolved'),
-                new Promise<string>((r) => setTimeout(() => r('timeout'), 500)),
-            ]);
-
-            expect(result).toBe('resolved');
+            const code = await mountPromise;
+            expect(code).toBe(0);
         });
     });
 


Pranaykarvi:
Hi @Karanjot786, the underlying fix for #996 was already merged in PR #778. This PR improves the existing test — replaced the fragile Promise.race timeout pattern with a direct await and explicit exit code assertion. 279 tests pass and typecheck is clean.

PR #1041, opened 2026-06-07
fix(core): RenderHook stdout hijack - static _originalWrite, re-entrancy guard, globalRestore

### LINKED ISSUE #1037: [bug] Process-Wide `process.stdout.write` Hijack via RenderHook Causes Global Instability, Data Loss, and Breaks on Hot-Reload
## Problem

The `RenderHook` class globally replaces `process.stdout.write` with a monkey-patched version to intercept console output. This approach has several interacting critical flaws.

### A) Loses original reference on double `start()` without `stop()`

**File:** `packages/core/src/renderer/render-hook.ts`, lines 15-30

```typescript
start(): void {
    if (this._isActive) return;
    this._isActive = true;
    this._originalWrite = process.stdout.write;

    process.stdout.write = (chunk, encodingOrCb?, cb?): boolean => {
        // ... intercept logic
    };
}
```

If `start()` is called twice, the second call captures the FIRST hook's patched `write` as `_originalWrite`, not the real `process.stdout.write`. Calling `stop()` then restores the first hook's wrapper instead of the genuine stdout, leaving the system in a permanently broken state.

### B) Crash during flush permanently disables stdout

**File:** `packages/core/src/terminal/Renderer.ts`, lines 117-127

```typescript
const isHookActive = this.hook.isActive;
if (isHookActive) this.hook.stop();
if (bufferedLogs) this._terminal.write(bufferedLogs);
this._terminal.write(output);  // <-- If this throws (EPIPE, EBADF)
if (isHookActive) this.hook.start();  // <-- This is NEVER reached
```

If `this._terminal.write(output)` throws (terminal closes, pipe breaks, write buffer full), the hook is never restarted. Because `_originalWrite` was already nulled by `stop()`, any future `process.stdout.write` call silently drops all output.

### C) Hot-reload creates orphaned hooks

**File:** `packages/dev-server/src/`

Hot-reload creates new `App` instances without reliably calling `hook.stop()` on the old instance first. The `_originalWrite` chain is corrupted as described in (a).

### D) Module-level global state with no isolation

`_originalWrite` is a simple class property with no protection against concurrent access. If two App instances exist simultaneously (e.g., during a rapid hot-reload cycle), their hooks trample each other's state.

## Impact

- **Critical**: All `console.log` output is silently lost when a render error occurs during flushing
- **Critical**: Third-party libraries writing to stdout (debug loggers, profiling tools, health checks) have their output permanently truncated
- **High**: Hot-reload cycles and rapid App restarts can permanently corrupt stdout, requiring a hard process restart
- **High**: Debugging production issues becomes impossible because application logs disappear silently

1. **`packages/core/src/renderer/render-hook.ts`** — Core fixes:
   - Guard against double-start: if `_isActive`, throw `Error('RenderHook already active')`
   - Capture `_originalWrite` at module load time (static reference), not at `start()`
   - Add `restore()` as a static fallback: `RenderHook.globalRestore()` that can repair stdout from any state

2. **`packages/core/src/terminal/Renderer.ts`** — Wrap `_flush()` output write in try/finally that ALWAYS restarts the hook:
   ```typescript
   try {
       this._terminal.write(output);
   } finally {
       if (isHookActive) this.hook.start();
   }
   ```

3. **`packages/core/src/app/App.ts`** — Ensure proper hook cleanup in all exit paths:
   - Call `hook.stop()` in `unmount()` before any new App instance starts
   - Add `finally` block in the mount lifecycle

4. **`packages/dev-server/src/`** — Add hot-reload lifecycle coordination that awaits old App unmount before creating new App.

5. **Move to stream transformer** — Consider replacing the global monkey-patch with a stream transformer/passthrough that sits between the App and stdout, avoiding global monkey-patching entirely.

ionfwsrijan:
### Description

Prevents RenderHook from permanently hijacking process.stdout.write when multiple instances are created and destroyed. Makes _originalWrite static so all instances share a single reference, wraps _flush() in try/finally to balance stop/start, and adds SIGINT/SIGTERM cleanup in App.ts.

### Related Issue

Closes #1037

### Which package(s)?

@termuijs/core

### Type of Change

- [x] 🐛 Bug fix (`type:bug`)

### Checklist

- [x] ⭐ You starred the repo.

### GSSoC 2026 Participation

N/A — internal refactor, no UI change.

### Notes for the Reviewer

The key insight is that `_originalWrite` was an instance property, so each RenderHook saved its own reference. When hook A was created then destroyed, it restored stdout to its own `_originalWrite` which might already be a patched version from hook B. Making it static ensures all instances share one canonical reference to the true `process.stdout.write`.

## Summary by CodeRabbit

* **New Features**
  * Application now gracefully handles interrupt and termination signals.

* **Bug Fixes**
  * Improved robustness of output rendering when handling multiple concurrent operations and error cases.

* **Tests**
  * Expanded test coverage for signal handling, output rendering, and error resilience.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -6,14 +6,27 @@
       "name": "termui",
       "dependencies": {
         "commander": "^15.0.0",
+        "dotenv": "^16.0.0",
       },
       "devDependencies": {
         "@types/bun": "latest",
         "@types/node": "^25.2.3",
         "tsup": "^8.3.0",
-        "turbo": "^2.9.16",
+        "turbo": "^2.9.12",
         "typescript": "^5.7.0",
         "vitest": "^2.1.0",
+      },
+    },
+    "examples/ai-assistant": {
+      "name": "@termuijs/example-ai-assistant",
+      "version": "0.1.0",
+      "dependencies": {
+        "@termuijs/adapters": "workspace:*",
+        "@termuijs/core": "workspace:*",
+        "@termuijs/jsx": "workspace:*",
+        "@termuijs/tss": "workspace:*",
+        "@termuijs/ui": "workspace:*",
+        "@termuijs/widgets": "workspace:*",
       },
     },
     "examples/ai-streaming": {
@@ -89,8 +102,26 @@
         "@termuijs/data": "workspace:*",
         "@termuijs/widgets": "workspace:*",
       },
+    },
+    "examples/data-grid": {
+      "name": "data-grid-example",
+      "version": "0.0.1",
+      "dependencies": {
+        "@termuijs/jsx": "workspace:*",
+        "@termuijs/widgets": "workspace:*",
+      },
+    },
+    "examples/db-browser": {
+      "name": "@termuijs/example-db-browser",
+      "version": "0.1.0",
+      "dependencies": {
+        "@termuijs/core": "workspace:*",
+        "@termuijs/jsx": "workspace:*",
+        "@termuijs/ui": "workspace:*",
+        "@termuijs/widgets": "workspace:*",
+      },
       "devDependencies": {
-        "rimraf": "^6.1.3",
+        "typescript": "^5.4.0",
       },
     },
     "examples/file-manager": {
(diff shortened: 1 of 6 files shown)


ionfwsrijan:

Karanjot786 (changes requested):
Thanks for the RenderHook fix. Please update before merge.

- Add a process-wide active owner guard in `RenderHook.start()`. With this patch, hook A and hook B both start. If hook A stops first, stdout restores while hook B still thinks active.
- Add tests for two `RenderHook` instances: A start, B start, A stop, B stop. Assert `process.stdout.write` returns to the original and logs are preserved.
- Add a test where `_terminal.write()` throws during render flush. Assert the hook restarts in `finally`.
- Keep App signal cleanup scoped, or explain why duplicate Terminal and App SIGINT handlers will not double-run cleanup.

Run `bun vitest run packages/core` and `bun run typecheck` after the update.

ionfwsrijan:

Karanjot786 (approved):
LGTM. CI is green. Solid fix making `_originalWrite` static to prevent multiple hijacks, adding re-entrancy guard, and providing `globalRestore` for cleanup.

PR #1042, opened 2026-06-07
fix(data): useWebSocket generation counter, close on error, stale callback guard

### LINKED ISSUE #1038: [bug] `useWebSocket` Cascading Reconnect Chain on URL Change or Unmount Causes Infinite WebSocket Creation and Resource Exhaustion
## Problem

The `useWebSocket` hook has a race condition where cleanup and reconnect handlers interact destructively, creating an infinite cascade of WebSocket connections and timer leaks.

**File:** `packages/data/src/hooks.ts`, lines 260-322

```typescript
export function useWebSocket(url: string): UseWebSocketReturn {
    const [message, setMessage] = useState<string | null>(null);
    const [state, setState] = useState<WebSocketState>('connecting');
    const socketRef = useRef<WebSocket | null>(null);
    const retryCountRef = useRef(0);

    useEffect(() => {
        let isMounted = true;

        function connect() {
            const socket = new WebSocket(url);
            socketRef.current = socket;

            socket.onclose = () => {
                if (!isMounted) return;
                setState('closed');
                const timeout = Math.min(1000 * Math.pow(2, retryCountRef.current), 10000);
                retryCountRef.current += 1;
                reconnectTimeoutRef.current = setTimeout(() => {
                    connect();  // <-- captures old 'connect' reference
                }, timeout);
            };
        }

        connect();

        return () => {
            isMounted = false;
            if (reconnectTimeoutRef.current) {
                clearTimeout(reconnectTimeoutRef.current);
            }
            if (socketRef.current) {
                socketRef.current.close();  // <-- TRIGGERS onclose
            }
        };
    }, [url]);
}
```

### The cascade mechanism

1. Component renders with `url = "ws://example.com/1"`. `useEffect` runs, calls `connect()`, creates WebSocket A.
2. `url` changes to `"ws://example.com/2"` (or component unmounts).
3. Cleanup runs: sets `isMounted = false`, clears the reconnect timeout, calls `socketRef.current.close()` on WebSocket A.
4. WebSocket A's `onclose` fires ASYNCHRONOUSLY. The `isMounted` flag catches this and returns — correct behavior.
5. **But** — if `close()` fires SYNCHRONOUSLY (possible in some environments), `onclose` runs INSIDE the cleanup function. It schedules `setTimeout(() => connect(), timeout)`. This new timeout calls `connect()`, creating NEW WebSocket B, which sets `socketRef.current = B`. The cleanup function returns, but `socketRef.current` now points to leaked socket B.
6. Socket B eventually closes, triggering another `onclose`, which schedules another `connect()`, creating Socket C — ad infinitum.

### Additional issues

- **`socket.onerror`** (line 312) does NOT close the socket. Connection errors are silently swallowed with no reconnect.
- **`retryCountRef`** is NOT reset when `url` changes, so exponential backoff picks up from an arbitrary count.

## Impact

- **Critical**: Every URL change creates an infinite cascade of WebSocket connections that are never GC'd → file descriptor exhaustion → process crash
- **Critical**: Memory leak from accumulated timer references that are never cleared
- **High**: Connection errors are silently swallowed with no reconnect attempt
- **Medium**: Exponential backoff state persists across URL changes, causing new connections to start with artificially long delays

1. **`packages/data/src/hooks.ts`** — Complete rewrite of `useWebSocket` reconnect logic:
   - Replace the boolean `isMounted` with a generation counter. Each effect run increments `generationRef`. All async callbacks check `if (myGen !== generationRef.current) return;` to handle stale closures
   - Defensively close the new socket if one is created during cleanup by checking `socketRef.current` after `oldSocket.close()`
   - Add `socket.onerror` that closes the socket: `socket.close()` so the error triggers reconnect via `onclose`
   - Reset `retryCountRef.current = 0` inside the effect (before `connect()`) so it resets on URL change
   - Clear `reconnectTimeoutRef` before setting a new one in `onclose` to prevent timer accumulation

2. **`packages/data/src/useWebSocket.test.tsx`** — Add tests for:
   - URL change mid-connection
   - Rapid URL changes (3+ in sequence)
   - Immediate unmount after mount
   - Error recovery (server goes down, comes back)
   - Cleanup after the component is unmounted while a reconnect is scheduled

ionfwsrijan:
### Description

Prevents stale callbacks from operating on superseded WebSocket instances during rapid reconnections. Adds a generation counter incremented before each connect attempt, explicit socket.close() on error to ensure reconnect fires, and guards on all event handlers that skip processing if the generation no longer matches.

### Related Issue

Closes #1038

### Which package(s)?

@termuijs/data

### Type of Change

- [x] 🐛 Bug fix (`type:bug`)

### Checklist

- [x] ⭐ You starred the repo.

### GSSoC 2026 Participation

N/A — internal state management, no UI change.

### Notes for the Reviewer

The generation counter pattern avoids the complexity of tracking pending async operations or implementing a full state machine. Each generation is a monotonically increasing integer; when a handler fires, it compares its captured generation against the ref's current value. Any mismatch means the handler is stale and should bail out. The `socket.close()` call on the error handler was missing, which meant certain error paths never triggered the close handler's reconnect logic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed WebSocket hook to properly clean up event handlers and reconnect timers when components unmount or connection URLs change, preventing memory leaks and stale state updates.

* **Tests**
  * Added comprehensive test coverage for WebSocket reconnection behavior, timer management, and cleanup scenarios during component lifecycle changes.

### DIFF
--- a/packages/data/src/hooks.ts
+++ b/packages/data/src/hooks.ts
@@ -264,6 +264,7 @@ export function useWebSocket(url: string): UseWebSocketReturn {
     useEffect(() => {
         let isMounted = true;
         const thisGeneration = ++generationRef.current;
+        retryCountRef.current = 0;
 
         function connect() {
             if (socketRef.current) {
@@ -295,6 +296,11 @@ export function useWebSocket(url: string): UseWebSocketReturn {
             socket.onclose = () => {
                 if (!isMounted || thisGeneration !== generationRef.current) return;
                 setState('closed')
+
+                if (reconnectTimeoutRef.current) {
+                    clearTimeout(reconnectTimeoutRef.current);
+                    reconnectTimeoutRef.current = null;
+                }
 
                 const timeout = Math.min(1000 * Math.pow(2, retryCountRef.current), 10000);
                 retryCountRef.current += 1;
--- a/packages/data/src/hooks.ts
+++ b/packages/data/src/hooks.ts
@@ -259,42 +259,56 @@ export function useWebSocket(url: string): UseWebSocketReturn {
     const socketRef = useRef<WebSocket | null>(null)
     const reconnectTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null)
     const retryCountRef = useRef(0)
+    const generationRef = useRef(0)
 
     useEffect(() => {
         let isMounted = true;
+        const thisGeneration = ++generationRef.current;
 
         function connect() {
+            if (socketRef.current) {
+                socketRef.current.onopen = null;
+                socketRef.current.onmessage = null;
+                socketRef.current.onclose = null;
+                socketRef.current.onerror = null;
+                if (socketRef.current.readyState === WebSocket.OPEN ||
+                    socketRef.current.readyState === WebSocket.CONNECTING) {
+                    socketRef.current.close();
+                }
+            }
+
             const socket = new WebSocket(url);
             socketRef.current = socket;
             setState('connecting')
 
             socket.onopen = () => {
-                if (!isMounted) return;
+                if (!isMounted || thisGeneration !== generationRef.current) return;
                 setState('open');
                 retryCountRef.current = 0;
             }
 
             socket.onmessage = (e) => {
-                if (!isMounted) return;
+                if (!isMounted || thisGeneration !== generationRef.current) return;
                 setMessage(e.data)
             }
 
             socket.onclose = () => {
-                if (!isMounted) return;
+                if (!isMounted || thisGeneration !== generationRef.current) return;
                 setState('closed')
 
                 const timeout = Math.min(1000 * Math.pow(2, retryCountRef.current), 10000);
                 retryCountRef.current += 1;
 
                 reconnectTimeoutRef.current = setTimeout(() => {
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

Karanjot786 (changes requested):
Thanks for tackling the reconnect leak. Please update before merge.

- Reset `retryCountRef.current = 0` at the start of each effect run. A new URL should not inherit the old exponential backoff count.
- Clear `reconnectTimeoutRef.current` before assigning a new timeout in `onclose`. The issue calls out timer accumulation.
- Add tests in `packages/data/src/useWebSocket.test.tsx` for URL change mid-connection, rapid URL changes, unmount with a pending reconnect, and error recovery.
- Keep the stale-generation guard. The pattern is the right direction.

Run `bun vitest run packages/data` and `bun run typecheck` after the update.

ionfwsrijan:

Karanjot786 (approved):
LGTM. CI is green. Good fixes: generation counter prevents stale callbacks, close-on-error prevents zombie connections, and callback guard prevents stale handler invocation.

PR #1043, opened 2026-06-07
fix(core): Terminal restore re-entrancy guard, copy-before-iterate for handlers

### LINKED ISSUE #1039: [bug] Terminal `restore()` Signal Handler Re-Entrancy and Cleanup Race Conditions Cause Terminal State Corruption
## Problem

Multiple interacting race conditions in `Terminal.restore()` and `_setupCleanup()` can leave the terminal in a permanently corrupted state.

### A) Re-entrant signal handlers cause double-cleanup

**File:** `packages/core/src/terminal/Terminal.ts`, lines 230-233

```typescript
private _setupCleanup(): void {
    const runCleanupHandlers = () => {
        for (const handler of this._cleanupHandlers) {
            try { handler(); } catch { /* swallow */ }
        }
        this.restore();
    };

    this._sigintHandler = () => { runCleanupHandlers(); process.exit(130); };
}
```

If SIGINT is received while `runCleanupHandlers` is iterating `_cleanupHandlers` (e.g., during a first SIGINT), the handler fires again. The `_cleanupHandlers` array is iterated a second time while the first iteration is still in progress. If a cleanup handler modifies the array, this causes concurrent modification. `restore()` is called twice — it returns early the second time due to `_restored` flag, but cleanup handlers ran TWICE, sending duplicate ANSI escape sequences.

### B) `process.exit()` during signal handler causes incomplete terminal restoration

```typescript
this._sigintHandler = () => { runCleanupHandlers(); process.exit(130); };
```

After `runCleanupHandlers()` calls `restore()`, the method calls `this.write(ansi.reset)`. If `process.exit()` is called before the output buffer is flushed, the terminal reset sequence may never arrive, leaving the terminal with corrupted state.

### C) `uncaughtException` handler can throw without restoring

```typescript
this._uncaughtExceptionHandler = (err: Error) => {
    this.restore();  // <-- if this throws...
    process.exit(1);  // exits without proper restoration
};
```

If `this.restore()` throws (e.g., `stdout.write()` fails), the exception goes uncaught (inside `uncaughtException`), causing immediate process termination with no terminal restoration.

### D) `process.on('exit')` handlers must not allocate

```typescript
this._exitHandler = runCleanupHandlers;
process.on('exit', this._exitHandler);
```

Node.js states exit listeners should only perform synchronous operations. This calls `restore()` → `this.write(ansi.reset)` → `stdout.write()`. The full cleanup chain involves iterating arrays, clearing timeouts, and calling user functions — potentially allocating or calling async APIs.

### E) Multiple signal registrations without removal

If `restore()` is called and then a new `_setupCleanup()` is called (on a new Terminal instance), the old signal handlers remain registered. The old handlers still reference the old Terminal instance, preventing GC and potentially corrupting the new instance's state.

## Impact

- **Critical**: Terminal left in corrupted state on double-SIGINT or exception during cleanup — cursor hidden, raw mode active, alternate screen not exited. User must manually run `reset` or restart terminal emulator.
- **High**: User-registered cleanup handlers via `onCleanup()` may run twice, causing double-side-effect issues (double file close, double resource release).
- **High**: Output buffer may not flush before `process.exit()`, losing the terminal reset ANSI sequence.
- **Medium**: If `stdout.write` fails during `restore()`, terminal state cannot be recovered.

1. **`packages/core/src/terminal/Terminal.ts`** — Multiple fixes:
   - **Re-entrancy guard**: Add `_isRestoring` flag checked at the top of `restore()`:
     ```typescript
     if (this._restored || this._isRestoring) return;
     this._isRestoring = true;
     ```
   - **Copy array before iteration**: `const handlers = [...this._cleanupHandlers];` to prevent concurrent modification
   - **Flush stdout after writing reset sequences**: Call `this.stdout._flush()` or `fs.writeSync(this.stdout.fd, ...)` after the ANSI reset
   - **Try/catch in uncaughtException handler**: Add a last-resort fallback:
     ```typescript
     try { this.restore(); } catch {
         fs.writeSync(process.stdout.fd, '\x1b[?25h\x1b[0m\x1b[?1049l');
     }
     ```
   - **Remove `process.on('exit')` handler** and rely only on signal handlers (SIGINT, SIGTERM)
   - **Unregister old signal handlers** before registering new ones

2. **`packages/core/src/app/App.ts`** — Ensure App-level exit paths use the guarded cleanup and don't double-call terminal methods.

3. **Add static emergency restore**: `Terminal.emergencyRestore()` that writes raw ANSI escape sequences via `fs.writeSync` to restore cursor visibility, reset attributes, and leave alternate screen, even when individual instances are corrupted.

4. **Tests** — Add tests for: double-SIGINT, exception during cleanup, output buffer flushing, and signal handler lifecycle across multiple Terminal instances.

ionfwsrijan:
### Description

Prevents stack overflow and double-free when Terminal.restore() is called re-entrantly from within a resize or cleanup handler. Adds a _restoring boolean flag for re-entrancy detection, copies handler arrays before iterating, and writes directly to stdout during restore to bypass the custom write function.

### Related Issue

Closes #1039

### Which package(s)?

@termuijs/core

### Type of Change

- [x] 🐛 Bug fix (`type:bug`)

### Checklist

- [x] ⭐ You starred the repo.

### GSSoC 2026 Participation

N/A — internal state management, no UI change.

### Notes for the Reviewer

The re-entrancy occurs because restore() emits a final resize event, which triggers resize handlers, which might call restore() again. The _restoring flag short-circuits the second call. Copy-before-iterate on _resizeHandlers and _cleanupHandlers prevents concurrent modification issues if a handler removes itself from the array. The direct stdout write bypass in restore avoids recursive calls through the custom _write function.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal restoration stability by preventing double-restore scenarios and safely handling handler mutations during cleanup operations. Enhanced resilience when terminal reset or exit sequences encounter errors.

### DIFF
--- a/packages/core/src/terminal/Terminal.test.ts
+++ b/packages/core/src/terminal/Terminal.test.ts
@@ -158,4 +158,88 @@ describe('Terminal', () => {
             expect(handler).not.toHaveBeenCalled();
         });
     });
+
+    describe('Reentrant restore', () => {
+        let term: Terminal;
+        let fakeStdout: EventEmitter & { columns: number; rows: number; write: any };
+
+        beforeEach(() => {
+            vi.useFakeTimers();
+            fakeStdout = Object.assign(new EventEmitter(), {
+                columns: 80,
+                rows: 24,
+                write: vi.fn()
+            });
+            term = new Terminal({
+                stdout: fakeStdout as unknown as NodeJS.WriteStream,
+                resizeDebounceMs: 16
+            });
+        });
+
+        afterEach(() => {
+            vi.restoreAllMocks();
+            vi.useRealTimers();
+            try { term.restore(); } catch { /* already restored */ }
+        });
+
+        it('reentrant restore from cleanup handler does not stack overflow', () => {
+            const handler = vi.fn(() => { term.restore(); });
+            term.onCleanup(handler);
+            // Run cleanup handlers directly to simulate the exit handler path
+            const cleanupFns = (term as any)._cleanupHandlers;
+            for (const fn of [...cleanupFns]) { fn(); }
+            expect(handler).toHaveBeenCalledTimes(1);
+        });
+
+        it('reentrant restore from resize handler is safe', () => {
+            const handler = vi.fn(() => { term.restore(); });
+            term.onResize(handler);
+            // Trigger resize handler before restore completes
+            term['_resizeHandlers']?.forEach((h: any) => h());
+            expect(() => term.restore()).not.toThrow();
+        });
+
+        it('cleanup handler can remove itself without crashing', () => {
+            const handler = vi.fn(() => {
+                const arr = (term as any)._cleanupHandlers;
+                const idx = arr.indexOf(handler);
+                if (idx >= 0) arr.splice(idx, 1);
+            });
+            term.onCleanup(handler);
+            const cleanupFns = (term as any)._cleanupHandlers;
+            for (const fn of [...cleanupFns]) { fn(); }
+            expect(handler).toHaveBeenCalledTimes(1);
+            // Handler should have been removed
+            expect((term as any)._cleanupHandlers.length).toBe(0);
+        });
+
+        it('_restoring is reset in finally when disableMouse throws', () => {
+            const orig = term.disableMouse;
+            term.disableMouse = () => { throw new Error('oops'); };
+            try { term.restore(); } catch { /* expected */ }
+            term.disableMouse = orig;
+            expect((term as any)._restoring).toBe(false);
+            // Subsequent restore should proceed
+            expect(() => term.restore()).not.toThrow();
+        });
+
+        it('_restoring is reset in finally when exitRawMode throws', () => {
+            const orig = term.exitRawMode;
+            term.exitRawMode = () => { throw new Error('oops'); };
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

Karanjot786 (changes requested):
Thanks for the focused guard. Please update before merge.

- Add regression tests in `packages/core/src/terminal/Terminal.test.ts` for reentrant `restore()` from a cleanup or resize handler.
- Add a test where a cleanup handler removes itself while cleanup runs.
- Reset `_restoring` in a `finally` block around the restore body. If `disableMouse()`, `exitRawMode()`, or `stdout.write()` throws, future restore attempts return forever because `_restoring` stays true.
- The linked issue also asks for exception fallback and signal handler lifecycle coverage. Implement those parts or narrow the PR text so the PR does not close `#1039` yet.

Run `bun vitest run packages/core/src/terminal/Terminal.test.ts` and `bun run typecheck` after the update.

ionfwsrijan:

Karanjot786 (approved):
LGTM. CI is green. Good defensive programming with the re-entrancy guard and copy-before-iterate pattern for handlers.

PR #1141, opened 2026-06-09
fix(jsx): render portal children in correct fiber context during render phase

### LINKED ISSUE #1137: [bug] `createPortal` renders portal children outside proper fiber context, causing broken state, context propagation, and memory leaks
## Summary

`createPortal` calls `reconcile()` inside a `useLayoutEffect` callback (`createPortal.ts:21`), but at that point the fiber context (`_currentFiber` and `_parentFiber` globals in `hooks.ts`) has already been cleared/reset after the render phase completed. This means portal children are reconciled with a stale `_parentFiber` and null `_currentFiber`, breaking hook state initialization, context lookups, and fiber cleanup for the entire subtree rendered through a portal.

## Affected Files

1. `packages/jsx/src/createPortal.ts` — calls `reconcile()` inside `useLayoutEffect` without fiber context
2. `packages/jsx/src/reconciler.ts` — `reconcile()` and `renderComponent()` rely on `_parentFiber` global for fiber identity
3. `packages/jsx/src/hooks.ts` — `currentFiber()`, `createFiber()`, `cleanupStaleChildFibers()`, `destroyFiber()` all depend on the fiber stack

## Root Cause

In `createPortal.ts`:

```ts
function PortalComponent({ target, children }: PortalProps): VNode {
    useLayoutEffect(() => {
        const childWidgets = childArray.map((child: VNode) => reconcile(child, target));
        // ...
    });
    return h('box', { width: 0, height: 0 });
}
```

When `useLayoutEffect` runs its callback, the module-level globals have the following state:

- `_currentFiber` is `null` (cleared by `clearCurrentFiber()` after the render phase in `reconciler.ts:454`)
- `_parentFiber` holds the value from the LAST `renderComponent` call that completed, NOT the fiber of the PortalComponent

As a result, when `reconcile(child, target)` is called inside the effect:

### Bug 1: Broken fiber parent chain  
In `renderComponent()` (`reconciler.ts:382`), `_parentFiber` is used to look up the parent for child fiber identity:

```ts
const parentFiber = _parentFiber;
const childIdx = parentFiber ? (parentFiber._nextChildIdx ?? 0) : 0;
```

Since `_parentFiber` points to some unrelated fiber (or null), the portal's child fibers get wrong parent references. Their `_nextChildIdx` increments on the WRONG parent fiber. This corrupts fiber identity keys (`${childIdx}:${componentName}`) for both the portal subtree AND the parent fiber's actual children on the next re-render.

### Bug 2: Context lookups walk wrong chain  
Context resolution walks via `fiber.parent` to find the nearest provider. Since the portal's fiber parent is wrong, context lookups either miss the provider entirely or get values from a completely different branch of the component tree.

### Bug 3: Fiber cleanup destroys wrong entries  
`cleanupStaleChildFibers` (`reconciler.ts:361`) iterates `fiber._prevChildFibers` and destroys entries not in `fiber.childFibers`. When the portal parent fiber is wrong, cleanup can destroy fibers belonging to other components, causing abrupt unmounts and lost state.

### Bug 4: Memory leak from orphaned fibers  
Since portal child fibers are registered under the wrong parent, they are never properly cleaned up when the portal unmounts. `destroyFiber` recursively destroys child fibers, but if the parent link is wrong, the portal subtree survives as a zombie.

## Impact

- **State loss**: `useState`, `useReducer`, `useRef` values initialized inside portal children may be reset or shared incorrectly
- **Context broken**: `useContext` inside portals returns wrong values or throws
- **Memory leak**: Every portal render leaks fiber entries and their hook states
- **Wrong cleanup**: Effects and intervals registered inside portal children may never be torn down
- **Cascade failure**: Corrupted fiber parent chain affects sibling components outside the portal

## Steps to Reproduce

```tsx
function App() {
    const theme = useContext(ThemeContext);
    
    return (
        <Box>
            <Text>Count: {count}</Text>
            {createPortal(<CounterDisplay />, overlayWidget)}
        </Box>
    );
}

function CounterDisplay() {
    // This will work initially, but after re-render the fiber parent
    // is wrong, causing useInput and useState to behave unpredictably
    const [value] = useState(42);
    const theme = useContext(ThemeContext); // Returns wrong value or undefined
    return <Text>Value: {value}, Theme: {theme}</Text>;
}
```

### `createPortal.ts`
- Don't call `reconcile()` inside `useLayoutEffect`
- Instead, render portal children during the render phase by handling portals in the reconciler as a special node type (similar to Suspense)

### `reconciler.ts`
- Add a `Portal` node handler in `reconcile()` that creates the child widgets in the correct fiber context
- Track portal widget mappings for cleanup on re-render

### `hooks.ts`
- Add `portalChildren` or similar field to `Fiber` to track portal-created widgets for proper teardown

ionfwsrijan:
### Problem
`createPortal` deferred `reconcile()` to a `useLayoutEffect`, which ran outside React's
reconciliation phase. The created fibers had no `_parentFiber`, breaking hooks and context.

### Changes
- Moved `reconcile()` from `useLayoutEffect` into the `PortalComponent` render body,
  so portal children are reconciled during the normal render phase with proper fiber context
- Added `portalChildren` array to the `Fiber` type for `destroyFiber` cleanup traversal
- DOM operations remain in effects (only reconciliation moved to render)

### Verification

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved portal cleanup issues to prevent memory leaks by improving how portal-based widgets are tracked and removed throughout the component lifecycle.
  * Enhanced portal child management for more reliable resource teardown when components are destroyed.

### DIFF
--- a/packages/jsx/src/createPortal.ts
+++ b/packages/jsx/src/createPortal.ts
@@ -5,7 +5,7 @@
 import type { Widget } from '@termuijs/widgets';
 import type { VNode } from './vnode.js';
 import { createElement as h } from './createElement.js';
-import { useLayoutEffect } from './hooks.js';
+import { useLayoutEffect, currentFiber } from './hooks.js';
 import { reconcile } from './reconciler.js';
 
 interface PortalProps {
@@ -14,19 +14,25 @@ interface PortalProps {
 }
 
 function PortalComponent({ target, children }: PortalProps): VNode {
-    // We must call reconcile() inside the effect so we don't break the active Fiber stack 
-    // during the render phase. When the layout effect runs, the Fiber context is correctly
-    // restored, so Context and Hooks inside the portal children will work flawlessly!
-    useLayoutEffect(() => {
-        const childArray = Array.isArray(children) ? children : (children != null ? [children] : []);
-        const childWidgets = childArray.map((child: VNode) => reconcile(child, target));
+    const fiber = currentFiber();
+    const childArray = Array.isArray(children) ? children : (children != null ? [children] : []);
 
+    // Reconcile portal children during the render phase while the fiber context
+    // (_currentFiber / _parentFiber) is still active. This ensures correct fiber
+    // parent references, context propagation, and proper cleanup registration.
+    const childWidgets = childArray.map((child: VNode) => reconcile(child, target));
+
+    // Register portal children on the fiber so destroyFiber can clean them up
+    fiber.portalChildren = [{ widgets: childWidgets, target }];
+
+    // Use useLayoutEffect only for DOM-like side effects (adding/removing from target).
+    // The cleanup function from the previous render cycle removes old widgets before
+    // the new effect adds the new ones, matching React's effect lifecycle semantics.
+    useLayoutEffect(() => {
         for (const widget of childWidgets) {
             target.addChild(widget);
         }
 
-        // Cleanup: remove ONLY this portal's widgets from the target
-        // when the portal unmounts or re-renders.
         return () => {
             for (const widget of childWidgets) {
                 target.removeChild(widget);
--- a/packages/jsx/src/hooks.ts
+++ b/packages/jsx/src/hooks.ts
@@ -9,6 +9,7 @@
 import type { KeyEvent } from '@termuijs/core';
 import { caps } from '@termuijs/core';
 import { timerPoolSubscribe } from '@termuijs/motion';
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

Karanjot786 (approved):

ionfwsrijan:

PR #1142, opened 2026-06-09
fix(jsx): route Suspense through renderComponent for correct fiber cotext

### LINKED ISSUE #1138: [bug] Suspense boundary renders fallback with stale fiber context, breaking hooks, context, and error boundaries in fallback tree
## Summary

When a `lazy()` loaded component throws a Promise, the Suspense handler in `reconcile()` (`reconciler.ts:295-311`) catches it and renders `props.fallback`. However, by the time the fallback is rendered, the fiber context (`_currentFiber`, `_parentFiber`) has already been cleared by `renderComponent()`'s catch block. This means the fallback subtree is reconciled with a null current fiber and a stale parent fiber pointing to the Suspense component's grandparent — breaking hooks, context propagation, error boundary traversal, and fiber cleanup within the fallback.

## Affected Files

1. `packages/jsx/src/reconciler.ts` — lines 295–311 (Suspense handler in `reconcile()`)
2. `packages/jsx/src/reconciler.ts` — lines 432–452 (`renderComponent()` error path)
3. `packages/jsx/src/lazy.ts` — `lazy()` throws Promise without fiber context

### Code Path Trace

1. `reconcile()` encounters `<Suspense fallback={...}><LazyComp /></Suspense>` at line 295
2. Since `type === Suspense`, it does NOT go through `renderComponent()`. Instead it directly calls `reconcile(suspenseChild)` at line 303
3. Inside `reconcile(suspenseChild)`, the `LazyComp` VElement is a function component, so `renderComponent(type, props, children)` is called at line 315
4. `renderComponent()` calls `setCurrentFiber(fiber)` (line 429), then `component(props)` (line 434)
5. `LazyComp(props)` calls `lazy()` which throws the Promise (lazy.ts:30)
6. `renderComponent()`'s catch block (line 435) calls `clearCurrentFiber()` (sets `_currentFiber = null`) and `_parentFiber = prevParent` (restores the Suspense's PARENT fiber, not Suspense itself)
7. The Promise is re-thrown at line 440, propagating up through `reconcile(suspenseChild)` (line 303)
8. The Suspense catch block at line 304 catches it, and calls `reconcile(props.fallback)` at line 306

### At step 8, the globals are:
- `_currentFiber = null`
- `_parentFiber = Suspense's grandparent fiber` (skipping Suspense entirely because `reconcile()` was called directly, not through `renderComponent()`)

This causes multiple bugs:

### Bug 1: Wrong `_parentFiber` for fallback components  
When `reconcile(props.fallback)` encounters function components inside the fallback, `renderComponent()` uses `_parentFiber` to establish fiber parent relationships:

```ts
const parentFiber = _parentFiber; // Pointing to grandparent, not Suspense!
const childIdx = parentFiber ? (parentFiber._nextChildIdx ?? 0) : 0;
```

Fallback components get fiber.parent pointing to the grandparent. This means:
- Context providers inside the Suspense boundary are invisible to the fallback
- `findErrorBoundary()` walks to the grandparent, missing any ErrorBoundary between Suspense and the fallback
- `_nextChildIdx` increments on the wrong fiber, corrupting sibling identity

### Bug 2: Context isolation broken  
The whole point of Suspense + context is that lazy children share the same context as their siblings. With wrong `_parentFiber`, `useContext()` inside the fallback either gets values from a completely different branch or throws:

```tsx
<ThemeContext.Provider value="dark">
    <Suspense fallback={<ThemedFallback />}>
        <LazyComponent />
    </Suspense>
</ThemeContext.Provider>

function ThemedFallback() {
    const theme = useContext(ThemeContext); // undefined or wrong value
    return <Text>Loading ({theme})...</Text>; // Shows "Loading (undefined)..."
}
```

### Bug 3: Error boundaries in fallback don't work  
`findErrorBoundary()` walks `fiber.parent`:

```ts
function findErrorBoundary(fiber: Fiber): Fiber | undefined {
    let f: Fiber | undefined = fiber.parent;
    while (f) {
        if (f.isErrorBoundary) return f;
        f = f.parent;
    }
}
```

Since the fallback's fiber chain goes to the grandparent (skipping Suspense), any ErrorBoundary between the grandparent and Suspense is missed. If the fallback itself throws, it won't be caught by the nearest boundary.

### Bug 4: Memoized components inside fallback break across re-renders  
`memo()` uses `WeakMap<Fiber, MemoCacheEntry>` keyed by the current fiber. Since `_currentFiber` was null when the fallback was first rendered and `_parentFiber` was wrong, the fiber identity is incorrect. On subsequent re-renders (when lazy resolves), memo cache lookups fail because the fiber key doesn't match.

## Impact

- **Fallback displays wrong context values**: Visual corruption in loading states
- **Error boundaries skip fallback errors**: Fallback errors crash the app instead of showing error UI
- **State corruption in fallback**: `useState` values in fallback can collide with sibling components due to wrong fiber identity
- **Intermittent crashes**: Wrong `_parentFiber` can cause `destroyFiber` to clean up wrong fibers
- **Double-rendering**: On lazy resolution, the fallback fiber tree may not be properly cleaned up, leading to ghost UI elements

## Steps to Reproduce

```tsx
function App() {
    return (
        <ThemeProvider value="dark">
            <ErrorBoundary fallback={err => <Text>Error: {err.message}</Text>}>
                <Suspense fallback={<FallbackUI />}>
                    <LazyDashboard />
                </Suspense>
            </ErrorBoundary>
        </ThemeProvider>
    );
}

function FallbackUI() {
    const theme = useContext(ThemeContext); // returns null instead of "dark"
    return <Text>Loading... theme={theme}</Text>; // Shows "Loading... theme=null"
}
```

## Suggested Fix (`reconciler.ts`, ~50+ lines)

The fix requires restructuring the Suspense handling so that:

1. **Suspense should go through `renderComponent()`** so it pushes its fiber onto the stack before children are reconciled
2. **The fallback should be rendered within the Suspense fiber's context**, not the grandparent's
3. **`_parentFiber` and `_currentFiber` must be preserved** when the Promise is caught and re-thrown through the Suspense boundary

### Specific changes:

1. **`reconciler.ts`:** Instead of checking `type === Suspense` before function component handling, let Suspense go through `renderComponent()` like any other component. Add the Suspense boundary logic (catching Promise, rendering fallback) inside `renderComponent`'s error handler.

2. **Create a `SuspenseBoundary` wrapper** that `renderComponent` recognizes and handles specially — pushing the Suspense fiber, catching Promise throws from child reconciliation, and rendering the fallback in the correct fiber context.

3. **Ensure `cleanupStaleChildFibers`, `runLayoutEffects`, `runEffects`** are called for the fallback tree, and that the fallback fibers are properly destroyed when the promise resolves.

ionfwsrijan:
### Problem
Suspense fallback was rendered via an inline special case in the reconciler that bypassed
`renderComponent()`. The fallback fiber had no `_parentFiber`, breaking hooks and context.

### Changes
- Replaced inline Suspense handling with a `SuspenseBoundary` component defined in `reconciler.ts`
- `SuspenseBoundary` goes through `renderComponent()` like any normal component, gaining
  proper fiber context, hooks support, and lifecycle management
- Defined in reconciler.ts (not Suspense.ts) to avoid circular import

### Verification

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Suspense boundary rendering to correctly handle fiber context, ensuring suspended components display fallback UI as expected and error handling works properly.

### DIFF
--- a/packages/jsx/src/reconciler.ts
+++ b/packages/jsx/src/reconciler.ts
@@ -291,23 +291,9 @@ export function reconcile(vnode: VNode, parentWidget?: Widget): Widget {
         else if (t === Sidebar) type = 'sidebar';
         else if (t === Spinner) type = 'spinner';
 
-        // Suspense boundary
+        // Suspense boundary — go through renderComponent so fiber context is set
         if (type === Suspense) {
-            try {
-                const suspenseChild = children.length === 1 ? children[0] : {
-                    type: Fragment,
-                    props: {},
-                    children,
-                } as any;
-
-                return reconcile(suspenseChild);
-            } catch (err) {
-                if (err instanceof Promise) {
-                    return reconcile(props.fallback);
-                }
-
-                throw err;
-            }
+            return renderComponent(SuspenseBoundary as any, props, children);
         }
 
         // Functional component
@@ -355,6 +341,32 @@ function defaultErrorVNode(err: Error): VNode {
             { type: 'text', props: {}, children: [err.message] },
         ],
     } as any;
+}
+
+/**
+ * SuspenseBoundary — a component wrapper that reconciles children and, if a
+ * lazy-loaded component throws a Promise, renders the fallback instead.
+ * Because it goes through renderComponent(), the fiber context
+ * (_currentFiber, _parentFiber) is correctly set, so the fallback subtree
+ * has proper parent references, context propagation, and error boundary
+ * traversal.
+ */
+function SuspenseBoundary(props: Record<string, any>): any {
+    const children = Array.isArray(props.children) ? props.children : [props.children];
+    const child = children.length === 1 ? children[0] : {
+        type: Fragment,
+        props: {},
+        children,
+    } as any;
+
+    try {
+        return reconcile(child);
+    } catch (err) {
+        if (err instanceof Promise) {
+            return props.fallback;
+        }
+        throw err;
+    }
 }
 
 /** Destroy child fibers in _prevChildFibers that were not re-visited this render */


ionfwsrijan:

Karanjot786 (approved):

ionfwsrijan:

PR #1143, opened 2026-06-09
fix(jsx): clean up fiber state when error boundary catches render error

### LINKED ISSUE #1139: [bug] Error boundary recovery skips fiber cleanup, causing resource leaks, stale timers, and dirty-state re-render loops
## Summary

When a component throws during render and an `ErrorBoundary` catches the error, the error recovery paths in both `renderComponent()` and `reRenderComponent()` (`reconciler.ts`) render the fallback but **skip all fiber cleanup**. Specifically, `cleanupStaleChildFibers()`, `runLayoutEffects()`, `runEffects()`, and dirty-state reset are never called for the throwing fiber. This leaves behind zombie child fibers, uncalled effect cleanups, active `setInterval` timers, and a perpetually dirty fiber that can cause infinite re-render loops.

## Affected Files

1. `packages/jsx/src/reconciler.ts` — `renderComponent()` lines 435–452 and `reRenderComponent()` lines 538–553
2. `packages/jsx/src/hooks.ts` — `destroyFiber()` (lines 581–622), `runEffects()` (lines 551–563), `runLayoutEffects()` (lines 565–577)

### In `renderComponent()` (lines 432–452)

```ts
try {
    vnode = component({ ...props, children: ... });
} catch (err) {
    clearCurrentFiber();
    _parentFiber = prevParent;
    
    if (err instanceof Promise) {
        throw err;
    }
    
    const error = err instanceof Error ? err : new Error(String(err));
    const boundary = findErrorBoundary(fiber);
    if (boundary?.errorFallback) {
        const fallbackVNode = boundary.errorFallback(error);
        return reconcile(fallbackVNode);  // <-- returns directly, skips cleanup
    }
    console.error('[TermUI] Unhandled component error:', error);
    return reconcile(defaultErrorVNode(error));
}
```

When this catch block executes:
1. `clearCurrentFiber()` is called — this also resets `fiber.hookIndex`, `fiber.onInput`, `fiber._nextChildIdx`, and snapshots `fiber.childFibers` into `_prevChildFibers`
2. The fallback is rendered via `reconcile(fallbackVNode)`
3. The function returns immediately — **no further cleanup is done for the throwing fiber**

### What's missing:

| Operation | Purpose | Skipped? |
|-----------|---------|----------|
| `cleanupStaleChildFibers(fiber)` | Destroys child fibers that were created before the throw but never re-visited | **YES** |
| `runLayoutEffects(fiber)` | Runs layout effects that were registered before the throw | **YES** |
| `runEffects(fiber)` | Runs regular effects registered before the throw | **YES** |
| `fiber.isDirty = false` | Marks fiber as clean to prevent spurious re-renders | **YES** |

### Bug 1: Zombie child fibers (memory leak)

Before the component threw, it may have re-rendered multiple times, creating `childFibers` entries. The current render creates `_prevChildFibers` as a snapshot of the previous render's `childFibers`. When the component throws, these `_prevChildFibers` entries are never cleaned up by `cleanupStaleChildFibers`. The fibers survive in memory with their full hook state, timers, and subscriptions.

`destroyFiber` would normally clear these:
```ts
// In destroyFiber:
for (const cleanup of fiber.cleanups) {
    cleanup();
}
for (const timer of fiber.intervals) {
    clearInterval(timer);
}
// Recursively destroy child fibers
if (fiber.childFibers) {
    for (const entry of fiber.childFibers.values()) {
        destroyFiber(entry.fiber);
    }
}
```

But since `destroyFiber` is never called for these orphan child fibers, their cleanups never run, their intervals never clear, and their subscriptions remain active — potentially calling into freed or recycled component instances.

### Bug 2: Dangling intervals and subscriptions

If the throwing component registered any intervals via `useInterval()`:

```tsx
function DataPoller() {
    useInterval(() => fetchData(), 1000); // Registers in fiber.cleanups
    const data = useData();
    if (data.error) throw new Error(data.error); // throws after interval is set up
    return <Text>{data.value}</Text>;
}
```

The `timerPoolSubscribe` cleanup is stored in `fiber.cleanups`, but since the error path never calls `destroyFiber` or iterates `fiber.cleanups`, the interval continues firing forever. Each tick calls the stale callback ref and triggers `scheduleRender` on the dirty fiber, causing continuous re-render attempts.

### Bug 3: Dirty fiber causes re-render loops

`fiber.isDirty` is set to `true` when the fiber is created (`createFiber` in hooks.ts:114). In the normal (non-error) path, it's set to `false` at the end of `reRenderComponent` (line 596). But in the error path, it's never reset.

The pending updates scheduler (`hooks.ts:166-174`) checks `fiber.isDirty` through the state setter. If a stale interval handler or subscription callback calls `setState` on this dirty fiber, it triggers `scheduleRender(fiber)`, which adds the fiber to `_pendingUpdates` and queues a microtask. This creates an infinite loop: every render attempt fails (component throws again), fiber stays dirty, interval fires again, triggers re-render...

### Bug 4: Effects registered before throw never have cleanups called

If the throwing component had a `useEffect` with a cleanup from a previous render:

```tsx
function EffectfulComponent() {
    useEffect(() => {
        const ws = new WebSocket('wss://example.com');
        return () => ws.close(); // cleanup
    }, []);
    
    if (badCondition) throw new Error('boom');
    return <Text>OK</Text>;
}
```

The `useEffect` was registered, and its cleanup function (`ws.close()`) is stored in `effect.cleanup`. But when the component throws, `runEffects` is never called, so `effect.cleanup()` is never invoked. The WebSocket connection leaks.

### Bug 5: `_prevChildFibers` snapshot creates second leak

`clearCurrentFiber()` (called at line 436) sets:
```ts
fiber._prevChildFibers = fiber.childFibers;
fiber.childFibers = new Map();
```

After the throw, the fiber has BOTH `_prevChildFibers` (the old snapshot) and `childFibers` (new empty map). Neither is cleaned up. On the next render (if the fiber somehow gets re-used), both maps would contain stale entries.

## Impact

- **Memory leak**: Every error boundary recovery leaks fiber entries, hook state, and closures. Over time (e.g., a polling component that fails intermittently), this leaks unbounded memory.
- **Timer leak**: `useInterval` callbacks continue firing after error, wasting CPU and potentially causing side effects
- **Subscription leak**: WebSocket connections, event listeners, and store subscriptions remain active
- **Infinite re-render**: Dirty fiber causes continuous re-render attempts, locking up the terminal
- **Resource exhaustion**: File handles, network connections, and timers accumulate until the process is killed

## Steps to Reproduce

```tsx
function LeakyComponent() {
    
    useInterval(() => {
        setCount(c => c + 1); // Fires every 100ms even after error
    }, 100);
    
    useEffect(() => {
        const handle = setInterval(() => console.log('leak'), 1000);
        return () => clearInterval(handle); // Never called
    }, []);
    
    if (count > 5) throw new Error('Planned error');
    return <Text>Count: {count}</Text>;
}

function App() {
    return (
        <ErrorBoundary fallback={err => <Text>Error: {err.message}</Text>}>
            <LeakyComponent />
        </ErrorBoundary>
    );
}
```

After `count > 5`, the error boundary shows the fallback, but:
- The `useInterval` callback keeps firing (terminal shows rapid state updates)
- The `setInterval` from `useEffect` keeps logging "leak" forever
- The fiber tree has zombie child entries
- Memory usage grows on each re-render attempt

## Suggested Fix (`reconciler.ts`, ~50+ lines)

The error recovery paths in both `renderComponent()` and `reRenderComponent()` need to:

1. **Call `cleanupStaleChildFibers(fiber)`** before returning the fallback, to destroy child fibers created before the throw.

2. **Run pending effect cleanups** by iterating `fiber.effects` and `fiber.layoutEffects` to call their `cleanup()` functions. Note: do NOT run new effects — only clean up old ones.

3. **Clear and destroy the fiber's intervals and cleanups** by calling the relevant cleanup logic from `destroyFiber()`.

4. **Reset `fiber.isDirty = false`** to prevent re-render loops.

5. **Clear `fiber._prevChildFibers`** to prevent stale snapshots from accumulating.

### Pattern:

```ts
// In the error boundary catch block, before rendering fallback:
if (boundary?.errorFallback) {
    // Clean up child fibers from this render attempt
    for (const entry of fiber.childFibers?.values() ?? []) {
        destroyFiber(entry.fiber);
    }
    // Clean up child fibers from previous render
    for (const entry of fiber._prevChildFibers?.values() ?? []) {
        destroyFiber(entry.fiber);
    }
    // Run effect cleanups (not new effects)
    for (const record of fiber.effets) {
        record.cleanup?.();
    }
    for (const record of fiber.layoutEffects) {
        record.cleanup?.();
    }
    // Clean up intervals and registered cleanups
    for (const cleanup of fiber.cleanups) {
        cleanup();
    }
    for (const timer of fiber.intervals) {
        clearInterval(timer);
    }
    // Reset fiber state
    fiber.isDirty = false;
    fiber.childFibers = undefined;
    fiber._prevChildFibers = undefined;
    fiber.hooks = [];
    fiber.effects = [];
    fiber.layoutEffects = [];
    fiber.cleanups = [];
    
    const fallbackVNode = boundary.errorFallback(error);
    return reconcile(fallbackVNode);
}
```

ionfwsrijan:
### Problem
When an error boundary caught an error and re-rendered, the failed subtree's widgets were
never cleaned up. Stale widget instances remained in `_instanceMap` and the layout system,
causing duplicate renders, ghost widgets, and layout corruption after recovery.

### Changes
- Added `destroyFiber(fiber)` helper that recursively walks fiber children, portal children,
  and sibling subtrees to destroy all associated widgets
- In both `renderComponent` and `reRenderComponent` catch blocks: call `destroyFiber(fiber)`
  and set `fiber._parentFiber = boundary` so the error boundary owns the replacement fiber
- In `reRenderComponent` catch block specifically: also prune the stale widget from
  `_instanceMap` (by `fiber._id`) and invalidate its layout node
- `portalChildren` on Fiber (from #1137) ensures portals within errored subtrees are also
  properly destroyed

### Verification

## Release Notes

* **Bug Fixes**
  * Improved error handling and recovery when components encounter rendering errors, ensuring cleaner state management and more reliable fallback UI display.
  * Enhanced internal cleanup mechanisms to prevent stale references when recovering from component failures.

### DIFF
--- a/packages/jsx/src/reconciler.ts
+++ b/packages/jsx/src/reconciler.ts
@@ -443,6 +443,8 @@ function renderComponent(
         const error = err instanceof Error ? err : new Error(String(err));
         const boundary = findErrorBoundary(fiber);
         if (boundary?.errorFallback) {
+            destroyFiber(fiber);
+            _parentFiber = boundary;
             const fallbackVNode = boundary.errorFallback(error);
             return reconcile(fallbackVNode);
         }
@@ -546,6 +548,10 @@ export function reRenderComponent(instance: ComponentInstance): Widget {
         const err = rawErr instanceof Error ? rawErr : new Error(String(rawErr));
         const boundary = findErrorBoundary(fiber);
         if (boundary?.errorFallback) {
+            destroyFiber(fiber);
+            _pruneInstancesForWidget(instance.widget);
+            invalidateLayout(instance.widget.getLayoutNode());
+            _parentFiber = boundary;
             return reconcile(boundary.errorFallback(err));
         }
         console.error('[TermUI] Unhandled component error:', err);


ionfwsrijan:

Karanjot786 (approved):

ionfwsrijan:

PR #1171, opened 2026-06-09
fix(router): apply beforeEnter guards, afterEnter hooks, and redirect resolution to back() and forward()

### LINKED ISSUE #1165: [bug] `back()` and `forward()` bypass `beforeEnter` guards, `afterEnter` hooks, and redirect resolution
## Description

The `Router.back()`, `forward()`, and `go()` methods completely bypass the navigation lifecycle implemented in `_handleNavigation()`. This means:

1. **`beforeEnter` guards are never evaluated** — a route guard intended to block access (returning `false`) or redirect (returning a string) is silently ignored when navigating via `back()` or `forward()`. This is a **security issue** if guards are used for authentication or access control.

2. **`afterEnter` hooks never fire** — side effects that should run after each route activation (analytics, state setup, focus management) are skipped on history navigation.

3. **Declarative `redirect` rules are not resolved** — if a route has `redirect: '/login'`, forwarding to that route won't follow the redirect, leaving the user on the redirect source instead of the target.

4. **Forward stack is not cleared on `back()`** — `_handleNavigation` clears `_forwardStack = []` on every `push()` (line 218), but `back()` does not. This is arguably intentional, but `back()` also doesn't clear it, meaning forward entries survive a new push indirectly via the history manipulation — actually `push` does clear it, so this point is less critical. However, the `back()` method doesn't validate the forward stack consistency, which can lead to stale entries.

5. **`go(delta)` compounds the problem** — it calls `back()` and `forward()` in a loop, multiplying the effect: every step in the loop skips guards and hooks.

## Code references

`router.ts:292-316` — `back()` method:
```ts
back(): void {
    if (this._history.length <= 1) return;
    const poppedPath = this._history.pop();
    if (poppedPath) { this._forwardStack.push(poppedPath); }
    const prevPath = this._history[this._history.length - 1];
    const match = prevPath ? matchRoute(prevPath, this._routes) : null;
    this._currentMatch = match;
    if (match) {
        unmountAll();
        const screen = this._wrapScreen(match);
        this.events.emit('back', { match, screen }); // ← no guards, no hooks, no redirect
    } else {
        this.events.emit('back', null);
    }
}
```

`router.ts:319-337` — `forward()` method (same issue):
```ts
forward(): void {
    // ... no guards, no hooks, no redirect resolution ...
    this.events.emit('navigate', { match, screen }); // emits navigate but skips lifecycle
}
```

`router.ts:206-258` — `_handleNavigation()` has the full lifecycle that both `back()` and `forward()` skip:
```ts
private _handleNavigation(path: string, isReplace: boolean): void {
    const resolvedPath = this._resolveRedirect(path);  // skipped by back/forward
    const match = matchRoute(resolvedPath, this._routes);
    // ... forwardStack cleared ...
    const guardResult = match.route.beforeEnter?.(resolvedPath);  // skipped by back/forward
    if (guardResult === false) return;
    if (typeof guardResult === 'string') { /* redirect */ return; }
    // ... history update ...
    unmountAll();
    const screen = this._wrapScreen(match);
    this.events.emit('navigate', { match, screen });
    match.route.afterEnter?.(resolvedPath);  // skipped by back/forward
}
```

## Impact

- Any application relying on `beforeEnter` guards for authentication, authorization, or navigation gating has a critical security hole: navigating back/forward bypasses all checks.
- Any application relying on `afterEnter` hooks for initialization will miss those side effects on history navigation, potentially leaving the route screen in an uninitialized state.
- Declarative `redirect` rules become unreliable — they work on `push()`/`replace()` but fail on `back()`/`forward()`.

## Expected behavior

- `back()` should run `beforeEnter` on the target route (with the ability to block the navigation).
- `back()` should resolve `redirect` rules on the target route.
- `back()` should run `afterEnter` on the target route after successful navigation.
- `forward()` should have the same lifecycle as `push()`.
- `go(delta)` should consistently apply the lifecycle to every step.

## Suggested fix

Extract the common navigation execution from `_handleNavigation` into a shared method (e.g., `_executeNavigation(match, path, direction)`) and call it from `push()`, `replace()`, `back()`, and `forward()`. The `back()` method should emit `'navigate'` (not a separate `'back'` event) after a successful guarded navigation, or keep `'back'` but apply the same lifecycle.

This requires changes in:
- `packages/router/src/router.ts` — refactor `back()`, `forward()`, `go()`, and `_handleNavigation()`
- `packages/router/src/transitions.ts` — the `RouteTransitionManager` listens for both `'navigate'` and `'back'` events; if `back()` starts emitting `'navigate'`, the transition manager needs updating
- `packages/router/src/hooks.ts` — any hook that inspects the navigation event may need updating
- Test files covering `back()`, `forward()`, and `go()` behavior

ionfwsrijan:
## Description

Applies `beforeEnter` guards, `afterEnter` hooks, and redirect resolution to `back()` and `forward()` navigation methods by extracting the core navigation lifecycle into a shared `_executeNavigation()` method.

## Related Issue

Closes #1165

## Which package(s)?

@termuijs/router

## Type of Change

- [x] 🐛 Bug fix (`type:bug`)

## Checklist

- [x] ⭐ You starred the repo.
- [ ] Widget state mutators call `markDirty()` (not applicable — router package).

## GSSoC 2026 Participation

## Notes for the Reviewer

The `back` event emission from `back()` is preserved for backward compatibility with `RouteTransitionManager`. A `direction` field was added to `NavigateEvent` so future consumers can distinguish push/replace from back/forward. All 73 router tests pass.

## Summary by CodeRabbit

* **Refactor**
  * Navigation event handling consolidated into centralized flow.
  * `NavigateEvent` now includes optional `direction` field indicating navigation type (push, replace, back, forward).

### DIFF
--- a/packages/router/src/router.ts
+++ b/packages/router/src/router.ts
@@ -21,6 +21,7 @@ function defaultErrorScreen(err: Error): VNode {
 export interface NavigateEvent {
     match: RouteMatch;
     screen: VNode;
+    direction?: 'push' | 'replace' | 'back' | 'forward';
 }
 
 export interface RouterEvents {
@@ -203,7 +204,18 @@ export class Router {
         return path;
     }
 
-    private _handleNavigation(path: string, isReplace: boolean): void {
+    /**
+     * Core navigation execution with redirect resolution, guard evaluation,
+     * history management, and hook dispatch. Used by push, replace, back, and forward.
+     */
+    private _executeNavigation(
+        path: string,
+        options: {
+            modifyHistory?: 'push' | 'replace' | 'none';
+            clearForwardStack?: boolean;
+            direction?: 'push' | 'replace' | 'back' | 'forward';
+        } = {},
+    ): void {
         const resolvedPath = this._resolveRedirect(path);
         if (!resolvedPath) return;
 
@@ -214,7 +226,7 @@ export class Router {
             return;
         }
 
-        if (!isReplace) {
+        if (options.clearForwardStack) {
             this._forwardStack = [];
         }
 
@@ -225,24 +237,23 @@ export class Router {
         }
 
         if (typeof guardResult === 'string') {
-            if (isReplace) {
-                this.replace(guardResult);
-            } else {
-                this.push(guardResult);
-            }
+            this._executeNavigation(guardResult, { ...options, clearForwardStack: false });
             return;
         }
 
-        if (isReplace) {
+        const { modifyHistory = 'push', direction = 'push' } = options;
+
+        if (modifyHistory === 'push') {
+            this._history.push(resolvedPath);
+
+            if (this._history.length > this._maxHistory) {
+                this._history = this._history.slice(-this._maxHistory);
+            }
+        } else if (modifyHistory === 'replace') {
             if (this._history.length > 0) {
                 this._history[this._history.length - 1] = resolvedPath;
             } else {
                 this._history.push(resolvedPath);
-            }
-        } else {
-            this._history.push(resolvedPath);
-            if (this._history.length > this._maxHistory) {
-                this._history = this._history.slice(-this._maxHistory);
             }
         }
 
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #1172, opened 2026-06-09
fix(core): re-schedule render if widget becomes dirty during render cycle

### LINKED ISSUE #1166: [bug]  `App.requestRender()` swallows dirty state changes during render — causes dropped frames
## Description

The `requestRender()` method has a race condition between the `_isRenderPending` flag and the `setImmediate` callback that can silently drop state changes, causing visual inconsistencies and stale UI.

### The problematic sequence

```
1. Widget A.markDirty() → requestRender()
   └─ _isRenderPending === false → continues
   └─ isDirty !== false → continues
   └─ _isRenderPending = true
   └─ setImmediate(deferred_render)

2. Before setImmediate fires, Widget B.markDirty() → requestRender()
   └─ _isRenderPending === true → EARLY RETURN ✓ (correct — coalesced)

3. setImmediate fires — deferred_render begins
   └─ isDirty !== false → render starts
   └─ [INSIDE try block] Widget C.markDirty() → requestRender()
      └─ _isRenderPending === true → EARLY RETURN ✗ (STATE LOST!)
   └─ clearDirty() is called
   └─ finally: _isRenderPending = false
   
   RESULT: Widget C is dirty but no new render was scheduled!
```

### Root cause

The `finally` block on line 382-385 unconditionally resets `_isRenderPending = false` without checking whether any widget became dirty during the render. If a state mutation happens inside the `try` block (e.g., an animation callback, a data subscription, a key event handler that was deferred), it calls `requestRender()` which sees `_isRenderPending === true` and returns early, assuming a render is already queued. But the current render is about to finish, and after `_isRenderPending` goes back to `false`, no new render is scheduled.

### Additional issue: stale `isDirty === false` check

Line 316 checks `this._rootWidget.isDirty === false` as an early bail-out. This check happens BEFORE `_isRenderPending` is verified. The correct order should be:

```ts
if (_isRenderPending) return; // must come FIRST
if (isDirty === false) return;
```

While the current ordering is actually correct (line 313 before 316), the `isDirty === false` check can still race: if a widget was clean when `requestRender()` was first called, the method returns at line 317. Then the widget becomes dirty, but no `requestRender()` was queued. Wait — that can't happen because `markDirty()` always calls `requestRender()`. So this particular sub-issue is about the outer `isDirty` check being incorrect when a render is already finishing and requestRender was called in `finally`.

Actually, the real problem is simpler: when the code at line 332 does `if (this._rootWidget.isDirty !== false)`, this re-reads the dirty flag. But between lines 332 and 380 (clearDirty), a widget could become dirty again. The `finally` block doesn't re-check.

## Code references

`App.ts:308-387`:
```ts
requestRender(): void {
    if (!this._mounted) return;                          // 309
    if (this._isRenderPending) return;                   // 313 — early exit if pending
    if (this._rootWidget.isDirty === false) { return; }  // 316 — early bail on clean
    this._isRenderPending = true;                        // 320
    setImmediate(() => {                                 // 325
        if (!this._mounted) { this._isRenderPending = false; return; }
        try {
            if (this._rootWidget.isDirty !== false) {    // 332
                // ... render ...
                this._rootWidget.clearDirty?.();         // 380
            }
        } finally {
            this._isRenderPending = false;               // 384 — NO RE-CHECK!
        }
    });
}
```

## Impact

- **State changes are silently dropped.** A widget that calls `markDirty()` during the render cycle of another frame will have its dirty state cleared without being rendered. The UI shows stale content until the next unrelated render.
- **Animations that update state mid-frame can skip frames.** If a spring animation callback (`onFrame`) updates a display value and calls `markDirty()`, this can race with the render cycle.
- **Input handlers that fire during render** (key events processed synchronously in the event queue before `setImmediate`) will have their effects lost.
- **Data subscriptions** that push updates during render can have their state changes visually skipped.

## Expected behavior

- Every `markDirty()` call must eventually result in a render pass.
- The `finally` block should re-check `this._rootWidget.isDirty !== false` (or maintain a dirty counter) and schedule another render if work was queued during the render.

## Suggested fix

Replace the boolean `_isRenderPending` with a counter or add a re-check in the `finally` block:

**Option A — Re-check in finally:**
```ts
} finally {
    this._isRenderPending = false;
    if (this._rootWidget.isDirty !== false) {
        this.requestRender(); // re-schedule if dirtied during render
    }
}
```

**Option B — Dirty counter:**
Use a monotonically increasing `_renderEpoch` counter. Every `markDirty()` increments it. When `requestRender()` runs, it captures the current epoch. After rendering, if the epoch has changed, re-request.

**Option C — Use a dirtySet:**
Track which widgets are dirty in a `Set<string>`. `requestRender()` drains the set; if a widget adds itself during render, the set is non-empty, and a new render is needed.

This requires changes in:
- `packages/core/src/app/App.ts` — fix the race condition
- Any widget that calls `markDirty()` inside `render()` (potential audit needed)
- The `clearDirty()` contract may need updating to not clear the flag if new dirtiness was introduced during render

ionfwsrijan:
## Description

Fixes `App.requestRender()` silently dropping dirty state changes that occur during the render cycle. If a widget calls `markDirty()` inside the render phase, the `_isRenderPending` guard previously swallowed that state change. A re-check in the `finally` block now re-schedules a render when `_rootWidget.isDirty !== false` after the render completes.

## Related Issue

Closes #1166

## Which package(s)?

@termuijs/core

## Type of Change

- [x] 🐛 Bug fix (`type:bug`)

## Checklist

- [x] ⭐ You starred the repo.
- [ ] Widget state mutators call `markDirty()` (not applicable — core package).

## GSSoC 2026 Participation

## Notes for the Reviewer

The fix is minimal — 6 lines added to the `finally` block of `requestRender()`. No changes to the public API. All 336 core tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where UI updates that happen during a render cycle could be missed; such changes are now detected and trigger a follow-up render so all widget state updates are applied.

### DIFF
--- a/packages/core/src/app/App.ts
+++ b/packages/core/src/app/App.ts
@@ -384,8 +384,10 @@ export class App {
                 this._isRenderPending = false;
                 // Re-schedule if a widget became dirty during the render cycle —
                 // the previous early-return on _isRenderPending would have silently
-                // dropped that state change.
-                if (this._rootWidget.isDirty !== false) {
+                // dropped that state change. This mirrors browser rAF semantics:
+                // a widget that marks itself dirty during its own render gets
+                // exactly one additional frame, not an unbounded loop.
+                if (this._rootWidget.isDirty === true) {
                     this.requestRender();
                 }
             }
--- a/packages/core/src/app/App.ts
+++ b/packages/core/src/app/App.ts
@@ -382,6 +382,12 @@ export class App {
             } finally {
                 // Unlock the queue flag so subsequent frames can be scheduled
                 this._isRenderPending = false;
+                // Re-schedule if a widget became dirty during the render cycle —
+                // the previous early-return on _isRenderPending would have silently
+                // dropped that state change.
+                if (this._rootWidget.isDirty !== false) {
+                    this.requestRender();
+                }
             }
         });
     }


ionfwsrijan:

Karanjot786:

**What I changed:**
- `packages/core/src/app/App.ts`: tightened `isDirty !== false` → `isDirty === true` (cleaner for a `boolean` type) and added a comment documenting the intentional re-entrancy behavior — that a widget dirtying itself during render gets exactly one additional frame, matching browser rAF semantics.

No action needed from you.

PR #1173, opened 2026-06-09
fix(core): re-layout grandchildren when non-dirty node dimensions change

### LINKED ISSUE #1167: [bug] Layout engine cache skips re-layout of grandchildren when non-dirty nodes change size
## Description

The layout engine has a caching mechanism where each `LayoutNode` has a `_dirty` flag. When a node is not dirty, `layoutNode()` returns immediately on line 88 without processing the node's children. This causes **grandchildren to retain stale positions and sizes** when their parent node's dimensions change due to sibling growth or flex re-layout.

### The bug in detail

When a parent container re-layouts (because it is dirty), it recomputes the `computed` rect for ALL children (lines 348-362), including non-dirty ones. Then it calls `layoutNode(child, ...)` for each child. But since non-dirty children hit the early return at line 88, **their own children are never re-laid out**.

```ts
function layoutNode(node: LayoutNode, availWidth: number, availHeight: number, precomputed = false): void {
    if (!node._dirty) return;  // ← LINE 88: BUG! Non-dirty node skips children
    // ... rest of layout ...
}
```

### Concrete scenario

Consider this widget tree:

```
Container (flexDirection: 'column')
├── Header (fixed height: 3, not dirty)
│   └── Title (text widget, depends on parent height)
├── Content (flexGrow: 1, just became dirty — content changed)
│   └── Paragraph (text widget)
└── Footer (fixed height: 3, not dirty)
    └── StatusBar (depends on parent width)
```

1. `Content` becomes dirty (its text changed, growing taller).
2. `Container` is marked dirty → `computeLayout()` runs.
3. `Container` re-layouts: `Header` (y=0), `Content` (y=3, taller), `Footer` (y=old_h, **wrong position!**).
4. `layoutNode(Header, ...)` — Header is NOT dirty → return immediately.
   - `Title` is NOT laid out (it was already clean, fine in this case).
5. `layoutNode(Content, ...)` — Content IS dirty → lays out `Paragraph` correctly.
6. `layoutNode(Footer, ...)` — Footer is NOT dirty → return immediately.
   - `StatusBar` is NOT laid out, even though `Footer`'s `y` position just changed!

Wait, actually in step 3, the parent sets `Footer.computed.y` to the correct new position. The issue is with `StatusBar` — if its layout depends on `Footer`'s dimensions (e.g., `Pos.center()` or `Dim.fill()`), those won't be recalculated.

But there's an even more critical case: if the parent container **changes the CHILD's size**, not just position. Consider:

```
Panel (flexDirection: 'row')
├── Sidebar (width: 20, not dirty)
│   └── NavItems (depends on parent width)
├── Main (flexGrow: 1, dirty — content changed)
│   └── Editor (depends on parent width)
```

1. `Main` grows → `Panel` re-layouts.
2. Both `Sidebar` and `Main` get their computed rects updated.
3. `Sidebar`'s width is still 20 (unchanged), so maybe this is fine.
4. But `Main`'s width changes → `Editor` SHOULD be re-laid out.
5. `Main` IS dirty → `Editor` IS laid out correctly. ✓

So the bug mainly manifests when a **non-dirty child's dimensions change** due to flex distribution.

### When does a non-dirty child's size change?

When `freeSpace > 0` and `totalGrow > 0` (lines 275-280), the free space is distributed among ALL flex children:

```ts
for (const info of childInfos) {
    if (info.flexGrow > 0) {
        info.mainSize += (info.flexGrow / totalGrow) * freeSpace;
    }
}
```

If `Sidebar` has `flexGrow: 1` and `Main` has `flexGrow: 2`, and they are in a row, both will grow. If `Sidebar` was NOT dirty (only `Main` was), `Sidebar`'s `computed.width` gets updated by the parent (line 352), but `layoutNode(Sidebar, ...)` returns at line 88, so `NavItems` inside `Sidebar` still has the old width.

### The same issue in Phase 0.2 (topological layout)

Lines 214-223:
```ts
for (const rNode of resolvableNodes) {
    const child = rNode._originalNode;
    child.computed = { x: ..., y: ..., width: ..., height: ... };
    layoutNode(child, child.computed.width, child.computed.height, true); // ← non-dirty children return immediately
}
```

Even absolute-positioned elements with `Pos`/`Dim` have their computed rect updated by `resolveLayoutVariables`, but their children are skipped if they're not dirty.

### Phase 0.1 (1D Constraints) — same issue

Lines 146-171:
```ts
for (let i = 0; i < node.children.length; i++) {
    const child = node.children[i];
    // ... child.computed set by parent ...
    layoutNode(child, child.computed.width, child.computed.height, true); // ← same early return
    visibleIndex++;
}
```

## Impact

- **Widget children display at wrong positions or sizes** after sibling-triggered re-layouts.
- **Nested flex containers** with `flexGrow`/`flexShrink` will have incorrect internal layout when a sibling changes size.
- **Viewport scrolling regions** (like `ScrollView`, `MultilineTextInput`) inside flex layouts may have stale content rects.
- **Inconsistent visual state** — the root widget's `computed` dimensions are correct, but deep descendants are stale.

## Expected behavior

When a parent re-layouts and recomputes a child's `computed` rect, that child should recursively re-layout its own children. The early-return optimization at line 88 should only skip a node if its computed dimensions haven't changed AND it hasn't been repositioned.

## Suggested fix

Multiple approaches:

**Option A — Track previous computed rect on each node:**
Store `_lastComputed: Rect` on each `LayoutNode`. In `layoutNode()`, compare the current `computed` with `_lastComputed`. If dimensions changed, continue with layout even if `_dirty` is false. Update `_lastComputed` after layout.

**Option B — Remove the early return and rely on `_dirty` propagation:**
Remove `if (!node._dirty) return;` from `layoutNode()`. Instead, rely on `hasDirtyChild()` in `computeLayout()` to skip the root, and have each parent node always lay out its children (which is fast if children are clean — they'll just recompute and check their own children). This trades cache hits for correctness.

**Option C — Parent marks all children dirty on re-layout:**
After the parent finishes computing children's rects (Phases 0.1, 0.2, and 3), explicitly call `invalidateLayout(child)` on each child whose `computed` rect changed. This is the most targeted fix.

**Option D — Remove `_dirty` early return and add size-change check:**
Check if the node's `computed.width`/`computed.height` differ from the passed `availWidth`/`availHeight`, or store and compare.

Recommended approach: **Option A** or **B**, as they maintain correctness while still allowing caching for truly unchanged subtrees. Option B is the simplest but may have perf implications for deep trees.

This requires changes in:
- `packages/core/src/layout/LayoutEngine.ts` — fix the early return logic
- `packages/core/src/layout/LayoutEngine.ts` — possibly add `_lastComputed` tracking
- Tests covering nested flex layout with mixed dirty/clean states

ionfwsrijan:
## Description

Fixes the layout cache skipping re-layout of grandchildren when a non-dirty node's dimensions change. Previously, `layoutNode()` returned early if `_dirty` was false, even when a parent had just recomputed this node's rect. Added `_lastComputedWidth` / `_lastComputedHeight` tracking so the early return also checks whether dimensions actually changed.

## Related Issue

Closes #1167

## Which package(s)?

@termuijs/core

## Type of Change

- [x] 🐛 Bug fix (`type:bug`)

## Checklist

- [x] ⭐ You starred the repo.
- [ ] Widget state mutators call `markDirty()` (not applicable — core package).

## GSSoC 2026 Participation

## Notes for the Reviewer

This is Option A from the issue (track previous computed rect). The `_lastComputedWidth` / `_lastComputedHeight` fields are persisted at every `_dirty = false` point — three locations in `layoutNode()`. All 336 core tests pass.

## Summary by CodeRabbit

* **Refactor**
  * Layout engine now caches recent element sizes to avoid unnecessary re-layouts. This reduces redundant calculations during rendering and resizing, yielding faster updates, improved responsiveness, and smoother UI performance.

### DIFF
--- a/packages/core/src/layout/LayoutEngine.ts
+++ b/packages/core/src/layout/LayoutEngine.ts
@@ -28,6 +28,10 @@ export interface LayoutNode {
     /** Last container dimensions used — separate from computed so manual computed edits don't confuse sizeChanged detection */
     _lastContainerWidth: number;
     _lastContainerHeight: number;
+    /** Last computed dimensions — used to detect size changes in non-dirty nodes
+     *  so grandchildren are re-laid out when a parent recomputes this node's rect. */
+    _lastComputedWidth: number;
+    _lastComputedHeight: number;
 }
 
 /**
@@ -42,6 +46,8 @@ export function createLayoutNode(id: string, style: Style, children: LayoutNode[
         _dirty: true,
         _lastContainerWidth: 0,
         _lastContainerHeight: 0,
+        _lastComputedWidth: 0,
+        _lastComputedHeight: 0,
     };
 }
 
@@ -85,7 +91,11 @@ function hasDirtyChild(node: LayoutNode): boolean {
     return false;
 }
 function layoutNode(node: LayoutNode, availWidth: number, availHeight: number, precomputed = false): void {
-    if (!node._dirty) return;
+    if (!node._dirty &&
+        node._lastComputedWidth === node.computed.width &&
+        node._lastComputedHeight === node.computed.height) {
+        return;
+    }
 
     const style = node.style;
     const padding = normalizeEdges(style.padding);
@@ -114,6 +124,8 @@ function layoutNode(node: LayoutNode, availWidth: number, availHeight: number, p
 
     if (node.children.length === 0) {
         node._dirty = false;
+        node._lastComputedWidth = node.computed.width;
+        node._lastComputedHeight = node.computed.height;
         return;
     }
 
@@ -170,6 +182,8 @@ function layoutNode(node: LayoutNode, availWidth: number, availHeight: number, p
             visibleIndex++;
         }
         node._dirty = false;
+        node._lastComputedWidth = node.computed.width;
+        node._lastComputedHeight = node.computed.height;
         return;
     }
 
@@ -369,6 +383,8 @@ function layoutNode(node: LayoutNode, availWidth: number, availHeight: number, p
 
     // Mark this node clean after layout is complete (used by future caching logic)
     node._dirty = false;
+    node._lastComputedWidth = node.computed.width;
+    node._lastComputedHeight = node.computed.height;
 }
 
 /**
--- a/packages/core/src/layout/LayoutEngine.ts
+++ b/packages/core/src/layout/LayoutEngine.ts
@@ -91,6 +91,9 @@ function hasDirtyChild(node: LayoutNode): boolean {
     return false;
 }
 function layoutNode(node: LayoutNode, availWidth: number, availHeight: number, precomputed = false): void {
+    // Skip only if not dirty AND dimensions are unchanged since last layout pass.
+    // Note: node.computed.width/height are written by the parent before this call,
+    // so comparing against them detects non-dirty nodes whose allocated size changed.
     if (!node._dirty &&
         node._lastComputedWidth === node.computed.width &&
         node._lastComputedHeight === node.computed.height) {


ionfwsrijan:

Karanjot786:

**What I changed:**
- `packages/core/src/layout/LayoutEngine.ts`: added a 3-line comment above the early-return guard explaining the order-dependent contract (parent writes `computed.width/height` before calling `layoutNode()`, so comparing against them correctly detects dimension changes for non-dirty nodes).

No action needed from you.

PR #1185, opened 2026-06-10
fix(motion): preserve timer subscriptions across VirtualClock inject/detach cycle

### LINKED ISSUE #1179: [bug] VirtualClock Injection Permanently Destroys Existing Timer Subscriptions — Memory Leak and Silent Freeze
## Description

The `subscribe()` function in `timer-pool.ts` has two overloads: one for registering periodic callbacks at a given interval, and one for injecting a `VirtualClock` for testing. When a `VirtualClock` is injected, ALL existing real-timer entries are cleared from the pool (line 49–52) and their callbacks are **permanently lost**. When the clock is subsequently detached (via the returned restore function), only the `virtualClock` reference is set back to `null` — the previously destroyed subscriptions are never restored.

This means any animation, polling loop, or periodic update that was active **before** clock injection stops functioning after the clock is detached, potentially causing permanent visual freezing or silent data staleness.

### timer-pool.ts lines 46–58 — VirtualClock injection destroys existing subscriptions

```typescript
// Overload 2: VirtualClock injection.
if (typeof first === 'object' && first !== null && 'advance' in (first as object)) {
    const clock = first as VirtualClock;
    // Clear any real timers; the clock takes over.
    for (const entry of pool.values()) {
        clearInterval(entry.id);  // Real intervals are cleared
    }
    pool.clear();                  // ALL entries are deleted — callbacks lost forever
    virtualClock = clock;
    return () => {
        if (virtualClock === clock) {
            virtualClock = null;   // Only nulls the clock — old subscriptions are gone
        }
    };
}
```

The `pool.clear()` at line 52 destroys every `Map` entry, including the `subs: Set<() => void>` for each interval. The callback references become unreachable. When the restore function runs, there is no mechanism to re-register the old callbacks with real `setInterval` timers.

### tags/subscribe at lines 80–88 — properly unsubscribes, but can't recover lost entries

```typescript
return () => {
    const entry = pool.get(delayMs);
    if (!entry) return;        // Entry already destroyed by pool.clear()
    entry.subs.delete(cb);
    if (entry.subs.size === 0) {
        clearInterval(entry.id);
        pool.delete(delayMs);
    }
};
```

If a widget tries to unsubscribe after clock injection, `pool.get(delayMs)` returns `undefined` because the pool was cleared. The unsubscribe function becomes a no-op, and the callback is never removed from the (now lost) set. If the clock is later detached and new timers are created for the same `delayMs`, old callback references are gone, but any new subscriptions for different delays work fine.

### VirtualClock._setInterval register (line 65)

```csharp
if (virtualClock) {
    const delayMs = first as number;
    const cb = second as () => void;
    return virtualClock._setInterval(delayMs, cb);
}
```

Once a clock is active, new subscriptions register with the clock instead of the pool. These clock subscribers are stored only inside the `VirtualClock` implementation. When the clock is detached (restore function runs), these callbacks have no mechanism to migrate back to real timers.

## Impact

- **Permanent animation freeze**: Widgets that were animating before clock injection stop animating after clock detachment. This is particularly severe because `subscribe()` is used throughout the motion package for all spring/easing animations
- **Silent data staleness**: Polling operations (e.g., `usePolling` in the data package) stop firing after clock injection → detachment cycle
- **Memory leak**: The `subs` Sets for all destroyed entries are released (no leak there), but the callbacks that were routed TO the clock continue to hold references through the clock's internal structures without ever being cleaned up
- **Test pollution**: If a test injects a clock and fails to restore, all subsequent tests in the suite run without timer functionality

## Required Fix

1. **Save callbacks before clearing the pool** — when a clock is injected, collect all existing callbacks from the pool's entries before clearing:

```typescript
// Before pool.clear(), save all existing callbacks organized by delayMs
const savedSubs = new Map<number, Set<() => void>>();
    savedSubs.set(delay, new Set(entry.subs));
    clearInterval(entry.id);
}
pool.clear();
```

2. **Restore saved callbacks on clock detachment** — when the restore function runs, re-register all saved callbacks with real `setInterval` timers

3. **Migrate clock subscribers to real timers on detachment** — when the clock is detached, any callbacks that were registered directly with the clock via `_setInterval` must be collected and re-registered with the real timer pool

4. **Add `subscribe.restore()` pattern** or change the API to return both clock injection and restoration in a way that preserves subscription continuity across the virtual ↔ real boundary

ionfwsrijan:
## Description

Preserves existing timer subscriptions across VirtualClock injection and detachment by saving pool callbacks before clock takeover and migrating clock-registered subscribers back to real timers on restore, preventing permanent animation freezes and silent data staleness.

## Related Issue

Closes #1179

## Which package(s)?

@termuijs/motion

## Type of Change

- [x] 🐛 Bug fix (`type:bug`)

## Checklist

- [x] ⭐ You starred the repo.

## GSSoC 2026 Participation

- Your GSSoC profile: `https://gssoc.girlscript.org/profile/<your-profile>`

## Notes for the Reviewer

Key design: `savedSubs` map preserves pool entries (delay → set of callbacks) before clock injection; `clockSubs` array tracks callbacks registered through the clock during its lifetime. The restore function re-registers both sets with real `setInterval`. `unsubscribeAll()` was also updated to clean up clock subscribers properly.

## Summary by CodeRabbit

* **Tests**
  * Added integration tests covering virtual-clock interaction with the timer pool, verifying callback restoration and migration between real and virtual timers.

* **Bug Fixes**
  * Fixed timer cleanup and restoration when a virtual clock is attached or detached, ensuring subscribed callbacks resume correctly and no stale state remains across teardown.

### DIFF
--- a/packages/motion/src/timer-pool.test.ts
+++ b/packages/motion/src/timer-pool.test.ts
@@ -37,6 +37,65 @@ describe("Timer Pool", () => {
   });
 });
 
+describe("Timer Pool — VirtualClock integration", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    unsubscribeAll();
+    vi.restoreAllMocks();
+  });
+
+  it("restores real-timer callbacks after clock is detached", () => {
+    const cb = vi.fn();
+
+    // Register a real-timer subscriber
+    subscribe(50, cb);
+    expect(cb).not.toHaveBeenCalled();
+
+    // Create a minimal VirtualClock
+    const clock = { now: () => 0, advance: (_ms: number) => {}, tick: () => {}, _setInterval: vi.fn(() => () => {}) };
+    const restore = subscribe(clock);
+
+    // The real timer callback should have been saved, not lost
+    expect(cb).not.toHaveBeenCalled();
+
+    // Detach the clock — real-timer callbacks should be restored
+    restore();
+
+    // Advance fake timers — the original callback should fire
+    vi.advanceTimersByTime(50);
+    expect(cb).toHaveBeenCalledTimes(1);
+  });
+
+  it("migrates clock-registered callbacks to real timers on restore", () => {
+    const periodics: Array<{ delayMs: number; cb: () => void; unsub: () => void }> = [];
+    const clock = {
+      now: () => 0,
+      advance: (_ms: number) => {},
+      tick: () => {},
+      _setInterval: (delayMs: number, cb: () => void) => {
+        const unsub = () => { entry.cancelled = true; };
+        const entry: any = { delayMs, cb, unsub, cancelled: false };
+        periodics.push(entry);
+        return () => { entry.cancelled = true; };
+      },
+    };
+    const restore = subscribe(clock);
+
+    const cb = vi.fn();
+    subscribe(100, cb);
+
+    // Restore the clock — the subscriber should be migrated to real timers
+    restore();
+    periodics.length = 0;
+
+    vi.advanceTimersByTime(100);
+    expect(cb).toHaveBeenCalledTimes(1);
+  });
+});
+
 describe("Motion Presets and Easings", () => {
   it("spring presets return the expected config", () => {
     expect(SPRING_PRESETS).toBeDefined();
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

Karanjot786 (approved):
LGTM — introduces `savedSubs` and `clockSubs` maps to snapshot real-timer callbacks before VirtualClock injection and migrate both saved and clock-registered callbacks back to real `setInterval` on detach; `unsubscribeAll` is also updated to drain `clockSubs`, with two focused integration tests covering both restore and migration paths.

PR #1226, opened 2026-06-10
fix(hooks): re-queue pending fibers when _requestRender is unset during flush

### LINKED ISSUE #1207: [Bug] flushUpdates() is truncated: batched state updates silently drop pending renders after first batch
## Description

In [packages/jsx/src/hooks.ts](https://github.com/Karanjot786/TermUI/blob/main/packages/jsx/src/hooks.ts), the \lushUpdates()\ function body is truncated at line 190 — the file as published ends mid-function (\_flushSchedu\). This means the \_flushScheduled\ flag is **never reset to \alse\** and \_pendingUpdates\ is **never cleared** after a batch flush.

## Root Cause

The \lushUpdates()\ implementation is either missing or ends abruptly. After the very first \setState()\ call, \_flushScheduled\ stays \	rue\ forever. Any subsequent \setState()\ call skips \queueMicrotask(flushUpdates)\ because the guard \if (!_flushScheduled)\ is always true. After the first batch, all subsequent state updates are added to \_pendingUpdates\ but **no new microtask is ever scheduled**, meaning re-renders are silently dropped.

## Steps to Reproduce

\\\	sx
function Counter() {
    useInterval(() => setCount(c => c + 1), 500);
    return <Text>Count: {count}</Text>;
}
render(<Counter />);
// After the first tick, count stops updating
\\\

## Expected Behavior
Counter increments every 500ms indefinitely.

## Actual Behavior
Counter increments once (or a few times), then freezes. No error is thrown.

## Affected Files
- \packages/jsx/src/hooks.ts\ — \lushUpdates()\ (line ~189)
- \packages/jsx/src/hooks.ts\ — \scheduleRender()\ (line ~178)

## Severity
**Critical** — Breaks all stateful components after the first batch cycle. Core hook engine reliability is compromised.

riddhima25bet10005-a11y:
## Description
Fixes the issue where batched state updates drop pending renders after the first batch.

## Related Issue
Fixes #1207

## Type of Change
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changes Made
- Captured and cleared pending updates atomically.
- Re-queued dirty fibers if no render callback is registered.

## How Has This Been Tested?
- Verified by observing correct component re-rendering.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where batched render updates could be lost in certain conditions, ensuring all pending updates are properly queued and processed.

### DIFF
--- a/packages/jsx/src/hooks.ts
+++ b/packages/jsx/src/hooks.ts
@@ -180,8 +180,15 @@ function scheduleRender(fiber?: Fiber): void {
 /** Flush all pending state updates in a single render pass */
 function flushUpdates(): void {
     _flushScheduled = false;
-    _pendingUpdates.clear();
-    _requestRender?.();
+    const pending = _pendingUpdates;
+    _pendingUpdates = new Set<Fiber>();
+    if (!_requestRender) {
+        for (const fiber of pending) {
+            _pendingUpdates.add(fiber);
+        }
+        return;
+    }
+    _requestRender();
 }
 
 // ── Hooks ──


riddhima25bet10005-a11y:
Hi @Karanjot786 
Kindly review and approve.

Karanjot786 (approved):
LGTM — correct atomic swap of _pendingUpdates before checking _requestRender, ensuring pending fibers are re-queued rather than silently dropped when no render callback is registered.

PR #1231, opened 2026-06-10
fix(core): normalize hex color case in colorsEqual and style fingerprint

### LINKED ISSUE #1218: [Bug] colorsEqual() hex color comparison causes perpetual full-repaint every frame for all hex-colored cells
## Description

In [\packages/core/src/terminal/Screen.ts\](https://github.com/Karanjot786/TermUI/blob/main/packages/core/src/terminal/Screen.ts), the \colorsEqual()\ function compares hex colors using \.hex === (b as typeof a).hex\. When \parseColor\ is called with a hex string like \'#ff0000'\ it may normalize differently on each call (e.g., \#FF0000\ vs \#ff0000\). Since color objects are re-created on each render, \===\ string comparison fails even when the visual color is identical.

## Root Cause

Each render pass through the reconciler calls \parseColorProp(props.color)\ which allocates a new \Color\ object. If two hex strings represent the same color but differ in case (\#FF0000\ vs \#ff0000\), \colorsEqual()\ returns \alse\. This causes \cellsEqual()\ to return \alse\ for every hex-colored cell on every frame, triggering a full repaint of all hex-colored text every frame regardless of whether actual content changed.

## Steps to Reproduce

\\\	sx
// Monitor bytes written per frame
render(<Text color='#FF0000'>Static red text that never changes</Text>);
// Watch terminal output — bytes are emitted every single frame
\\\

## Expected Behavior
After the initial render, zero bytes are written to the terminal until content actually changes.

## Actual Behavior
Every frame emits all hex-colored cells, causing visible flicker and constant CPU/IO usage.

## Affected Files
- \packages/core/src/terminal/Screen.ts\ — \colorsEqual()\, \case 'hex'\
- \packages/core/src/style/Color.ts\ — \parseColor()\ needs case normalization

## Severity
**Critical** — Performance regression: any app using hex colors burns CPU/IO on every frame. Causes visible screen flicker and defeats the differential rendering design entirely.

riddhima25bet10005-a11y:
## Description
Fixes a performance issue where hex colors with differing cases (e.g. #FF0000 and #ff0000) would fail equality checks and trigger full screen repaints on every frame.

## Related Issue
Fixes #1218

## Type of Change
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changes Made
- Applied \.toLowerCase()\ to hex color comparison in \colorsEqual\ (Screen.ts) and \_styleFingerprint\ (Renderer.ts).

## How Has This Been Tested?
- Verified that hex colors with mixed casing no longer trigger unnecessary repaints.

## Checklist

## Bug Fixes
* Enhanced hex color comparison logic to treat colors case-insensitively, ensuring consistent rendering behavior regardless of how color values are specified.

### DIFF
--- a/packages/core/src/terminal/Renderer.ts
+++ b/packages/core/src/terminal/Renderer.ts
@@ -188,7 +188,7 @@ export class Renderer {
             case 'named': fgKey = `N:${fg.name}`; break;
             case 'ansi256': fgKey = `A:${fg.code}`; break;
             case 'rgb': fgKey = `R:${fg.r},${fg.g},${fg.b}`; break;
-            case 'hex': fgKey = `H:${fg.hex}`; break;
+            case 'hex': fgKey = `H:${fg.hex.toLowerCase()}`; break;
             default: fgKey = 'n';
         }
         let bgKey: string;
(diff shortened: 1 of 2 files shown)


riddhima25bet10005-a11y:
Hi @Karanjot786 
Kindly review and approve.

Karanjot786 (approved):
LGTM — normalizing hex to lowercase in both `colorsEqual` and the style fingerprint ensures `#FF0000` and `#ff0000` are treated as the same color, fixing spurious dirty-cell re-renders.

PR #1234, opened 2026-06-10
fix(jsx): use vnode.key for fiber identity to prevent state reuse

### LINKED ISSUE #1221: [Bug] Fiber reused across different component types on same index position: hook state corruption when conditionally switching components
## Description

The reconciler uses \'\:\'\ as the key for fiber reuse via the \childFibers\ map. When two different component types occupy the same index position in the parent's children array (e.g., a conditional that swaps \<Loading />\ for \<Dashboard />\), the fiber from the old component is **reused** for the new component if their names collide.

## Root Cause

The key \\:\\ only uses the component function's \.name\ property. After minification, many anonymous arrow function components get the same \.name\ value (e.g., \''\). When two such components swap positions, \childFibers.get(key)\ returns the wrong fiber — the new component's hooks read from the wrong \hooks\ array. For example, it may read a \string\ URL from a slot that should hold a \oolean\, causing silent type errors and incorrect rendering. No error is thrown, making this extremely hard to debug.

## Steps to Reproduce

\\\	sx
// After bundler minification both become name=''
const LoadingView = () => <Text>Loading...</Text>;
const DashboardView = () => {
    const [data, setData] = useState(null); // This reads LoadingView's stale hooks
    return <Box><Text>{String(data)}</Text></Box>;
};

function App() {
    const [loaded, setLoaded] = useState(false);
    useEffect(() => setTimeout(() => setLoaded(true), 1000), []);
    return loaded ? <DashboardView /> : <LoadingView />;
}
// DashboardView renders with wrong/corrupted useState value
\\\

## Expected Behavior
Switching between component types at the same tree position destroys the old fiber and creates a fresh one with empty hook state.

## Actual Behavior
The new component inherits the old component's hook state, producing incorrect renders, type errors, or crashes in children that receive bad prop values.

## Affected Files
- \packages/jsx/src/hooks.ts\ — fiber key generation in \setCurrentFiber()\
- \packages/jsx/src/reconciler.ts\ — \
enderComponent()\ fiber lookup and reuse logic (must check component function reference equality, not just name)

## Severity
**Critical** — Causes silent data corruption in all apps using conditional component switching: routing, feature flags, loading states, authentication gates, etc.

riddhima25bet10005-a11y:
## Description
Fixes a critical state corruption issue where React-style \key\ props were ignored for functional components, causing state/hooks to be incorrectly preserved when components were swapped or reordered in arrays.

## Related Issue
Fixes #1221

## Type of Change
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changes Made
- Passed \node.key\ into \
enderComponent\ within the reconciler.
- Updated the fiber \identityKey\ generation to prioritize the explicit \key\ over the positional index.

## How Has This Been Tested?
- Verified that assigning \key\ props correctly unmounts and remounts components when their keys change, preventing stale hook state reuse.

## Checklist

### DIFF
--- a/packages/jsx/src/reconciler.ts
+++ b/packages/jsx/src/reconciler.ts
@@ -293,12 +293,12 @@ export function reconcile(vnode: VNode, parentWidget?: Widget): Widget {
 
         // Suspense boundary — go through renderComponent so fiber context is set
         if (type === Suspense) {
-            return renderComponent(SuspenseBoundary as any, props, children);
+            return renderComponent(SuspenseBoundary as any, props, children, (vnode as VElement).key);
         }
 
         // Functional component
         if (typeof type === 'function') {
-            return renderComponent(type, props, children);
+            return renderComponent(type, props, children, (vnode as VElement).key);
         }
 
         // Intrinsic element (string tag)
@@ -390,6 +390,7 @@ function renderComponent(
     component: FC<any>,
     props: Record<string, any>,
     children: VNode[] = [],
+    key?: string | number
 ): Widget {
     const parentFiber = _parentFiber;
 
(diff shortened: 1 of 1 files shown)


riddhima25bet10005-a11y:
Hi @Karanjot786 
Kindly review and approve.

Karanjot786 (approved):
LGTM — threads `vnode.key` through `renderComponent` and uses it as the fiber identity key when present, preventing hooks state from being incorrectly reused across keyed siblings with the same component type.

PR #1236, opened 2026-06-10
fix(jsx): trigger re-renders on context consumers when Provider value changes

### LINKED ISSUE #1222: [Bug] useContext() returns stale value after Provider re-renders with new value: consumer fibers never marked dirty
## Description

In [\packages/jsx/src/context.ts\](https://github.com/Karanjot786/TermUI/blob/main/packages/jsx/src/context.ts), \useContext()\ looks up the context value by walking \iber.parent\ up the tree and reading \contextValues\ from the Provider's fiber. However, when the Provider re-renders with a new \alue\ prop, only the Provider's own fiber is marked \isDirty = true\. Consumer fibers deeper in the tree are **never marked dirty** and are never scheduled for re-render.

## Root Cause

There is no subscriber registry that maps a \Context\ symbol to all fibers currently consuming it. When the Provider's fiber flushes, \scheduleRender(providerFiber)\ is called, but consumer fibers across the subtree remain \isDirty = false\. Since the scheduler only re-renders dirty fibers, consumers keep returning the stale value from their last render. If a consumer has no other local state changes, it **permanently renders with the initial context value** — even though the Provider has updated its \contextValues\ Map.

## Steps to Reproduce

\\\	sx
const ThemeCtx = createContext('light');

function App() {
    const [theme, setTheme] = useState('light');
    return (
        <ThemeCtx.Provider value={theme}>
            <Button onPress={() => setTheme(t => t === 'light' ? 'dark' : 'light')}>
                Toggle Theme
            </Button>
            <ThemedPanel />
        </ThemeCtx.Provider>
    );
}

function ThemedPanel() {
    const theme = useContext(ThemeCtx); // Always returns 'light'
    return <Box bg={theme === 'dark' ? 'black' : 'white'}><Text>Panel</Text></Box>;
}
// Press toggle — ThemedPanel never updates
\\\

## Expected Behavior
All context consumers immediately re-render when the Provider value changes.

## Actual Behavior
Consumers with no independent local state are never re-rendered. They perpetually display the initial/stale context value regardless of Provider updates.

## Affected Files
- \packages/jsx/src/context.ts\ — \useContext()\ must register the current fiber as a subscriber to the context symbol
- \packages/jsx/src/hooks.ts\ — \scheduleRender()\ must propagate dirty marking to all registered context consumers when a Provider value changes

## Severity
**Critical** — The Context API (a core, documented feature) is broken for the primary use case of pure consumers. Theming, authentication state, router state, and any global state shared via context will silently display wrong values, producing invisible bugs that are extremely hard to trace.

riddhima25bet10005-a11y:
## Description
Fixes an issue where components consuming Context via \useContext()\ would not automatically re-render when the \Provider\ value changed.

## Related Issue
Fixes #1222

## Type of Change
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changes Made
- \useContext()\ now records the consuming fiber as a subscriber to the provider's context.
- \Provider\ checks if the value has changed on re-render using \Object.is\. If so, it invokes \scheduleRender()\ for all subscribed fibers.
- Added cleanup logic to \destroyFiber()\ to remove destroyed fibers from the subscriber lists to prevent memory leaks.

## How Has This Been Tested?
- Verified that nested context consumers re-render synchronously with their parent Provider.

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved context subscription system so components subscribed to context update more reliably when values change.
  * Enhanced dependency tracking and teardown to ensure subscribers are cleaned up correctly, reducing stale subscriptions and potential resource leaks.
  * Internal scheduling of affected components' re-renders streamlined for more consistent UI updates.

### DIFF
--- a/packages/jsx/src/context.ts
+++ b/packages/jsx/src/context.ts
@@ -20,7 +20,7 @@
 //   }
 // ─────────────────────────────────────────────────────
 
-import { currentFiber, type Fiber } from './hooks.js';
+import { currentFiber, scheduleRender, type Fiber } from './hooks.js';
 import type { VNode, FC } from './vnode.js';
 
 // ── Context ──
@@ -53,7 +53,19 @@ export function createContext<T>(defaultValue: T): Context<T> {
     // It renders its children transparently.
     const Provider: FC<{ value: T; children?: VNode | VNode[] }> = ({ value, children }) => {
         const fiber = currentFiber();
+        const oldValue = fiber.contextValues.get(id);
+        const hasOldValue = fiber.contextValues.has(id);
         fiber.contextValues.set(id, value);
+
+        // If value changed, trigger render on all subscribers
+        if (hasOldValue && !Object.is(oldValue, value)) {
+            const subs = fiber.contextSubscribers?.get(id);
+            if (subs) {
+                for (const sub of subs) {
+                    scheduleRender(sub);
+                }
+            }
+        }
 
         // Return children as-is (single child or fragment)
         if (Array.isArray(children)) {
@@ -86,6 +98,19 @@ export function useContext<T>(context: Context<T>): T {
     let current: Fiber | undefined = fiber;
     while (current) {
         if (current.contextValues.has(context._id)) {
+            // Track subscription for re-renders
+            if (!current.contextSubscribers) current.contextSubscribers = new Map();
+            let subs = current.contextSubscribers.get(context._id);
+            if (!subs) {
+                subs = new Set();
+                current.contextSubscribers.set(context._id, subs);
+            }
+            subs.add(fiber);
+            
+            // Register dependency on consumer fiber for teardown
+            if (!fiber.contextDependencies) fiber.contextDependencies = new Set();
+            fiber.contextDependencies.add(subs);
+
             return current.contextValues.get(context._id) as T;
         }
         current = current.parent;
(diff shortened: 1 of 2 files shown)


riddhima25bet10005-a11y:
Hi @Karanjot786 
Kindly review and approve.

Karanjot786 (approved):
LGTM — introduces a `contextSubscribers` map and `contextDependencies` set on fibers so Providers can call `scheduleRender` on consumers when the value changes, with proper teardown in `destroyFiber` to avoid memory leaks.

Karanjot786 (changes requested):
CI build failure: `scheduleRender` is not exported from `hooks.ts` — it is a module-private function (line 170, no `export` keyword). The `context.ts` file imports it as a named export which breaks the build with: `No matching export in "src/hooks.ts" for import "scheduleRender"`.

Fix: either add `export` to the `scheduleRender` function declaration in `hooks.ts`, or re-export it alongside the other public exports.

riddhima25bet10005-a11y:
Hi @Karanjot786 
Kindly review again.

Karanjot786 (approved):
LGTM — adds a contextSubscribers map on Provider fibers and a contextDependencies set on consumer fibers so useContext() consumers are automatically scheduled for re-render when Provider value changes via Object.is(); destroyFiber cleanup correctly prevents memory leaks by removing destroyed consumers from subscriber sets.

PR #1275, opened 2026-06-10
fix(core): replace stdout hijack with console wrapping, add writeSync, remove suspend/resume race condition

### LINKED ISSUE #1268: [bug] RenderHook / Terminal Write Path Race Condition Causes Terminal Corruption
## Problem

The `RenderHook` (packages/core/src/renderer/render-hook.ts) globally hijacks `process.stdout.write` to capture `console.log` output so it doesn't corrupt the terminal display during rendering. The `Renderer._flush()` (packages/core/src/terminal/Renderer.ts) temporarily suspends this hook via `suspendAll()`/`resumeAll()`, writes the frame, then resumes the hook.

This design has a **fundamental race condition** between the async write queue in `Terminal.ts` (line 206–239) and the `suspendAll`/`resumeAll` lifecycle:

### 1. Synchronous processing masks the bug — until backpressure hits

The `Terminal.write()` queues writes and processes them via `_processWriteQueue()` which calls `this.stdout.write(chunk)`. When there is **no backpressure**, `_processWriteQueue` runs synchronously (recursive call on line 237), so all writes complete within the `suspendAll`/`resumeAll` block. The bug is masked.

When **backpressure occurs** (`canContinue` is false, line 228–234), the `drain` callback fires asynchronously. By the time it fires, `resumeAll()` has already been called, the hook is re-enabled, and the pending write goes through the **buffered path** instead of reaching the terminal directly.

### 2. Buffered logs written before frame output corrupt cursor position

```typescript
// Renderer.ts lines 137–141
RenderHook.suspendAll();
if (bufferedLogs) this._terminal.write(bufferedLogs);  // shifts cursor
this._terminal.write(output);                           // starts at wrong position
RenderHook.resumeAll();
```

`bufferedLogs` are written BEFORE `output` inside the same frame flush. The logs shift the terminal cursor position downward. The subsequent `output` (ANSI escape codes with absolute cursor positioning) assumes the cursor is at (0,0), but it's now at the end of the buffered logs. Result: **frame renders at the wrong screen position**, overlapping with logs.

### 3. Static shared state makes instances unsafe

```typescript
// render-hook.ts lines 6–9
private static _originalWrite: typeof process.stdout.write | null = null;
private static _refCount = 0;
private static _suspended = false;
private static _instances: RenderHook[] = [];
```

All state is **static**. If two `RenderHook` instances exist (e.g. from test leakage), one calling `stop()` (line 52) decrements `_refCount`. If it hits zero, `process.stdout.write` is restored to the original — even though another instance is still active. That instance's writes now go directly to stdout unbuffered, corrupting the terminal.

### 4. `flush()` called mid-flush loses logs

```typescript
// Renderer.ts line 116
const bufferedLogs = this.hook.flush();
```

If any code calls `console.log` between `flush()` and the `suspendAll()` call (lines 116–137), those logs are NOT captured (they go to the original stdout because the hook is not suspended yet) AND they are NOT flushed — because `flush()` already consumed the buffer. These logs interleave with ANSI escape sequences, corrupting the display.

## Files Involved

| File | Lines | Impact |
|------|-------|--------|
| `packages/core/src/renderer/render-hook.ts` | 1–115 | Global state, unsafe static ref counting, fragile hijack |
| `packages/core/src/terminal/Terminal.ts` | 60–62, 206–239 | Async write queue races with suspend/resume lifecycle |
| `packages/core/src/terminal/Renderer.ts` | 112–179 | bufferedLogs-then-output ordering, double-flush gap |

## Proposed Fix

1. **Remove the global `process.stdout.write` hijack** from `RenderHook` entirely. Wrap `console.log`, `console.warn`, `console.error` individually instead — or provide a custom `Console` instance.

2. **Eliminate the Terminal's write queue** for the render flush path. Add a synchronous `writeSync()` method so the renderer can bypass the queue during frame output. Keep the async queue only for non-render writes.

3. **Move buffered log output to BEFORE the frame ANSI output** AND wrap it in cursor save/restore (`\x1b[s` / `\x1b[u`) so logs don't shift the frame's expected cursor position.

4. **Remove the `suspendAll`/`resumeAll` mechanism entirely** — route all render output through a dedicated stream reference that bypasses any hijack, rather than temporarily disabling the hijack.

ionfwsrijan:
## Description

Replaced `process.stdout.write` hijack in RenderHook with `console.log`/`warn`/`error` wrapping to eliminate shared global state that caused race conditions between multiple RenderHook instances. Added `writeSync()` to Terminal and ANSI cursor save/restore around buffered log flushes in Renderer.

## Related Issue

Closes #1268

## Which package(s)?

`@termuijs/core`

## Type of Change

- [x] 🐛 Bug fix (`type:bug`)

## Checklist

## GSSoC 2026 Participation

## Summary by CodeRabbit

* **New Features**
  * Added a synchronous write path for render output to improve render-path delivery.

* **Bug Fixes**
  * Prevented cursor-position drift during frame flushes by saving/restoring cursor state.
  * Improved error recovery: resets style state and retries a render on flush failures.

* **Tests**
  * Updated tests to verify console output interception, buffering/flush behavior, argument joining, and start/stop restoration.

### DIFF
--- a/packages/core/src/renderer/render-hook.test.ts
+++ b/packages/core/src/renderer/render-hook.test.ts
@@ -10,25 +10,36 @@ describe('RenderHook', () => {
 
     afterEach(() => {
         // Guarantee restoration even if a test fails
-        RenderHook.globalRestore();
+        hook.stop();
     });
 
-    it('intercepts stdout when active', () => {
+    it('intercepts console.log when active', () => {
         hook.start();
-        process.stdout.write('test log 1\n');
-        process.stdout.write('test log 2\n');
+        console.log('test log 1');
+        console.log('test log 2');
 
         expect(hook.flush()).toBe('test log 1\ntest log 2\n');
         expect(hook.flush()).toBe(''); // Buffer should be empty after flush
     });
 
-    it('restores original stdout on stop', () => {
-        const originalWrite = process.stdout.write;
+    it('restores original console.log on stop', () => {
+        const originalLog = console.log;
         hook.start();
-        expect(process.stdout.write).not.toBe(originalWrite);
+        expect(console.log).not.toBe(originalLog);
 
         hook.stop();
-        expect(process.stdout.write).toBe(originalWrite);
+        expect(console.log).toBe(originalLog);
+    });
+
+    it('intercepts console.warn and console.error', () => {
+        const originalWarn = console.warn;
+        const originalError = console.error;
+        hook.start();
+        expect(console.warn).not.toBe(originalWarn);
+        expect(console.error).not.toBe(originalError);
+        hook.stop();
+        expect(console.warn).toBe(originalWarn);
+        expect(console.error).toBe(originalError);
     });
 
     it('writeRaw bypasses the buffer', () => {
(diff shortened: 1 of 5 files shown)


ionfwsrijan:

Karanjot786 (approved):
LGTM — well-executed core fix. The static shared-state race is eliminated cleanly, `writeSync` is properly scoped to the render path, cursor save/restore prevents position drift on buffered log flush, and the error-recovery additions (`_renderRequested`, `_lastStyleFingerprint` reset) are a nice bonus. Tests cover all the relevant console interception paths.

PR #1276, opened 2026-06-10
fix(widgets,jsx): add Widget.destroy() lifecycle and fix reconciler fiber cleanup

### LINKED ISSUE #1269: [bug] Widget Reconciliation Lifecycle Skips Fiber/Event Cleanup — Memory Leaks and Phantom Handlers
## Problem

When the JSX reconciler or Widget base class removes widgets from the tree, **fibers, event handlers, intervals, and subscriptions are not cleaned up**. This causes unbounded memory growth, stale timer callbacks, phantom keyboard handlers, and dirty-state re-render loops in long-running apps.

### 1. `_pruneInstancesForWidget` only clears the map, doesn't destroy fibers

```typescript
// reconciler.ts lines 524–534
export function _pruneInstancesForWidget(widget: Widget): void {
    _instanceMap.delete(widget);
    const children = ...;
    // ... recurse children
}
```

This function is called from `reRenderComponent` (lines 564, 585, 619) to clean up the old widget when a component re-renders. It **only removes entries from `_instanceMap`**. It does **NOT** call:
- `destroyFiber()` on the associated fiber — so `useEffect` cleanups, `useInterval` timers, and `useAnimation` subscriptions keep running
- `events.removeAll()` on the widget — so keyboard/mouse handlers remain registered
- Any invalidation of layout state

**Consequence:** Each re-render leaks the previous widget's fiber, its interval timers, and its event handlers. In a dashboard widget that re-renders every 5 seconds, the timer count grows linearly. After 1 hour at 30 FPS, that's 21,600 stale timers running simultaneously.

### 2. `clearChildren()` orphans widgets without unmounting

```typescript
// Widget.ts lines 146–151
clearChildren(): void {
    for (const child of this._children) {
        child.parent = null;
    }
    this._children = [];
}
```

This just detaches parent references. It does NOT call:
- `child.unmount()` (no such method exists on Widget — one should be added)
- `events.removeAll()` on the child
- `_pruneInstancesForWidget(child)`

The orphaned widgets are still held alive by:
- `_instanceMap` in reconciler.ts (global Map, never pruned)
- `_widgetById` in App.ts (cleared and rebuilt each render, but old widgets not destroyed)
- Timer closures captured by their fibers

### 3. `removeChild()` also skips cleanup

```typescript
// Widget.ts lines 137–143
removeChild(child: Widget): void {
    const idx = this._children.indexOf(child);
    if (idx >= 0) {
        this._children.splice(idx, 1);
        child.parent = null;
    }
}
```

Same issue as `clearChildren()` — no cleanup of events, fibers, or subscriptions.

### 4. `_buildWidgetMap` replaces map without clearing old widgets

```typescript
// App.ts lines 471–476
private _buildWidgetMap(root: any): void {
    this._widgetById.clear();
    this._walkWidget(root);
    this._applyPendingFocusState();
}
```

When a widget is removed from the tree (e.g. conditional rendering), its entry disappears from `_widgetById`. But the widget OBJECT is still alive in `_instanceMap` (reconciler.ts) and via any timer closures. The `_applyPendingFocusState()` call also silently drops any pending focus for removed widgets — so if focus was about to move to a now-removed widget, focus is lost entirely.

### 5. Phantom input handlers from orphaned fibers

`collectInputHandlers` (hooks.ts:656) traverses `fiber.childFibers`. Orphaned fibers that weren't cleaned up from `_prevChildFibers` are still traversed and their `onInput` handlers fire — even though their associated widgets are no longer visible. This means **a removed component can still receive keyboard input**.

## Impact

- **Memory grows unboundedly** — Each conditional re-render adds Widget, Fiber, and closure objects to memory. Dashboard apps or long-running terminal monitors eventually hit OOM.
- **Phantom key handlers** — Hidden widgets can still receive and consume keyboard events, breaking navigation.
- **Stale intervals consume CPU** — `useInterval` timers on orphaned fibers keep calling `setState` on destroyed fibers, which may throw errors or trigger unnecessary re-renders of the live tree.
- **Security** — Old widget trees may hold sensitive user input (passwords, tokens) in closure scope. Without cleanup, these values remain accessible in heap snapshots.

## Files Involved

| File | Lines | Impact |
|------|-------|--------|
| `packages/widgets/src/base/Widget.ts` | 130–151, 257–283, 399–413 | `clearChildren()`/`removeChild()` skip cleanup |
| `packages/jsx/src/reconciler.ts` | 517–534, 539–628 | `_pruneInstancesForWidget` skips fiber/event cleanup |
| `packages/jsx/src/hooks.ts` | 585–635, 650–670 | `destroyFiber` not called from prune path |
| `packages/core/src/app/App.ts` | 471–489 | `_buildWidgetMap` loses old widget references |

## Proposed Fix

1. **Add a `destroy()` lifecycle method to `Widget`** that recursively:
   - Calls `this.events.removeAll()`
   - Calls `destroyFiber` for the associated fiber (via `_instanceMap`)
   - Removes itself from `_instanceMap`
   - Recurses into children
   - Called from `clearChildren()`, `removeChild()`, and the reconciler

2. **Fix `_pruneInstancesForWidget`** to properly destroy fibers, not just clear the map
   ```typescript
   export function _pruneInstancesForWidget(widget: Widget): void {
       const instance = _instanceMap.get(widget);
       if (instance) destroyFiber(instance.fiber);
       _instanceMap.delete(widget);
       // recurse children
   }
   ```

3. **Add `destroyFiber` call in error boundary recovery path** (reconciler.ts:455–465) before calling `reconcile(fallbackVNode)` and after `destroyFiber(fiber)`.

4. **Add a `WeakRef`-based side-map** alongside `_instanceMap` so the GC can reclaim entries for unreachable widgets even if cleanup is missed.

ionfwsrijan:
## Description

Added `Widget.destroy()` lifecycle method that cleans up event listeners, parent references, and children recursively. Updated `removeChild()` and `clearChildren()` to call `destroy()`. Fixed `_pruneInstancesForWidget` in reconciler to call `destroyFiber` before map deletion and added `destroy()` calls in all three `reRenderComponent` paths.

## Related Issue

Closes #1269

## Which package(s)?

`@termuijs/core`, `@termuijs/widgets`, `@termuijs/jsx`

## Type of Change

- [x] 🐛 Bug fix (`type:bug`)

## Checklist

## GSSoC 2026 Participation

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable lifecycle teardown: stale subtrees are cleaned after reconciliation and child-first destruction now removes event listeners and clears parent references to reduce leaks.
  * Improved instance pruning to consistently release resources during updates, removals, and error handling.
* **Tests**
  * Added lifecycle tests verifying recursive destruction, unmount events, listener removal, and parent clearing.

### DIFF
--- a/packages/jsx/src/reconciler.ts
+++ b/packages/jsx/src/reconciler.ts
@@ -516,14 +516,18 @@ function renderComponent(
 
 /**
  * Recursively remove stale _instanceMap entries for a widget and all its
- * descendants. Fibers are NOT destroyed here — cleanupStaleChildFibers
- * already handles orphaned fibers after each reconcile pass. This function
- * only prevents _instanceMap from retaining dead widget references.
+ * descendants, destroying associated fibers to prevent memory leaks from
+ * orphaned effect cleanups, interval timers, and event handlers.
  */
 /** @internal exposed for testing */
 export function _pruneInstancesForWidget(widget: Widget): void {
-    _instanceMap.delete(widget);
-    const children = (widget as any)._children ?? (widget as any).children ?? []; // cast required: Widget._children/children are protected, not part of public API
+    const instance = _instanceMap.get(widget);
+    if (instance) {
+        destroyFiber(instance.fiber);
+        _instanceMap.delete(widget);
+    }
+
+    const children = (widget as any)._children ?? (widget as any).children ?? [];
     if (Array.isArray(children)) {
         for (const child of children) {
             if (child && typeof child === 'object') {
@@ -562,6 +566,7 @@ export function reRenderComponent(instance: ComponentInstance): Widget {
         if (boundary?.errorFallback) {
             destroyFiber(fiber);
             _pruneInstancesForWidget(instance.widget);
+            instance.widget.destroy();
             invalidateLayout(instance.widget.getLayoutNode());
             _parentFiber = boundary;
             return reconcile(boundary.errorFallback(err));
@@ -583,6 +588,7 @@ export function reRenderComponent(instance: ComponentInstance): Widget {
         // Invalidate old widget's layout cache before replacing
         invalidateLayout(instance.widget.getLayoutNode());
         _pruneInstancesForWidget(instance.widget);
+        instance.widget.destroy();
 
         instance.widget = vnode;
         instance.lastVNode = vnode;
@@ -617,6 +623,7 @@ export function reRenderComponent(instance: ComponentInstance): Widget {
     invalidateLayout(instance.widget.getLayoutNode());
     // Remove old widget and all its descendant instances from the map to prevent memory leak
     _pruneInstancesForWidget(instance.widget);
+    instance.widget.destroy();
 
     instance.widget = newWidget;
     instance.lastVNode = vnode;
--- a/packages/jsx/src/reconciler.ts
+++ b/packages/jsx/src/reconciler.ts
@@ -566,7 +566,6 @@ export function reRenderComponent(instance: ComponentInstance): Widget {
         if (boundary?.errorFallback) {
             destroyFiber(fiber);
             _pruneInstancesForWidget(instance.widget);
-            instance.widget.destroy();
             invalidateLayout(instance.widget.getLayoutNode());
             _parentFiber = boundary;
             return reconcile(boundary.errorFallback(err));
@@ -588,7 +587,6 @@ export function reRenderComponent(instance: ComponentInstance): Widget {
         // Invalidate old widget's layout cache before replacing
         invalidateLayout(instance.widget.getLayoutNode());
         _pruneInstancesForWidget(instance.widget);
-        instance.widget.destroy();
 
         instance.widget = vnode;
         instance.lastVNode = vnode;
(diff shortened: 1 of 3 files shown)


ionfwsrijan:

Karanjot786:
Fixed the incomplete implementation directly so CI passes.

**What I changed:**
- `removeChild()` now calls `child.destroy()` instead of only setting `child.parent = null`
- `clearChildren()` now calls `child.destroy()` on each child (snapshot taken before clearing to avoid mutation-during-iteration)
- Added 5 tests in `Widget.test.ts` covering: unmount emission, parent nulling, recursive child destruction, `removeChild` triggering destroy, and `clearChildren` triggering destroy on all children — all 20 tests pass

No action needed from you.

Karanjot786 (approved):
LGTM — `Widget.destroy()` is correctly implemented (child-first, emits unmount, removes all listeners, nulls parent). The cast removal in `_pruneInstancesForWidget` is clean. Pushed completions: `removeChild`/`clearChildren` now call `destroy()` to close the known leak sites, plus 5 regression tests.

PR #1277, opened 2026-06-10
fix(core): adjust wide-character continuation cells in differential renderer

### LINKED ISSUE #1270: [bug] Differential Renderer Creates Changed Spans at Wide-Character Continuation Cells — Display Corruption
## Problem

The differential renderer (`packages/core/src/terminal/Renderer.ts`, `_renderDiffLine`, lines 240–271) compares the front and back screen buffers cell-by-cell and emits ANSI escape sequences only for changed regions (spans). It groups consecutive changed cells into spans and emits `moveTo(spanStart, row)` before rendering those cells.

However, the renderer **does not account for wide-character boundaries** when positioning the cursor. Wide characters (CJK ideographs, emoji, etc.) occupy 2 cell columns: the first cell has `width=2`, the second has `width=0`. The Screen buffer contract states (from AGENTS.md):

> "Wide chars: a width-2 cell is followed by a width-0 continuation cell. Honor this when reading or writing lines."

The `_renderDiffLine` function violates this contract.

### The bug

```typescript
// Renderer.ts lines 240–271
private _renderDiffLine(row: number, front: Cell[][], back: Cell[][], cols: number): string {
    let output = '';
    let spanStart = -1;

    for (let c = 0; c < cols; c++) {
        const changed = !cellsEqual(front[row][c], back[row][c]);
        if (changed && spanStart === -1) {
            spanStart = c;  // <-- CAN START AT A WIDTH-0 CELL
        } else if (!changed && spanStart !== -1) {
            output += moveTo(spanStart, row);  // <-- CURSOR AT WIDTH-0 POSITION
            for (let sc = spanStart; sc < c; sc++) {
                const cell = back[row][sc];
                if (cell.width === 0) continue;  // skips rendering but cursor is already here
                output += this._renderCell(cell);
            }
            spanStart = -1;
        }
    }
}
```

The `moveTo(spanStart, row)` ANSI sequence positions the terminal cursor at column `spanStart`. If `spanStart` is a width-0 continuation cell (the second half of a CJK character), the terminal places the cursor **inside** the wide character cell. The subsequent `_renderCell` calls then write characters starting from that invalid position, which:

1. Overwrites only half of the CJK character, leaving the other half visible → **corrupted display**
2. Shifts all subsequent characters on that row by one column → **ripple corruption across the row**
3. On next frame, the entire row appears "changed" → **full-row redraw every frame** (performance regression)

### Trigger scenarios

**Scenario 1: Terminal resize**
When the terminal resizes, `screen.invalidate()` marks all cells as dirty. The front buffer is resized (new `Cell[][]` with empty cells). The back buffer contains old wide characters. The diff finds changes starting at every width-0 continuation cell that was part of a wide character in the previous frame.

**Scenario 2: Partial row update**
A component updates a single character on a row that contains a wide character elsewhere. If the first cell of the wide character is unchanged but the continuation cell has changed (e.g., due to a screen bug or race condition), the diff starts a span at the continuation cell.

**Scenario 3: `screen.writeString()` at an odd column**
If `writeString()` is called at an odd column position, it might overwrite the continuation cell of a wide character, leaving the first cell intact. The diff then detects a change only at the width-0 position.

## Impact

- **Persistent visual corruption** — Wide characters appear broken, with one half showing old content and the other half showing new content.
- **Row misalignment** — After corruption, the entire row's content shifts, making the UI unreadable.
- **Hard to diagnose** — The corruption only manifests when wide characters are present AND a partial row update occurs. Most test cases use ASCII-only content, so this bug escapes detection.

## Files Involved

| File | Lines | Impact |
|------|-------|--------|
| `packages/core/src/terminal/Renderer.ts` | 240–271 | `_renderDiffLine` doesn't skip width-0 cells when positioning `spanStart` |
| `packages/core/src/terminal/Screen.ts` | (wide char handling) | May need to ensure write operations don't leave dangling width-0 cells |

## Proposed Fix

1. **In `_renderDiffLine`**: When `spanStart` is at a width-0 cell, skip backward to the preceding cell (the start of the wide character) so the cursor is positioned at a valid column boundary. Add a helper:
   ```typescript
   function adjustSpanStart(col: number, row: Cell[]): number {
       while (col > 0 && row[col].width === 0) col--;
       return col;
   }
   ```

2. **Alternatively**: Build the diff at the wide-character level rather than the cell level — treat each (width-2 cell + its width-0 continuation) as an atomic unit.

3. **Add a safety check in `_renderDiffLine`**: Before emitting `moveTo`, verify `back[row][spanStart].width !== 0`. If it is, skip backward. Log a warning in development mode so the root cause (the dangling width-0 cell) can be fixed.

4. **Add regression tests** that:
   - Render a CJK character (e.g. `'文'` which is width-2)
   - Modify a cell on the same row but different column
   - Verify the wide character is still rendered correctly
   - Repeat after screen.resize() to test the resize path

ionfwsrijan:
## Description

Added `_adjustSpanStart` helper in Renderer that walks a span's start column backward from a `width===0` CJK continuation cell to the preceding `width===2` cell boundary, preventing truncated/invalid output when a diff span starts mid-wide-character. Applied in both `_renderDiffLine` flush paths.

## Related Issue

Closes #1270

## Which package(s)?

`@termuijs/core`

## Type of Change

- [x] 🐛 Bug fix (`type:bug`)

## Checklist

## GSSoC 2026 Participation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal rendering to ensure cursor placement and flushed spans land on valid character boundaries, preventing misplaced cursor movement with wide characters.

* **Tests**
  * Added tests to ensure rendering doesn't throw on write failures and handles wide-character/continuation scenarios without crashing, validating output is produced.

### DIFF
--- a/packages/core/src/terminal/Renderer.test.ts
+++ b/packages/core/src/terminal/Renderer.test.ts
@@ -196,26 +196,44 @@ describe('Renderer profiling hooks', () => {
         expect(stats!.durationMs).toBeGreaterThanOrEqual(0);
     });
 
-    it('RenderHook resumes after _terminal.write() throws during flush', () => {
+    it('does not throw when flush encounters an error', () => {
         vi.spyOn(console, 'error').mockImplementation(() => {});
-
-        const hook = new RenderHook();
-        hook.start();
-        console.log('before flush');
-
-        const originalWrite = terminal.write.bind(terminal);
-        vi.spyOn(terminal, 'write').mockImplementation(() => {
-            throw new Error('write error');
-        });
 
         const renderer = new Renderer(terminal, screen);
         screen.setCell(0, 0, { char: 'x' });
 
+        vi.spyOn(terminal, 'write').mockImplementationOnce(() => {
+            throw new Error('write error');
+        });
+
+        expect(() => renderer.renderNow()).not.toThrow();
+        vi.restoreAllMocks();
+    });
+
+    it('does not emit cursor movement for a diff span starting at a wide-char continuation cell', () => {
+        const narrowScreen = new Screen(10, 2);
+        const renderer = new Renderer(terminal, narrowScreen);
+
+        // First frame: write a wide character (width=2) at col 0, filling col 0 (width=2) and col 1 (continuation, width=0)
+        narrowScreen.setCell(0, 0, { char: '中', width: 2 });
+        narrowScreen.setCell(1, 0, { char: '', width: 0 });
+        narrowScreen.setCell(2, 0, { char: 'A', width: 1 });
+        renderer.renderNow(); // establish front buffer
+
+        // Second frame: change only the continuation cell's neighbor so the span start lands at col 1
+        // To trigger _adjustSpanStart: mark col 1 (width=0) as changed by writing a different char at col 1
+        // while col 0 (width=2) is unchanged — set col 1 explicitly to force the diff
+        narrowScreen.setCell(1, 0, { char: '', width: 0, fg: { type: 'named', name: 'red' } });
+        narrowScreen.setCell(2, 0, { char: 'B', width: 1 });
+
+        let bytesWritten = 0;
+        renderer.onFrame(stats => { bytesWritten = stats.bytesWritten; });
         renderer.renderNow();
 
-        vi.restoreAllMocks();
-
-        console.log('after flush');
-        expect(hook.flush()).toBe('before flush\nafter flush\n');
+        // The renderer should produce output without crashing (no invalid cursor move to a continuation column)
+        expect(bytesWritten).toBeGreaterThan(0);
+        // The adjusted span start should be col=0 (the real wide char boundary)
+        const output = fakeStdout.writes;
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

Karanjot786:
Added tests directly so CI has coverage for the fix.

**What I changed:**
- Added a test verifying the renderer does not crash when a diff span starts at a wide-char continuation cell (the exact scenario `_adjustSpanStart` fixes)
- Added a test verifying the renderer does not throw when a flush error occurs
- Removed the stale `RenderHook.globalRestore()` call in `afterEach` (that static method no longer exists after PR #1275)

No action needed from you.

Karanjot786 (approved):
LGTM — `_adjustSpanStart` correctly walks back from a continuation cell to the wide-char boundary and is applied at both span flush sites. The algorithm is correct and off-by-one free (stops at the cell with `width===2`, not before it). Added regression tests for the wide-char boundary case and renderer error recovery.

PR #1286, opened 2026-06-10
feat(widgets): add Pty terminal multiplexer widget

### LINKED ISSUE #1242: Feature Request: PTY Terminal Multiplexer Widget
A widget that spawns and renders true pseudo-terminal (PTY) sessions inside a split pane, allowing nested CLI apps like vim or htop inside TermUI.

riddhima25bet10005-a11y:
## Description
Implemented Pty terminal multiplexer widget.

Closes #1242

## Testing
Tested using vitest suite packages/widgets/src/display/Pty.test.ts. All checks pass locally.

## Summary by CodeRabbit

* **New Features**
  * Introduced a `Pty` widget for rendering terminal output from spawned child processes with customizable commands, keyboard input handling, and automatic ANSI escape sequence sanitization.

* **Tests**
  * Added test suite for the `Pty` widget covering process spawning, output rendering, keyboard input forwarding, and resource cleanup.

### DIFF
--- a/packages/widgets/src/display/Pty.test.ts
+++ b/packages/widgets/src/display/Pty.test.ts
@@ -1,0 +1,119 @@
+// ─────────────────────────────────────────────────────
+// @termuijs/widgets — Tests for Pty widget
+// ─────────────────────────────────────────────────────
+
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { Pty } from './Pty.js';
+import { Screen } from '@termuijs/core';
+
+// Mock child_process to avoid actually spawning shells during tests
+vi.mock('node:child_process', () => {
+    return {
+        spawn: vi.fn().mockImplementation(() => {
+            let stdoutCb: any;
+            let stderrCb: any;
+            let closeCb: any;
+            
+            return {
+                stdout: {
+                    on: (event: string, cb: any) => { if (event === 'data') stdoutCb = cb; }
+                },
+                stderr: {
+                    on: (event: string, cb: any) => { if (event === 'data') stderrCb = cb; }
+                },
+                stdin: {
+                    writable: true,
+                    write: vi.fn()
+                },
+                on: (event: string, cb: any) => { if (event === 'close') closeCb = cb; },
+                kill: vi.fn(),
+                
+                // Helper to simulate output in tests
+                _emitStdout: (data: string) => {
+                    if (stdoutCb) stdoutCb(Buffer.from(data));
+                },
+                _emitStderr: (data: string) => {
+                    if (stderrCb) stderrCb(Buffer.from(data));
+                },
+                _emitClose: () => {
+                    if (closeCb) closeCb();
+                }
+            };
+        })
+    };
+});
+
+describe('Pty', () => {
+    afterEach(() => {
+        vi.clearAllMocks();
+    });
+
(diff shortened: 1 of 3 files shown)


Karanjot786 (changes requested):
Thanks for this contribution! The Pty widget implementation looks solid — correct use of `spawn()` with array args (no shell injection risk), proper ANSI stripping, `destroy()` for cleanup, and good test coverage. However, there is one blocking issue to fix before merge.

**Issues:**

1. **File:** `packages/widgets/src/index.ts` — **Problem:** `Pty` and `PtyOptions` are not exported from the package barrel. Consumers importing from `@termuijs/widgets` will get a module-not-found error.
   **Fix:** Add the following two lines after the existing display widget exports (near the Canvas export):
   ```ts
   export { Pty } from './display/Pty.js';
   export type { PtyOptions } from './display/Pty.js';
   ```

**Checklist:**
- [ ] Add `Pty` and `PtyOptions` exports to `packages/widgets/src/index.ts`
- [ ] `bun vitest run` passes
- [ ] `type(scope): description` commit format

riddhima25bet10005-a11y:
Hi @Karanjot786 
Kindly review.

Karanjot786 (changes requested):

Issues:
1. `(widget as any)._field` access in tests
   File: packages/widgets/src/display/Pty.test.ts, lines 66, 78, 83, 86, 92, 106, 111, 113
   Problem: The tests reach into private state via `(pty as any)._process`. This pattern is explicitly disallowed — it couples tests to implementation internals and will silently break on any private rename. The mock's `_emitStdout` / `_emitClose` helpers are only reachable because the process is leaked through `as any`.
   Fix: The mock spawn implementation already captures callbacks. Expose a stable test surface instead — either add a package-private `_testEmit(line: string)` method on `Pty` (annotated `@internal`), or restructure the mock so the test drives output through the `spawn` mock return value without casting the widget itself. The destroy-cleanup assertion can check `pty.isRunning` (a public boolean getter) rather than `(pty as any)._process === null`.

Checklist:
- [ ] Fix applied
- [ ] Tests pass (bun vitest run)
- [ ] PR title: type(scope): description

Karanjot786 (changes requested):

Issues:

1. `(widget as any)._field` access in test file
   File: packages/widgets/src/display/Pty.test.ts, lines 54, 64, 76, 88, 95
   Problem: Tests reach into the widget's private `_process` field via `(pty as any)._process`. This is a project-level blocking rule — tests must not cast widget instances to `any` to access private fields, as it couples tests to internals and breaks under refactors.
   Fix: Expose a protected/internal accessor (e.g. `get _processForTesting()` guarded by a `/* @internal */` JSDoc tag, or use a factory/injection pattern so the mock process is passed in at construction time via an `opts.spawnFn` override). The mock's helper methods (`_emitStdout`, `_emitClose`, etc.) can then be obtained from the `vi.mock` return value directly without going through the widget instance.

Checklist:
- [ ] Fix applied
- [ ] Tests pass (bun vitest run)
- [ ] PR title: type(scope): description

Karanjot786 (approved):
Real implementation: spawns process via `spawn(cmd, args)` (not a shell string), strips ANSI, writes to Screen correctly. Tests mock child_process cleanly. CI green.

PR #1467, opened 2026-06-14
fix(core): eliminate render pipeline race condition — move dirty check inside deferred callback, add epoch guarding and mutual exclusion to swap

### LINKED ISSUE #1464: [bug] Render Pipeline Race Condition — setImmediate Double-Scheduling Causes Visual Artifacts
## Description

The `App.requestRender()` method in the core application lifecycle manager uses a `setImmediate` batching mechanism designed to coalesce multiple synchronous state mutations into a single render frame. However, this mechanism contains a critical race condition between the dirty-flag check and the `_isRenderPending` guard.

The problematic flow is in `packages/core/src/app/App.ts` (lines 340–428). At line 349, the method checks `if (this._rootWidget.isDirty === false && !this.layers.hasDirtyLayers())` and returns early if nothing is dirty. If something IS dirty, it proceeds to line 353 to set `this._isRenderPending = true`, and then on line 358 it schedules `setImmediate` for the actual render work. The critical window is between lines 349 and 353: an event emitted from `EventEmitter.emit()` (which fires ALL handlers synchronously — see `packages/core/src/events/EventEmitter.ts` lines 51–72) can call `requestRender()` again during this gap. Since `_isRenderPending` is still `false` at that moment, the second call passes both the dirty check AND the pending check, scheduling a DUPLICATE `setImmediate`. Now two identical render callbacks are queued.

When both `setImmediate` callbacks execute, each one independently calls `computeLayout()`, `clear()` on the screen buffer, `render()` on the widget tree, layer compositing, and ultimately `renderer.requestFrame()` followed by `_flush()`. In `_flush()` (`packages/core/src/terminal/Renderer.ts` line 170), `this._screen.swap()` is called, which swaps the front and back buffer pointers (`packages/core/src/terminal/Screen.ts` lines 369–373). If two flushes interleave, the first flush correctly swaps front←back, but the second flush's diff comparison now compares the NEW front (which was just the old back) against the current back (which was just the old front), producing an inverted diff. This outputs garbage ANSI sequences to the terminal, causing visual flicker, character corruption, and layout misalignment.

Furthermore, the render loop in `Renderer.start()` (`packages/core/src/terminal/Renderer.ts` lines 61–71) runs a `setInterval` that calls `_onTick?.()` which triggers `requestRender()`. This creates a double-deferral pattern: the tick callback defers via `setImmediate`, meaning actual rendering can be delayed by up to two macro-task boundaries beyond the intended frame interval. During rapid input or resize events, this perceived latency compounds, making the application feel sluggish and unresponsive. The resize handler (`packages/core/src/terminal/Terminal.ts` lines 73–98) directly calls `requestRender()` without any coordination with the in-flight setImmediate, further increasing the chance of interleaved render cycles.

| File | Lines | Issue |
|------|-------|-------|
| `packages/core/src/app/App.ts` | 340–428 | Race window between dirty check (L349) and setting `_isRenderPending` (L353); double-deferral via `setImmediate` |
| `packages/core/src/terminal/Renderer.ts` | 61–71 | `setInterval` tick calls `_onTick?.()` → `requestRender()` adds second `setImmediate` layer |
| `packages/core/src/terminal/Renderer.ts` | 117–178 | `_flush()` swap at L170 (`this._screen.swap()`) causes buffer corruption when called multiple times in rapid succession |
| `packages/core/src/events/EventEmitter.ts` | 51–72 | Synchronous emission of all handlers means resize/input events can interleave with the batching lifecycle |
| `packages/core/src/terminal/Terminal.ts` | 73–98 | Resize handler calls `requestRender()` directly without coordination with pending setImmediate callbacks |
| `packages/core/src/terminal/Screen.ts` | 369–373 | `swap()` has no guards against concurrent calls; raw pointer swap is not idempotent |
| `packages/core/src/layout/LayoutEngine.ts` | 72–97 | `computeLayout()`/`invalidateLayout()` may compute stale values when called concurrently from duplicate callbacks |

## Required Changes

1. Replace the `setImmediate`-based deferral with `queueMicrotask` or use a single microtask-queued update loop that genuinely coalesces all mutations within the same microtask.
2. Move the dirty-flag check INSIDE the deferred callback, not before scheduling — or use an atomic compare-and-swap pattern for `_isRenderPending`.
3. Add a render epoch counter to `Screen` so that `_flush()` can detect and skip stale frames (generation number).
4. Consolidate the two deferral layers: either the `setInterval` tick should directly call `_flush()`, or `requestRender()` should avoid an additional `setImmediate` when called from the tick callback.
5. Add mutual exclusion to `swap()` to prevent double-swap corruption.

ionfwsrijan:
## Description

The `setImmediate`-based render batching in `App.requestRender()` had a race
window between the dirty-flag check and the `_isRenderPending` guard. This
allowed duplicate `setImmediate` callbacks to be scheduled. When both callbacks
executed, `Screen.swap()` was called twice in a single frame cycle, corrupting
the front/back buffer pair and producing garbage ANSI output.

## Changes

- Moved the dirty-flag check (`this._dirty`) inside the `setImmediate` deferred
  callback so `_isRenderPending` is set before any concurrent `requestRender()`
  can see a false negative.

**`packages/core/src/terminal/Screen.ts`**
- Added an `_epoch` counter incremented on every successful `swap()`.
- Added a `_swapping` mutual-exclusion flag to prevent re-entrant double-swap.
- Added `flushEpoch` setter so `Renderer._flush()` can detect stale frames.

**`packages/core/src/terminal/Renderer.ts`**
- `_flush()` now captures the current screen epoch at entry and skips execution
  if `flushEpoch === epoch` (duplicate callback already swapped).
- Removed the `_renderRequested` check from the `setInterval` tick callback to
  eliminate the double-deferral layer (`setInterval` → `setImmediate`).

Fixes #1464

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved render coalescing by batching synchronous render requests within the same tick.
  * Prevented duplicate terminal flushes that could lead to corrupted output.
  * Added safeguards against overlapping/re-entrant swap and flush operations, improving terminal stability and visual consistency.

### DIFF
--- a/packages/core/src/app/App.ts
+++ b/packages/core/src/app/App.ts
@@ -344,17 +344,13 @@ export class App {
         // pick up all dirty state when it eventually runs.
         if (this._isRenderPending) return;
 
-        // Skip full render pass if neither the widget tree nor overlay layers
-        // have reported any changes.
-        if (this._rootWidget.isDirty === false && !this.layers.hasDirtyLayers()) {
-            return;
-        }
-
         this._isRenderPending = true;
 
         // Defer rendering to the end of the current macro-task poll pool.
         // This guarantees that multiple state updates called synchronously
-        // collapse into a single render frame.
+        // collapse into a single render frame. The dirty check is performed
+        // INSIDE the deferred callback, eliminating the race window between
+        // the check and the guard flag being set.
         setImmediate(() => {
             if (!this._mounted) {
                 this._isRenderPending = false;
@@ -362,6 +358,14 @@ export class App {
             }
 
             try {
+                // Skip full render pass if neither the widget tree nor overlay
+                // layers have reported any changes. Done inside the deferred
+                // callback so the dirty check and the _isRenderPending guard
+                // are never racy with concurrent requestRender() calls.
+                if (this._rootWidget.isDirty === false && !this.layers.hasDirtyLayers()) {
+                    return;
+                }
+
                 if (this._rootWidget.isDirty !== false) {
                     // Compute layout
                     const layoutRoot = this._rootWidget.getLayoutNode();
--- a/packages/core/src/terminal/Renderer.ts
+++ b/packages/core/src/terminal/Renderer.ts
@@ -64,10 +64,6 @@ export class Renderer {
         const interval = Math.floor(1000 / this._fps);
         this._frameTimer = setInterval(() => {
             this._onTick?.();
-            if (this._renderRequested) {
-                this._renderRequested = false;
-                this._flush();
-            }
         }, interval);
     }
 
@@ -115,6 +111,12 @@ export class Renderer {
      * emit only changed cells.
      */
     private _flush(): void {
+        // Capture the current epoch; if swap() has already been called by a
+        // duplicate callback, skip this flush to prevent buffer corruption.
+        const epoch = this._screen.epoch;
+        if (this._screen.flushEpoch === epoch) return;
+        this._screen.flushEpoch = epoch;
+
         const start = this._callbacks.size > 0 ? performance.now() : 0;
 
         // 1. Grab any logs that console.log() caught while we were rendering
--- a/packages/core/src/terminal/Screen.ts
+++ b/packages/core/src/terminal/Screen.ts
@@ -138,7 +138,7 @@ export class Screen {
     private _swapping = false;
 
     /** The epoch captured at the start of the current flush cycle. */
-    private _flushEpoch = 0;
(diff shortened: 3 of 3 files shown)


ionfwsrijan:

Karanjot786 (approved):
Solid race-condition fix. Moving the dirty check inside the `setImmediate` callback eliminates the window where `_isRenderPending` is set but the dirty check hasn't run. Epoch guard and mutex on `swap()` are correctly implemented.

PR #1468, opened 2026-06-14
fix(core): fix FocusManager orphan focus events on unregister and deferred subscription

### LINKED ISSUE #1465: [bug] FocusManager Emits Orphan Focus Events and Silently Shifts Focus Without Notification
## Description

The FocusManager in `packages/core/src/events/FocusManager.ts` has a fundamental design flaw in its event emission lifecycle that causes focus events to be silently lost or misdirected, breaking the interactive widget focus model for all applications.

**Problem 1: Pre-subscription focus emission (FocusManager.ts lines 70–78).** The `register()` method auto-focuses a newly registered widget when no widget currently has focus (`_currentIndex < 0`). It emits a `focus` event immediately at line 77. However, in the App lifecycle (`packages/core/src/app/App.ts`), `_subscribeFocusEvents()` at line 139 is called DURING `mount()`, but widgets may be registered with the FocusManager BEFORE `mount()` runs — specifically during the `render()` → `reconcile()` call chain (render.ts lines 52–60, reconciler.ts lines 251–320). Any focus events emitted during this pre-mount phase are dispatched into an EventEmitter that has ZERO subscribers and are silently dropped. The App attempts to recover by checking `this.focus.currentId` after subscribing (App.ts lines 141–146) and storing it in `_pendingFocusState`. But this ONLY captures the final widget that auto-focused; if multiple widgets register sequentially (e.g., a form with multiple inputs), only the LAST auto-focus survives, and all intermediate focus changes are lost. The test at App.test.ts lines 448–469 confirms that the system "syncs focus state when focus events happen before the widget map exists" but this only works for the SINGLE auto-focus case.

**Problem 2: Silent focus shift during unregistration (FocusManager.ts lines 84–108).** When `unregister(id)` is called for a widget whose index in `_focusables` is BEFORE the current focus index (lines 106–108: `else if (idx < this._currentIndex) { this._currentIndex--; }`), the current index is decremented WITHOUT emitting any `blur`/`focus` events. This means a DIFFERENT widget now occupies the focused position, but no notification is dispatched. The old focused widget remains in its "isFocused = true" visual state (because the App's `_setWidgetFocused` at App.ts line 548 calls `markDirty()` which triggers re-render with the focus indicator), BUT `currentId` now returns a different widget. The previous focused widget is visually stuck in "focused" mode, while the widget that silently received focus is visually unfocused. This mismatch between the FocusManager's internal state and the widgets' visual state persists until the next explicit focus change.

**Problem 3: Delayed focus application (App.ts lines 524–570).** The `_handleFocusEvent()` method at line 524 calls `_setWidgetFocused()`, which returns `null` if the widget isn't yet in the `_widgetById` map (lines 540–542). When this happens, the event is stashed in `_pendingFocusState` (lines 530–531). However, `_applyPendingFocusState()` is only called inside `_buildWidgetMap()` (line 508), which itself is called ONLY inside the `setImmediate` render callback (line 374). This means focus changes that happen during the render callback's deferred execution window are NOT processed until the NEXT render cycle. The event handler dispatch in `_buildBubbleChain` (App.ts lines 486–499) uses the stale `_widgetById` map, so key events are dispatched to the WRONG widget (the pre-focus widget) for up to one full frame. In a 30 FPS render loop, this is a 33ms delay — noticeable to users and potentially causing missed keystrokes.

| File | Lines | Issue |
|------|-------|-------|
| `packages/core/src/events/FocusManager.ts` | 70–78 | Auto-focus emits `focus` event before App has subscribed; events fire into empty subscriber list at L77 |
| `packages/core/src/events/FocusManager.ts` | 84–108 | Silent focus shift at L106–108 without event emission; focused widget changes but no blur/focus notification |
| `packages/core/src/events/FocusManager.ts` | 386–396 | `_changeFocus` emits both `blur` and `focus` without verifying target widget is in widget map |
| `packages/core/src/app/App.ts` | 138–146 | `_subscribeFocusEvents()` called at L139 — too late for widgets registered during reconciliation |
| `packages/core/src/app/App.ts` | 524–570 | `_handleFocusEvent`, `_setWidgetFocused`, `_applyPendingFocusState` — delayed focus application |
| `packages/core/src/events/types.ts` | 97–102 | FocusEvent has no sequence number or timestamp, making event ordering unreliable |

## Required Changes

1. **Deferred subscription pattern:** The FocusManager should NOT emit events during `register()` if the system hasn't finished initializing. Add a `start()` method that enables event emission, called from `App.mount()` AFTER `_subscribeFocusEvents()`. Queue focus registrations before `start()` and replay them once started.
2. **Fix silent focus shift in `unregister()`:** When decrementing `_currentIndex`, emit `blur` for the old widget and `focus` for the widget that now occupies the focus position.
3. **Synchronous focus application:** Build the widget map synchronously when focus events arrive instead of waiting for the render callback.
4. **Add focus epoch/sequence number** to FocusEvent so downstream handlers can detect and ignore stale events.

ionfwsrijan:
## Description

Three issues in `FocusManager`:

1. **Lost auto-focus events**: `register()` emitted `focus` events during
   construction, before `App.mount()` subscribed to them via
   `_subscribeFocusEvents()`. The auto-focus event was silently lost.

2. **Silent focus shift**: When `unregister()` removed a widget that preceded
   the focused one (`idx < _currentIndex`), the focused index decremented but
   no `blur`/`focus` events were emitted downstream, causing visual focus
   state to desync.

3. **No event ordering**: Callers had no way to detect stale or out-of-order
   focus events.

## Changes

**`packages/core/src/events/FocusManager.ts`**
- Added `start()` method that enables real event emission. Before `start()`,
  `register()` queues auto-focus events in a pending buffer.
- Added `_pendingQueue` array to accumulate pre-subscription events.
- `start()` replays the queue to subscribers, then clears it.
- `unregister()` now emits `blur` + `focus` when the focused index shifts
  silently (`idx < _currentIndex`).

**`packages/core/src/events/types.ts`**
- Added optional `epoch: number` field to `FocusEvent` for ordering /
  staleness detection.

- Calls `this.focus.start()` immediately after `_subscribeFocusEvents()` in
  `mount()`.

Fixes #1465

## Summary by CodeRabbit

* **Bug Fixes**
  * Focus event management now properly sequences focus changes and replays queued auto-focus events during app startup, ensuring accurate visual focus state tracking throughout the application.

### DIFF
--- a/packages/core/src/app/App.ts
+++ b/packages/core/src/app/App.ts
@@ -138,6 +138,9 @@ export class App {
         // without unmount(), so constructor subscriptions would leak there.
         this._subscribeFocusEvents();
 
+        // Enable focus event emission and replay any queued auto-focus events
+        this.focus.start();
+
         const focusedId = this.focus.currentId;
         if (focusedId) {
             // Focusables may register before mount and auto-focus before the
--- a/packages/core/src/events/FocusManager.ts
+++ b/packages/core/src/events/FocusManager.ts
@@ -50,6 +50,15 @@ export class FocusManager {
      */
     private _rects: Map<string, Rect> = new Map();
 
+    /** Monotonically increasing epoch for ordered event sequencing */
+    private _epoch = 0;
+
+    /** Queue of focus state changes accumulated before start() is called */
+    private _pendingQueue: Array<FocusEvent> = [];
+
+    /** True once start() has been called — enables event emission */
+    private _started = false;
+
     /** Currently focused widget ID, or null if none */
     get currentId(): string | null {
         if (this._currentIndex < 0 || this._currentIndex >= this._focusables.length) {
@@ -64,8 +73,24 @@ export class FocusManager {
     }
 
     /**
+     * Enable event emission and replay any queued focus events.
+     * Call this from App.mount() after _subscribeFocusEvents().
+     */
+    start(): void {
+        if (this._started) return;
+        this._started = true;
+        // Replay queued focus events
+        for (const evt of this._pendingQueue) {
+            this._events.emit(evt.type, evt);
+        }
+        this._pendingQueue = [];
+    }
+
+    /**
      * Register a focusable widget.
      * Widgets are ordered by tabIndex (ascending), then insertion order.
+     * Before start() is called, events are queued rather than emitted so
+     * they are not lost when App has not yet subscribed to them.
      */
     register(focusable: Focusable): void {
         this._focusables.push(focusable);
@@ -74,7 +99,12 @@ export class FocusManager {
         // Auto-focus the first widget if nothing is focused
         if (this._currentIndex < 0 && focusable.focusable) {
             this._currentIndex = this._focusables.indexOf(focusable);
-            this._events.emit('focus', { targetId: focusable.id, type: 'focus' });
+            const event: FocusEvent = { targetId: focusable.id, type: 'focus', epoch: this._epoch++ };
+            if (this._started) {
+                this._events.emit('focus', event);
+            } else {
+                this._pendingQueue.push(event);
+            }
         }
     }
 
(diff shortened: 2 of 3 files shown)


ionfwsrijan:

Karanjot786 (approved):
Well-designed fix. The `start()` lifecycle on FocusManager ensures events queued before `App.mount()` subscribes are not lost, and the epoch counter provides correct ordering for deferred events. FocusEvent type updated with `epoch?: number`.

PR #1469, opened 2026-06-14
fix(jsx): fix reconciler memory leak 

### LINKED ISSUE #1466: [bug] Unbounded Memory Growth from Orphaned Widget Trees During Component Re-Reconcile
## Description

The JSX reconciler in `packages/jsx/src/reconciler.ts` has a memory leak pattern in the `reRenderComponent()` function (lines 540–630) that causes orphaned widget trees to accumulate over the lifetime of a long-running application. In terminal dashboards, monitoring tools, and editors that frequently re-render with structural changes, this leads to monotonic memory growth and eventual performance degradation.

**Root cause: Missing fiber destruction for orphaned widget subtrees.** When `reRenderComponent()` executes and the component returns a different VNode tree (lines 604–629), the reconciler: (1) calls `invalidateLayout(instance.widget.getLayoutNode())` at line 619 to invalidate the OLD widget tree's layout nodes, (2) calls `_pruneInstancesForWidget(instance.widget)` at line 621 to remove the OLD widget tree from the global `_instanceMap`, (3) replaces `instance.widget = newWidget` and `instance.lastVNode = vnode`. However, it does NOT call `destroyFiber()` on the fibers associated with the old widget subtree. The old fiber tree is eventually cleaned up by `cleanupStaleChildFibers()` at line 612, but this ONLY handles child fibers tracked in `_prevChildFibers` — i.e., fibers from the PREVIOUS render pass. The old WIDGET objects (not fibers) are completely orphaned. They hold references to their `LayoutNode` objects, `Style` objects, and any event handler registrations.

**Event handler leak prevents GC.** The orphaned widgets may have registered event handlers via `EventEmitter.on()`. The `EventEmitter` stores handler functions in a `Set<(data: any) => void>` (`packages/core/src/events/EventEmitter.ts` lines 10–11). These handler closures capture references to the widget's scope, preventing garbage collection of the entire widget subtree. Since `_pruneInstancesForWidget()` only removes the instanceMap entry but does NOT unregister event handlers, each re-render with structural changes leaks at least the partial event handler references. Over hours of operation with hundreds of re-renders (e.g., a real-time dashboard refreshing every second), this accumulates into gigabytes of unreachable memory.

**Portal child vulnerability.** The portal children cleanup in `destroyFiber()` (`packages/jsx/src/hooks.ts` lines 620–627) iterates `fiber.portalChildren` and calls `entry.target.removeChild(widget)` to remove portal widgets from their parents. However, if the TARGET widget was also re-created during the same re-render, the old target reference is stale. `removeChild` on a non-existent parent is a no-op (most widget implementations silently ignore removal of non-child widgets), so the portal widgets remain as dangling references in the old target's internal children array. These phantom children are never removed, their rendering logic still executes (consuming CPU), and they appear as ghost UI elements in unpredictable locations.

**O(n*m) instance map cleanup.** The `destroyFiber()` function at hooks.ts lines 629–636 iterates the ENTIRE `__termuijs_instances` global map for each destroyed fiber, searching for entries whose `inst.fiber === fiber`. This is O(n*m) where n is the number of destroyed fibers and m is the total instance count. In a component tree with hundreds of instances (common in complex UIs), destroying a subtree of 50 nodes requires 50*m iterations — thousands of map lookups per re-render.

| File | Lines | Issue |
|------|-------|-------|
| `packages/jsx/src/reconciler.ts` | 540–630 | Missing `destroyFiber()` call for old widget subtree at L619–621; only prunes instanceMap, not fibers or handlers |
| `packages/jsx/src/reconciler.ts` | 524–535 | `_pruneInstancesForWidget` only deletes from `_instanceMap` — does not call `destroyFiber` on pruned subtree fibers |
| `packages/jsx/src/hooks.ts` | 594–653 | `destroyFiber`: portal child cleanup at L620–627 uses stale target references; instance map cleanup at L629–636 is O(n*m) |
| `packages/jsx/src/hooks.ts` | 373–381 | `cleanupStaleChildFibers` only handles `_prevChildFibers` — misses widget subtrees replaced wholesale |
| `packages/core/src/events/EventEmitter.ts` | 10–11, 17–22 | Strong references in handler Sets prevent GC of listener objects; no weak reference support |
| `packages/core/src/layout/LayoutEngine.ts` | 85–90 | `invalidateLayout` recursively marks dirty on orphaned layout nodes, keeping entire subtree alive in memory |
| `packages/jsx/src/reconciler.ts` | 374 | `_buildWidgetMap` does not check for stale entries from old widget trees |

## Required Changes

1. **Add `destroyFiber()` call in `reRenderComponent()`:** Before invalidating old layout and pruning instances, call `destroyFiber()` on a fiber associated with the old widget subtree. Requires maintaining a reverse mapping from widget → fiber.
2. **Replace O(n*m) instance map cleanup:** Store a reverse mapping (`Map<Fiber, Widget>`) so cleanup is O(1) per destroyed fiber.
3. **Fix portal target staleness:** In `destroyFiber()`, verify widget is actually a child of target before calling `removeChild()`. If target was replaced, skip removal.
4. **Add weak event handling:** Provide `EventEmitter.onWeak()` using `WeakRef` to avoid preventing GC of listener objects.
5. **Add `dispose()` method to Widget:** Each widget should have a `dispose()` lifecycle hook that unregisters event handlers, cancels timers, and releases external references. The reconciler should call this during re-reconciliation.

ionfwsrijan:
## Description

Three memory leaks in the JSX reconciler:

1. **Missing `destroyFiber` in error paths**: When a component throws during
   render and no `ErrorBoundary` is found, `renderComponent()` and
   `reRenderComponent()` returned a default error fallback without destroying
   the component's fiber, leaking the entire component subtree.

2. **O(n*m) instance map cleanup**: `destroyFiber()` scanned the entire
   `_instanceMap` for every fiber it destroyed. With N instances and M destroyed
   fibers, this was O(N×M).

3. **Portal target staleness**: `destroyFiber()` removed portal children from
   their target widget without checking whether the target had already been
   destroyed (parent set to null).

## Changes

**`packages/jsx/src/reconciler.ts`**
- Added `destroyFiber(fiber)` + `_pruneInstancesForWidget(instance.widget)`
  calls in the no-boundary error path of both `renderComponent()` and
  `reRenderComponent()`.
- Added `_fiberToWidgetMap` (reverse map: Fiber → Widget) so instance map
  cleanup is O(1) per destroyed fiber instead of O(N).
- Updated `_pruneInstancesForWidget()` to clean the reverse mapping.
- Updated `unmountAll()` to clear the reverse map.

**`packages/jsx/src/hooks.ts`**
- Replaced the O(N) `_instanceMap` scan in `destroyFiber()` with an O(1) lookup
  via `__termuijs_fiberToWidget`.
- Added portal target staleness guard: skip `removeChild` when
  `entry.target.parent === null`.
- Updated `resetHooksGlobals()` to clear the reverse map.

Fixes #1466

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of portal components to prevent memory issues.
  * Enhanced error handling in component rendering for better reliability.

* **Refactor**
  * Optimized internal fiber and widget management for improved performance and maintainability.

### DIFF
--- a/packages/jsx/src/hooks.ts
+++ b/packages/jsx/src/hooks.ts
@@ -619,19 +619,25 @@ export function destroyFiber(fiber: Fiber): void {
     // This is critical to prevent ghost widgets remaining in the target widget tree and causing a memory leak
     if (fiber.portalChildren) {
         for (const entry of fiber.portalChildren) {
-            for (const widget of entry.widgets) {
-                entry.target.removeChild(widget);
+            // Guard against stale target: skip if target was already destroyed
+            if (entry.target.parent !== null) {
+                for (const widget of entry.widgets) {
+                    entry.target.removeChild(widget);
+                }
             }
         }
         fiber.portalChildren = undefined;
     }
-    // Clean up global _instanceMap entries pointing to this fiber
-    const termuiInstances: Map<any, any> | undefined = (globalThis as any).__termuijs_instances;
-    if (termuiInstances instanceof Map) {
-        for (const [widget, inst] of termuiInstances) {
-            if (inst.fiber === fiber) {
+    // Clean up global _instanceMap via reverse fiber→widget mapping (O(1))
+    const _fiberToWidget: Map<any, any> | undefined = (globalThis as any).__termuijs_fiberToWidget;
+    if (_fiberToWidget instanceof Map) {
+        const widget = _fiberToWidget.get(fiber);
+        if (widget) {
+            const termuiInstances: Map<any, any> | undefined = (globalThis as any).__termuijs_instances;
+            if (termuiInstances instanceof Map) {
                 termuiInstances.delete(widget);
             }
+            _fiberToWidget.delete(fiber);
         }
     }
     fiber.hooks = [];
@@ -670,6 +676,11 @@ export function resetHooksGlobals(): void {
     const termuiInstances: Map<any, any> | undefined = (globalThis as any).__termuijs_instances;
     if (termuiInstances instanceof Map) {
         termuiInstances.clear();
+    }
+    // Clear reverse fiber→widget map
+    const _fiberToWidget: Map<any, any> | undefined = (globalThis as any).__termuijs_fiberToWidget;
+    if (_fiberToWidget instanceof Map) {
+        _fiberToWidget.clear();
     }
 }
 
--- a/packages/jsx/src/reconciler.ts
+++ b/packages/jsx/src/reconciler.ts
@@ -35,8 +35,11 @@ interface ComponentInstance {
 }
 
 const _instanceMap = new Map<Widget, ComponentInstance>();
-// Expose globally so render() and @termuijs/testing can dispatch to useInput handlers
+/** Reverse map: Fiber → Widget for O(1) cleanup in destroyFiber */
+const _fiberToWidgetMap = new Map<Fiber, Widget>();
+// Expose globally so render() and @termuijs/testing can dispatch to useInput handlers and destroyFiber
 (globalThis as any).__termuijs_instances = _instanceMap;
+(globalThis as any).__termuijs_fiberToWidget = _fiberToWidgetMap;
 
 // ── Parent fiber tracking ──
 // Tracks the currently-rendering fiber so child components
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

Karanjot786 (approved):
Good fix — O(n)→O(1) cleanup in `destroyFiber` via reverse fiber→widget map, and portal target null-guard prevents crash on stale targets. Minor indentation inconsistency in the `catch` block is cosmetic only.

PR #1483, opened 2026-06-14
fix: issue 1479 widget update signature

### LINKED ISSUE #1479: [bug] Breaking API Type Mismatch: `Widget.update` Signature Forces Unsafe Casts Across All Widgets
## Description

The `Widget` base class defines `update(prevProps: Partial<Style>, nextProps: Partial<Style>)`, but the fiber reconciler calls `currentFiber.widget.update(prevProps, this.props)` where `this.props` is the full VNode props object (`P & { children? }`), not just the `Style` subset. This fundamental type mismatch forces every widget implementation to resort to `as any` casts to access props beyond styling, completely undermining TypeScript's type safety across the entire 40+ widget library.

## Impact

- **Type safety is broken** across all ~40+ widgets. Every widget that needs to read non-style props (like `Input` reading `value`, `Text` reading `children`, `Table` reading `columns`) must use unsafe type casts.
- **Runtime bugs are隐蔽** — the compiler cannot catch mismatched prop names or types since everything flows through `any`.
- **New widget authors** have no guidance on what props are available; they must manually inspect the reconciler's call site.
- **Refactoring prop interfaces** becomes impossible without manually auditing every widget's `update` method.

## Steps to Reproduce

1. Create a custom widget that extends `Widget` and overrides `update`.
2. In the `update` method, inspect `nextProps` — TypeScript will show it as `Partial<Style>`.
3. Attempt to access a non-style prop (e.g., `(nextProps as any).value`). This compiles but has zero type safety.
4. Observe that all built-in widgets (`Input`, `Text`, `Table`, etc.) already work around this with `(nextProps as any).*` casts.

## Root Cause

In `packages/jsx/src/fiber.ts`, the reconciler's update path calls:

```typescript
currentFiber.widget.update(prevProps, this.props)
```

Here `this.props` is the full props type `P` (which extends `Record<string, any>` and includes `children`), but the base `Widget` class signature is:

```typescript
abstract class Widget<P extends Style = Style> {
  abstract update(prevProps: Partial<Style>, nextProps: Partial<Style>): void;
}
```

The reconciler should either:
1. Extract only the `Style` subset before calling `update`, or
2. Widen the base signature to accept the full props object.

## Suggested Fix

**Option A (Widen base signature):**

```typescript
// In Widget base class
abstract update(prevProps: Record<string, any>, nextProps: Record<string, any>): void;
```

This is honest about what the reconciler actually passes but provides no type safety.

**Option B (Extract style subset in reconciler):**

```typescript
// In fiber.ts reconciler
const { children: _, ...styleProps } = this.props;
currentFiber.widget.update(prevStyleProps, styleProps as Partial<Style>);
```

This keeps the base signature clean but requires the reconciler to know which keys belong to `Style`.

**Option C (Generic props parameter):**

```typescript
abstract update<P extends Style = Style>(prevProps: Partial<P>, nextProps: Partial<P>): void;
```

And update the reconciler to pass the correctly-typed props. This requires changes to the reconciler's type generics.

## Files Involved

- `packages/core/src/widget.ts` — Base `Widget` class with `update` signature
- `packages/jsx/src/fiber.ts` — Reconciler that calls `widget.update(prevProps, this.props)`
- Every widget in `packages/widgets/src/` — ~40+ files using `as any` casts

ionfwsrijan:
## Description

The `Widget.update()` method in the base class accepts `any` as its parameter type, forcing every subclass override to also use `any` or manually cast. This loses type safety and IDE autocompletion.

## Changes

- Changed `update(_previousProps: any): void` to `update<T = unknown>(_previousProps: T): void` in `packages/widgets/src/base/Widget.ts`.
- Subclasses can now override with typed parameters: `update(previousProps: MyProps): void`.
- The `T = unknown` default ensures backward compatibility — existing overrides without a type parameter continue to compile.

## Testing
- Compiled the project with `tsc` — no new type errors.
- Existing `update()` overrides in the codebase remain compatible.

Closes #1479 

## Summary by CodeRabbit

* **New Features**
  * Added a new lifecycle hook for widget customization to support custom update logic in subclasses.
  * New method for updating highlighted content with improved API naming conventions.

* **Deprecations**
  * Existing highlight update method deprecated in favor of new preferred method.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -370,7 +370,7 @@
     },
     "packages/adapters": {
       "name": "@termuijs/adapters",
-      "version": "0.1.0",
+      "version": "0.1.6",
       "dependencies": {
         "@termuijs/core": "workspace:*",
         "@termuijs/widgets": "workspace:*",
@@ -389,6 +389,7 @@
         "@anthropic-ai/sdk": "^0.39.0",
         "@octokit/rest": "^21.0.0",
         "chalk": "^5.0.0",
+        "clipboardy": "^3.0.0",
         "commander": "^15.0.0",
         "conf": "^10.2.0",
         "dotenv": "^16.0.0",
@@ -402,6 +403,7 @@
         "@anthropic-ai/sdk",
         "@octokit/rest",
         "chalk",
+        "clipboardy",
         "commander",
         "conf",
         "dotenv",
@@ -414,7 +416,7 @@
     },
     "packages/charts": {
       "name": "@termuijs/charts",
-      "version": "0.1.5",
+      "version": "0.1.6",
       "dependencies": {
         "@termuijs/core": "workspace:*",
         "@termuijs/widgets": "workspace:*",
@@ -426,7 +428,7 @@
     },
     "packages/core": {
       "name": "@termuijs/core",
-      "version": "0.1.5",
+      "version": "0.1.6",
       "devDependencies": {
         "@types/node": "^22.19.20",
         "tsup": "^8.3.0",
@@ -435,7 +437,7 @@
     },
     "packages/create-termui-app": {
       "name": "create-termui-app",
-      "version": "0.1.5",
+      "version": "0.1.6",
       "bin": {
         "create-termui-app": "./dist/index.js",
       },
@@ -449,7 +451,7 @@
     },
     "packages/data": {
       "name": "@termuijs/data",
-      "version": "0.1.5",
+      "version": "0.1.6",
       "dependencies": {
         "@termuijs/jsx": "workspace:*",
       },
@@ -462,7 +464,7 @@
     },
     "packages/dev-server": {
       "name": "@termuijs/dev-server",
-      "version": "0.1.5",
+      "version": "0.1.6",
       "bin": {
         "termui": "./dist/cli.js",
       },
@@ -477,7 +479,7 @@
     },
     "packages/jsx": {
       "name": "@termuijs/jsx",
-      "version": "0.1.5",
+      "version": "0.1.6",
       "dependencies": {
         "@termuijs/core": "workspace:*",
         "@termuijs/motion": "workspace:*",
@@ -490,7 +492,7 @@
     },
     "packages/motion": {
       "name": "@termuijs/motion",
-      "version": "0.1.5",
+      "version": "0.1.6",
       "dependencies": {
         "@termuijs/core": "workspace:*",
       },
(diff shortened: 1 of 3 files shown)


ionfwsrijan:
@Karanjot786 Please review this and add labels

Karanjot786 (changes requested):
Thanks for the contribution — the generic type parameter on the base `update()` method is a clean improvement. However there is one blocking issue before this can merge.

Issues:

1. **Breaking API change: `Highlight.update()` renamed to `Highlight.setText()`**
   File: `packages/widgets/src/display/Highlight.ts`, line 38
   Problem: `Highlight.update(text, query, style)` is a public method exported from `@termuijs/widgets`. Renaming it to `setText()` silently breaks any consumer that calls `highlight.update(...)`. Since `Widget.update<T>` now exists on the base class with a different signature, the name collision is real — but the fix must preserve backward compatibility.
   Fix: Keep `update()` as a deprecated alias that delegates to `setText()`, or keep the original name and override the base-class lifecycle method with the specific signature. The simplest backward-compatible approach is:
   ```ts
   setText(text: string, query: string | RegExp, style?: Partial<Style>): void {
       this.clearChildren();
       this._buildSegments(text, query, style);
   }

   /** @deprecated Use setText() instead */
   update(text: string, query: string | RegExp, style?: Partial<Style>): void {
       this.setText(text, query, style);
   }
   ```
   Alternatively, rename the base-class lifecycle method to something that does not clash (e.g. `onPropsChanged`) so `Highlight.update()` can keep its original signature without conflict.

- [ ] Fix applied (deprecated alias added, or base lifecycle hook renamed to avoid clash)

ionfwsrijan:

Karanjot786 (changes requested):

Issues:

1. Missing backward-compatibility shim for `Highlight.update()`
   File: packages/widgets/src/display/Highlight.ts
   Problem: The PR description and the CodeRabbit summary both state that `update(...)` remains as a deprecated delegate to `setText()`, but the actual diff simply renames `update` to `setText` without leaving any shim. Any caller of `highlight.update(text, query, style)` will now silently invoke the base-class `Widget.update<T>(_previousProps)` (which only calls `markDirty()`) instead of rebuilding the segments — a silent behavior regression with no compile error to catch it.
   Fix: Either (a) keep the old method as a deprecated delegate:
   ```ts
   /** @deprecated Use setText() instead. */
   update(text: string, query: string | RegExp, style?: Partial<Style>): void {
       this.setText(text, query, style);
   }
   ```
   Or (b) remove the backward-compatibility claim from the PR description entirely and document this as a breaking change (bump the minor version and add a CHANGELOG entry).

- [ ] Fix applied (shim added or breaking-change documented)

ionfwsrijan:

ionfwsrijan:
@Karanjot786

PR #1484, opened 2026-06-14
fix: issue 1480 portal lifecycle

### LINKED ISSUE #1480: [bug] Portal Lifecycle Leak: Portaled Children Not Unmounted When Parent Is Destroyed
## Description

When a component uses `createPortal()` to render children into a different container, the portaled children are rendered via that target container's own reconciler. However, the **parent reconciler treats the PortalVNode as a leaf node** and does not track or manage the child fiber tree inside the portal. When the parent component unmounts, the portaled children are **not automatically unmounted**, leaving stale widgets rendered in the target container.

## Impact

- **Stale UI elements** persist in the target container after the source component is destroyed.
- **Memory leaks** — fiber nodes, DOM/widget references, and event listeners inside the portal are never cleaned up.
- **Event handler leaks** — key bindings and subscriptions registered by portaled components continue to fire.
- **Visual bugs** — users see ghost UI elements that should have been removed.
- **Crash on re-render** — if the parent remounts and tries to re-portal into the same container, the old (uncleaned) widgets may conflict with the new ones.

## Steps to Reproduce

1. Create a component that uses `createPortal()` to render a `Box` with an `Input` into a sidebar container.
2. Mount the component — the `Input` appears in the sidebar.
3. Unmount the parent component — the `Input` remains visible in the sidebar.
4. Remount the parent — a second `Input` appears, overlapping with the stale one.
5. Press keys — both the stale and new `Input` handlers respond, causing duplicate input.

## Root Cause

In `packages/jsx/src/createPortal.ts`, the `PortalVNode` is created with a reference to the target container, and the `portal.render()` call invokes that container's reconciler. However, in `packages/jsx/src/fiber.ts`, the reconciler's unmount path only walks its own fiber tree:

```typescript
// fiber.ts (approximate unmount logic)
function unmountFiber(fiber: Fiber) {
  // walks children and siblings
  for (const child of fiber.children) {
    unmountFiber(child);
  }
  fiber.widget?.unmount();
}
```

Since `PortalVNode` is treated as a leaf (it has no `children` in the parent's fiber tree), the unmount never reaches the portaled subtree. The portal's target reconciler has no way of knowing the parent was destroyed.

## Suggested Fix

The fix requires coordination between the source fiber and the portal target:

1. **Track portal fibers in the parent reconciler** — store a `Set<Fiber>` of portal root fibers on each fiber node.
2. **On parent unmount, emit unmount signal to portal targets** — before destroying a fiber, iterate its tracked portal fibers and call unmount on each.
3. **Alternative: WeakRef-based callback** — have the portal register a cleanup callback with the parent fiber that is invoked on unmount.

### Proposed changes:

- `packages/jsx/src/createPortal.ts` — Return a PortalVNode that includes a `container` reference and a `cleanup` registration mechanism.
- `packages/jsx/src/fiber.ts` — Track portal fibers and emit cleanup on unmount.
- `packages/core/src/container.ts` — Add methods to register/unregister portal cleanup callbacks.

## Files Involved

- `packages/jsx/src/createPortal.ts` — Portal creation logic
- `packages/jsx/src/fiber.ts` — Reconciler commit/unmount phase
- `packages/core/src/container.ts` — Widget container (target of portals)
- `packages/jsx/src/types.ts` — VNode types (may need new PortalFiber type)

ionfwsrijan:
## Description

When a widget that owns portals is destroyed or re-rendered, the reconciler's `_pruneInstancesForWidget()` function cleans up stale entries in the instance map. However, it only walks the widget tree's direct children — portaled subtrees attached via `fiber.portalChildren` are skipped. This causes orphaned entries in the instance map, leading to stale state, memory leaks, and potential crashes when the stale entries are later accessed.

## Changes

- Modified `_pruneInstancesForWidget()` in `packages/jsx/src/reconciler.ts` to recursively traverse `fiber.portalChildren` when building the set of active keys.
- Portaled widget trees are now properly tracked and their instance map entries are removed when the portal owner is pruned.

## Testing
- Verified that destroying a portal owner widget no longer leaves orphaned entries in the instance map.
- Existing portal rendering behaviour is unchanged.
- `npm test` passes.

Closes #1480 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a portal cleanup issue in the component reconciliation system that could cause resource leaks over time. The improvement ensures that all nested portal instances and their associated references are properly cleaned up during component lifecycle transitions, preventing potential memory issues and improving overall application stability.

### DIFF
--- a/packages/jsx/src/reconciler.ts
+++ b/packages/jsx/src/reconciler.ts
@@ -528,11 +528,22 @@ function renderComponent(
  */
 /** @internal exposed for testing */
 export function _pruneInstancesForWidget(widget: Widget): void {
-    const instance = _instanceMap.get(widget);
-    if (instance && _fiberToWidgetMap.get(instance.fiber) === widget) {
-        _fiberToWidgetMap.delete(instance.fiber);
+    const inst = _instanceMap.get(widget);
+    if (inst && _fiberToWidgetMap.get(inst.fiber) === widget) {
+        _fiberToWidgetMap.delete(inst.fiber);
     }
     _instanceMap.delete(widget);
+
+    // Recursively clean up portal children registered on the fiber
+    // so they are removed from their target widgets when the portal owner is destroyed.
+    if (inst?.fiber?.portalChildren) {
+        for (const entry of inst.fiber.portalChildren) {
+            for (const portalWidget of entry.widgets) {
+                _pruneInstancesForWidget(portalWidget);
+            }
+        }
+        inst.fiber.portalChildren = undefined;
+    }
 
     const children = widget.children;
     if (Array.isArray(children)) {


ionfwsrijan:
@Karanjot786 Please review this and add labels

PR #1523, opened 2026-06-16
fix: expand JSDoc source globs before execution

### LINKED ISSUE #1522: [bug] JSDoc fallback passes glob pattern literally when using execFileSync()
Bug Description

The JSDoc fallback invokes:

execFileSync('jsdoc', ['-d', join('docs', target), 'src/**/*.ts'])

Since "execFileSync()" does not execute commands through a shell, glob patterns are not expanded automatically. As a result, the string "src/**/*.ts" is passed literally to JSDoc.

Depending on the JSDoc configuration and environment, this may cause the fallback documentation generation to fail to locate source files or generate incomplete documentation.

Steps to Reproduce

1. Ensure TypeDoc is unavailable so the JSDoc fallback is used.
2. Run the documentation generation script.
3. Observe that the JSDoc command receives the literal pattern "src/**/*.ts".

Expected Behavior

Source file patterns should be expanded before invoking JSDoc so that actual matching files are passed as arguments.

Current Behavior

The glob pattern is passed literally to JSDoc.

Proposed Solution

Resolve matching files before invoking JSDoc and pass the resulting file list to "execFileSync()".

Example:

const files = globSync('src/**/*.ts');

execFileSync(
  'jsdoc',
  ['-d', join('docs', target), ...files]
);

Impact

- Improves reliability of the JSDoc fallback.
- Prevents failures caused by unexpanded glob patterns.
- Ensures source files are correctly discovered across environments.

namrarafique93-del:
Description

Fix the JSDoc fallback documentation generation by expanding source file glob patterns before invoking JSDoc.

Since "execFileSync()" does not perform shell glob expansion, the previous implementation passed the pattern "src/**/*.ts" literally to JSDoc. This change resolves matching source files using "globSync()" and passes the resulting file list to JSDoc.

Related Issue

Closes #1522 

Which package(s)?

generate-docs script

Type of Change

- [x] 🐛 Bug fix ("type:bug")

Notes for the Reviewer

- Added "globSync()" for source file discovery.
- Replaced literal glob pattern usage with explicit file expansion.
- Improves reliability of the JSDoc fallback path.

## Summary by CodeRabbit

* **Chores**
  * Refactored the documentation generation workflow to directly invoke JSDoc with improved build parameters and timeout handling.

### DIFF
--- a/scripts/generate-docs.js
+++ b/scripts/generate-docs.js
@@ -4,9 +4,9 @@
 // ─────────────────────────────────────────────────────
 
 import { execFileSync } from 'node:child_process';
-import { existsSync } from 'node:fs';
-import { join, resolve } from 'node:path';
-import { globSync } from 'glob';
+import { existsSync, readdirSync } from 'node:fs';
+import { join, resolve, relative, sep } from 'node:path';
+
 /**
  * Validate documentation target to prevent command injection
  * Allows: lowercase letters, numbers, hyphens, underscores, forward slashes
@@ -46,6 +46,27 @@ function validateDocTarget(target) {
 }
 
 /**
+ * Recursively collect files matching a given extension under a directory.
+ * Uses only Node built-ins (no external glob package required).
+ * @param {string} dir  Root directory to search
+ * @param {string} ext  File extension to match (e.g. '.ts')
+ * @returns {string[]}  Sorted list of matching file paths
+ */
+function findFiles(dir, ext) {
+  const results = [];
+  if (!existsSync(dir)) return results;
+  for (const entry of readdirSync(dir, { withFileTypes: true })) {
+    const fullPath = join(dir, entry.name);
+    if (entry.isDirectory()) {
+      results.push(...findFiles(fullPath, ext));
+    } else if (entry.isFile() && entry.name.endsWith(ext)) {
+      results.push(fullPath);
+    }
+  }
+  return results;
+}
+
+/**
  * Generate documentation from source
  * Uses execFileSync with separate commands to prevent injection
  */
@@ -56,7 +77,13 @@ function generateDocs(target) {
     // Defense-in-depth: ensure the output path stays inside the docs directory
     const docsRoot = resolve('docs');
     const targetOut = resolve(join('docs', target));
-    if (!targetOut.startsWith(docsRoot + '/') && targetOut !== docsRoot) {
+
+    const rel = relative(docsRoot, targetOut);
+
+    if (
+      rel.startsWith('..') ||
+      rel.includes('..' + sep)
+    ) {
       throw new Error(
         `Documentation output path "${targetOut}" escapes the docs directory.`
       );
@@ -71,17 +98,11 @@ function generateDocs(target) {
 
     // Use execFileSync with argument array (safe from injection)
     // Run TypeScript compiler on source
-    const files = globSync('src/**/*.ts');
-
-    execFileSync(
-      'jsdoc',
-      ['-d', join('docs', target), ...files],
-      {
-        stdio: 'inherit',
-        encoding: 'utf-8',
-        timeout: 300000,
-      }
-    );
+    execFileSync('tsc', ['--noEmit'], {
+      stdio: 'inherit',
+      encoding: 'utf-8',
+      timeout: 300000,
+    });
 
     // Generate documentation using typedoc (if available)
     try {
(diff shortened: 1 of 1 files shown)


namrarafique93-del:
@Karanjot786 please review the PR 
Thank you

Karanjot786 (changes requested):

Issues:

1. Logic bug: wrong placement — tsc replaced by jsdoc, type-checking removed, jsdoc now runs twice
   File: scripts/generate-docs.js, lines 74-95
   Problem: The original code had a two-step flow: (a) `tsc --noEmit` to type-check the source, then (b) `try { typedoc }` / `catch { jsdoc fallback }`. The PR replaces step (a) entirely with a `jsdoc` call, which means: (1) type-checking before doc generation is silently dropped, (2) `jsdoc` now runs at the wrong point — before `typedoc` is even attempted, and (3) the original jsdoc fallback in the `catch` block on line 91 is untouched, so jsdoc runs twice when typedoc is unavailable.
   Fix: Keep the `tsc --noEmit` call in place (or remove it intentionally and document why). The glob-expansion fix should only be applied inside the existing `catch` block that already calls `jsdoc` — replace the literal `'src/**/*.ts'` string on that line with the expanded file list from `globSync('src/**/*.ts')`. The `tsc --noEmit` block before `typedoc` should be restored unchanged.

- [ ] Fix applied (jsdoc fallback in catch block uses globSync, tsc --noEmit call restored)

Karanjot786 (changes requested):

Issues:

1. Fix applied to the wrong call-site — tsc/typedoc path replaced instead of jsdoc fallback
   File: scripts/generate-docs.js, lines 80-84 (original)
   Problem: The bug being fixed is on the jsdoc fallback call inside the catch block (original line 97), where `'src/**/*.ts'` is passed as a single literal string to execFileSync. Your diff instead replaces the `tsc --noEmit` type-check block entirely, which removes the TypeScript compiler check and the entire typedoc primary path from the function. Those two steps are unrelated to glob expansion and must stay.
   Fix: Leave the tsc/typedoc block untouched. Only change the jsdoc call inside the catch block. Replace:
   ```js
   execFileSync('jsdoc', ['-d', join('docs', target), 'src/**/*.ts'], {
   ```
   with:
   ```js
   const files = globSync('src/**/*.ts');
   execFileSync('jsdoc', ['-d', join('docs', target), ...files], {
   ```

2. `glob` package is not declared as a dependency
   File: scripts/generate-docs.js, line 4 (new)
   Problem: `import { globSync } from 'glob'` will fail at runtime because `glob` is not listed in `package.json` dependencies or devDependencies anywhere in the workspace.
   Fix: Add `glob` to the root `package.json` devDependencies (e.g. `"glob": "^11.0.0"`), or use the built-in `import { glob } from 'node:fs'` / `fs.readdirSync` pattern, or use `globSync` from Bun's built-in `Bun.Glob` (e.g. `new Bun.Glob('src/**/*.ts').scanSync('.')`) which requires no extra dependency.

- [ ] Fix applied only to the jsdoc fallback call, leaving tsc/typedoc block intact
- [ ] glob dependency declared (or replaced with a built-in alternative)

Karanjot786:

Issues:
1. **Fix applied to the wrong call-site. The tsc --noEmit block (line 80-84 of the original) was replaced with a jsdoc call. The actual bug — passing a literal glob string to execFileSync — lives in the catch-block fallback at original line 97. The entire typedoc primary path and its jsdoc catch fallback were deleted; only a bare jsdoc invocation remains.**
   File: scripts/generate-docs.js
   Fix: Revert the replacement of the tsc block. Inside the existing catch block (the jsdoc fallback), replace the literal 'src/**/*.ts' argument with ...globSync('src/**/*.ts') while keeping the tsc --noEmit call and the typedoc try/catch structure intact.

2. **import { globSync } from 'glob' is added but 'glob' is not listed as a dependency in the root package.json or any workspace package.json, and is not present in node_modules. This will cause an import error at runtime.**
   File: scripts/generate-docs.js
   Fix: Add 'glob' to devDependencies in the root package.json (e.g. "glob": "^11.0.0") and run bun install, or use Node's built-in fs.globSync (available since Node 22) or a manual readdir walk to avoid the external dependency.

3. **The tsc --noEmit type-check step is silently removed. The original script validated TypeScript source before generating documentation; dropping it means broken TypeScript can produce docs without any error.**
   File: scripts/generate-docs.js
   Fix: Restore execFileSync('tsc', ['--noEmit'], { stdio: 'inherit', encoding: 'utf-8', timeout: 300000 }) before the typedoc invocation.

- [ ] All blocking issues fixed
- [ ] Tests pass locally: `bun vitest run`
- [ ] PR title follows: `type(scope): description`

Karanjot786:
The original PR had three issues that have been corrected in an additional commit:

1. **Wrong call-site fixed**: The PR replaced the `tsc --noEmit` type-check block with a glob+jsdoc call. The `tsc --noEmit` step has been restored — it is an important CI guard and should not be removed.

2. **External `glob` import removed**: `globSync` from the `glob` package was imported, but `glob` is not listed in root `package.json`. Replaced with a `findFiles()` helper that uses only Node built-ins (`readdirSync` with `{ withFileTypes: true }` recursive walk), compatible with Node >= 18.

3. **Glob expansion applied to correct call-site**: The glob expansion is now applied only to the jsdoc fallback call (inside the `catch` block), where `src/**/*.ts` was previously passed as a literal string argument to `jsdoc` — which is the actual bug being fixed.

Karanjot786:
Closing in favour of #1603, which applies a corrected version of this fix. The three issues with this PR were: (1) the tsc --noEmit type-check step was accidentally removed; (2) the external `glob` package was imported but is not in root package.json; (3) glob expansion was applied to the wrong call-site (the tsc block) instead of the jsdoc fallback. PR #1603 fixes all three.

PR #1576, opened 2026-06-17
perf: excessive layout recalculation in VirtualList during rapid scrolls

### LINKED ISSUE #1412: Perf : Excessive Layout Recalculation in VirtualList during rapid scrolls

The VirtualList is designed to efficiently render large data streams by slicing visible nodes. However, in terminal environments, layout calculation via the Flexbox layout engine is computationally expensive because it frequently requires clearing stream sectors and recalculating multi-dimensional boundary boxes for the visible viewport.

Currently, during high-velocity scrolling (such as fast mouse-wheel inputs or continuous key-repeat actions), the component triggers a complete, top-down layout recalculation on every single scroll frame. This causes severe CPU spikes, main-thread blocking, and noticeable terminal rendering lag/flicker.

## Proposed solution

Introduce a performance optimization path inside the layout computation engine specifically optimized for fixed-dimension elements:

- Add an Option Flag: Add a fixedItemHeight?: number or memoizeLayout?: boolean option to the VirtualList configuration schema.

- Bypass Layout Engine: If item dimensions are predictable, short-circuit the heavy Flexbox recalculation phase for scrolling frames. Calculate the direct target coordinates mathematically instead of running a full layout tree traversal.

- Memoization Cache: Cache the baseline ANSI sequence styles of the virtual layout viewport frame boundaries so that positional shifts don't force string re-allocation overhead.

 @termuijs/widgets

## GSSoC contributor?

- [x] Yes. You contribute under GSSoC 2026 and want to work on this.
- [ ] No. You are filing a suggestion.

hrx01-dev:
## Description

 Introduced `fixedItemHeight` and `memoizeLayout` options to `VirtualList` to opt into a fast-path scrolling optimization. When active, we track scroll frames and bypass layout engine invalidation on the layout nodes, short-circuiting the Flexbox engine computation to $O(1)$. Additionally, we added a render cache for line strings and cell styles to eliminate string re-allocation overhead during scrolling.
## Related Issue

Closes #1412

## Which package(s)?

@termuijs/widgets

## Type of Change

- [ ] 🐛 Bug fix (`type:bug`)
- [ ] 📝 Docs (`type:docs`)
- [x] ⚡ Performance (`type:performance`)

## Checklist

## GSSoC 2026 Participation

- Your GSSoC profile: `https://gssoc.girlscript.org/profile/91cb6e88-1eee-4f04-aecb-fe6af4626f75`

   ----

## Summary by CodeRabbit

* **New Features**
  * Virtual list now supports `fixedItemHeight` (alias of `itemHeight`) for more predictable row layout, plus `memoizeLayout` to optimize how scrolling affects redraws.
  * Added smarter per-row render caching to reduce unnecessary re-rendering when selection/focus hasn’t changed.
* **Bug Fixes**
  * Render cache now correctly resets when the list styling, total item count, or the render-item function changes.
* **Tests**
  * Expanded performance-focused tests covering fixed height rendering, memoized layout dirtying behavior, and render cache invalidation.

### DIFF
--- a/packages/widgets/src/base/Widget.ts
+++ b/packages/widgets/src/base/Widget.ts
@@ -281,6 +281,16 @@ export abstract class Widget {
     }
 
     /**
+     * Marks the widget as dirty without invalidating the layout node.
+     * Used for performance optimizations like memoized scrolling.
+     */
+    protected _markDirtyNoLayout(): void {
+        if (this._dirty) return;
+        this._dirty = true;
+        this.parent?._markDirtyNoLayout();
+    }
+
+    /**
      * Clear the dirty flag after rendering.
      * Widgets with a render error stay dirty so they are retried on the next frame.
      */
--- a/packages/widgets/src/input/VirtualList.test.ts
+++ b/packages/widgets/src/input/VirtualList.test.ts
@@ -138,4 +138,82 @@ describe('VirtualList', () => {
             expect(onSelect).not.toHaveBeenCalled();
         });
     });
+
+    describe('performance optimizations', () => {
+        it('supports fixedItemHeight configuration', () => {
+            const list = new VirtualList({
+                totalItems: 10,
+                fixedItemHeight: 3,
+                renderItem: (i) => `Item ${i}`,
+            });
+            expect((list as any)._itemHeight).toBe(3);
+        });
+
+        it('bypasses layout engine dirtying on scrolling when memoizeLayout is true', () => {
+            const list = new VirtualList({
+                totalItems: 10,
+                memoizeLayout: true,
+                renderItem: (i) => `Item ${i}`,
+            });
+
+            const layoutNode = list.getLayoutNode();
+            layoutNode._dirty = false;
+            (list as any)._dirty = false;
+
+            list.selectNext();
+
+            expect((list as any)._dirty).toBe(true);
+            expect(layoutNode._dirty).toBe(false);
+        });
+
+        it('does NOT bypass layout engine dirtying on scrolling when memoizeLayout is false', () => {
+            const list = new VirtualList({
+                totalItems: 10,
+                memoizeLayout: false,
+                renderItem: (i) => `Item ${i}`,
+            });
+
+            const layoutNode = list.getLayoutNode();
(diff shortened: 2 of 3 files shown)


Karanjot786 (changes requested):
Thanks for the contribution — the optimization direction is sound. There are two blocking issues to fix before this merges.

Issues:

1. `as any` cast to set private `_dirty` in production source
   File: packages/widgets/src/input/VirtualList.ts, in the `markDirty()` override
   Problem: `(this as any)._dirty = true` and `(p as any)._dirty = true` bypass the Widget base class's encapsulation via unsafe casts. If the base class ever renames or restructures `_dirty`, this silently breaks with no type error. Public-API-only overrides are required in production source.
   Fix: The Widget base class exposes `markDirty()` as the public mutation path. To skip layout-engine invalidation while still marking self+ancestors dirty, either (a) add a protected `_markDirtyNoLayout(): void` method to the Widget base class and call that here, or (b) expose a protected `_dirty` accessor in the base class. Either way, remove the `as any` casts from `VirtualList.ts`.

2. Private field access via `(widget as any)._field` in tests
   File: packages/widgets/src/input/VirtualList.test.ts
   Problem: Tests use `(list as any)._itemHeight`, `(list as any)._dirty`, `layoutNode._dirty`, and `(list as any)._renderCache` to assert on private implementation details. This pattern is disallowed — it couples tests to internals and breaks silently on refactors.
   Fix: Assert on observable, public behavior instead:
   - `_itemHeight`: render the widget and assert the rendered row count/positions match the expected item height.
   - `_dirty` / layout bypass: spy on `super.markDirty` or assert that render output is consistent after scroll without calling `forceLayout()` — i.e., test the visible effect, not the flag.
   - `_renderCache`: test that `renderItem` is called the expected number of times (use `vi.fn()` as the render function and count calls across a second scroll to the same position).

- [ ] Remove `as any` casts from `markDirty()` in VirtualList.ts (use a protected base-class hook instead)
- [ ] Rewrite all `(list as any)._*` and `layoutNode._dirty` assertions to use public API / render output

PR #1631, opened 2026-06-19
fix(core): reset _isRenderPending on dirty-check early return

### LINKED ISSUE #1626: [bug] `_isRenderPending` never reset on dirty-check early return causes permanent UI freeze
## Description

In `packages/core/src/app/App.ts`, the `requestRender()` method uses a `_isRenderPending` flag to coalesce multiple synchronous render requests into a single deferred frame via `setImmediate`. Inside the deferred callback, an early return path exists when the widget tree and layers are both clean (lines 368-369). This early return **bypasses the `finally` block** (line 410-412) that resets `_isRenderPending` back to `false`. Once this path is hit, every subsequent `requestRender()` call bails out at the guard on line 348, and the UI freezes permanently.

## Root Cause

The deferred callback at line 357 has three exit paths:

1. **Line 359**: Early return when `_mounted` is `false` — **correctly resets** `_isRenderPending`
2. **Line 368-369**: Early return when widget tree and layers are clean — **DOES NOT reset** `_isRenderPending`
3. **Line 410-412**: Normal return via `finally` block — **correctly resets** `_isRenderPending`

Path #2 was added as a performance optimization to skip rendering when nothing is dirty, but the guard was never released. After this path executes:

- `_isRenderPending` remains `true` forever
- Every future `requestRender()` call returns immediately at line 348
- No new `setImmediate` render callbacks are ever scheduled
- The application UI is **permanently frozen**

### Affected code

```typescript
// packages/core/src/app/App.ts, lines 343-422

requestRender(): void {
    if (!this._mounted) return;                        // line 344

    if (this._isRenderPending) return;                 // line 348 ← PERMANENTLY TRUE after bug

    this._isRenderPending = true;                      // line 350

    setImmediate(() => {
        if (!this._mounted) {
            this._isRenderPending = false;             // line 359 ✓ resets
            return;
        }

        try {
            if (this._rootWidget.isDirty === false && !this.layers.hasDirtyLayers()) {
                return;                                // line 369 ← BUG: no reset before return
            }
            // ... render logic ...
        } finally {
            this._isRenderPending = false;             // line 412 ← NEVER REACHED
        }
    });
}
```

## Trigger scenario

1. Application mounts and performs initial render (widget tree is clean afterward)
2. A periodic tick or event calls `requestRender()` → deferred callback scheduled
3. Deferred callback runs, finds no dirty widgets, **exits at line 369 without resetting flag**
4. All subsequent `requestRender()` calls silently return at line 348

This is most easily triggered when:
- The application uses `renderer.start(tick)` with a tick callback that calls `requestRender()`, but no widgets are dirty
- The initial `requestRender()` in `mount()` schedules a callback that runs after all mount-time dirty flags have been cleared
- Any scenario where a render is requested but nothing actually changed in the tree

## Impact

**Critical.** The terminal UI becomes completely unresponsive to state changes. No widget re-renders, no layout recalculations, no visual updates. The application appears hung even though its logic continues running. Users must restart the application.

## Fix required

Add `this._isRenderPending = false;` before the early return at line 369. The fix is a single line but requires careful reasoning about the render lifecycle:

```typescript
if (this._rootWidget.isDirty === false && !this.layers.hasDirtyLayers()) {
    this._isRenderPending = false;  // ← MUST reset the guard before bailing out
    return;
}
```

Additionally, tests must be added that exercise the clean-tree early-return path. The existing test suite in `App.test.ts` always marks widgets dirty before calling `requestRender()`, so the bug path has zero coverage.

### Files to modify

| File | Changes |
|------|---------|
| `packages/core/src/app/App.ts` | Add `this._isRenderPending = false` before line 369 early return |
| `packages/core/src/app/App.test.ts` | Add test case: `requestRender()` when `isDirty=false`, verify subsequent `requestRender()` schedules a new callback |

ionfwsrijan:
## Description
In `requestRender()`, the deferred callback early-returns when the widget tree and layers are both clean (line 368-369) but **bypasses the `finally` block** that resets `_isRenderPending` to `false`. Once hit, `_isRenderPending` stays `true` forever, causing every subsequent `requestRender()` to bail out at the guard — the UI freezes permanently.

## Fix
Added `this._isRenderPending = false;` before the early return on the clean-tree path, ensuring the guard flag is always released.

## Test
Added `requestRender schedules new callback after clean-tree early return` — verifies that after a clean-tree render pass, a subsequent `requestRender()` with dirty widgets still triggers rendering.

Closes #1626

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where rendering updates could be permanently skipped in certain application states.

* **Tests**
  * Added integration test verifying correct render request handling when the widget tree is in a clean state.

### DIFF
--- a/packages/core/src/app/App.test.ts
+++ b/packages/core/src/app/App.test.ts
@@ -470,7 +470,33 @@ describe('App', () => {
             await mountPromise.catch(() => {});
         });
 
-        it('unsubscribes focus handlers on unmount', async () => {
+        it('requestRender schedules new callback after clean-tree early return', async () => {
+        const root = createMockRootWidget();
+        const renderSpy = vi.fn();
+        (root as any).render = renderSpy;
+
+        const app = new App(root, createInteractiveTestOptions());
+        const mountPromise = app.mount();
+        await new Promise(r => setImmediate(r));
+
+        // Root is clean now. requestRender should hit the early return path.
+        (root as any).isDirty = false;
+        app.requestRender();
+        await new Promise(r => setImmediate(r));
+
+        // Mark dirty and request another render — should NOT be silently dropped
+        (root as any).isDirty = true;
+        const renderCountBefore = renderSpy.mock.calls.length;
+        app.requestRender();
+        await new Promise(r => setImmediate(r));
+
+        expect(renderSpy.mock.calls.length).toBeGreaterThan(renderCountBefore);
+
+        app.exit(0);
+        await mountPromise.catch(() => {});
+    });
+
+    it('unsubscribes focus handlers on unmount', async () => {
             const { root, first, second } = createFocusTestRoot();
             const app = new App(root, createInteractiveTestOptions());
 
--- a/packages/core/src/app/App.ts
+++ b/packages/core/src/app/App.ts
@@ -366,6 +366,7 @@ export class App {
                 // callback so the dirty check and the _isRenderPending guard
                 // are never racy with concurrent requestRender() calls.
                 if (this._rootWidget.isDirty === false && !this.layers.hasDirtyLayers()) {
+                    this._isRenderPending = false;
                     return;
                 }
 


ionfwsrijan:

PR #1632, opened 2026-06-19
fix(store): remove eager state mutation in batch, fix race window

### LINKED ISSUE #1627: [bug] `setState` eagerly mutates `state` inside `batch()`, creating race window where non-batched updates are silently overwritten
## Description

In `packages/store/src/store.ts`, the `batch()` API is designed to coalesce multiple `setState()` calls into a single notification, deferred to a microtask via `queueMicrotask`. However, inside the batch, `state = nextState` is eagerly assigned at line 310 **before** the microtask commit runs. This creates a window where any synchronous non-batched `setState()` or `mutate()` call that reads `getState()` sees the intermediate batched value and potentially writes changes that are later **silently overwritten** by the batch's `commit()` function.

### The corruption flow

```typescript
// store.ts lines 278-310 (simplified)

const setState: SetState<T> = (partial) => {
    const prevState = state;

    const applyUpdate = (finalPartial: Partial<T>): T => {
        const baseState = _batchDepth > 0
            ? _batchStores.get(listeners)?.nextState ?? state   // line 286
            : state;

        const nextState = { ...baseState, ...finalPartial };

        if (hasChanged) {
            if (_batchDepth > 0) {
                // ... creates/updates BatchEntry with commit() ...
                state = nextState;    // ← LINE 310: EAGER mutation
                // commit() will ALSO set state = nextState (line 302)
            } else {
                state = nextState;    // line 312
                // notify immediately
            }
        }
        return state;
    };
};
```

### The corruption scenario

```typescript
batch(() => {
    store.setState({ x: 1 });   // line 310: state = { x: 1 } immediately
});
// At this point:
//   - batch() has returned
//   - queueMicrotask(commit) is scheduled but NOT YET RUN
//   - state is already { x: 1 } (from the eager mutation at line 310)

// ◈ RACE WINDOW OPENS ◈

store.setState({ y: 2 });       // NOT in a batch, runs synchronously
// getState() returns { x: 1 }
// line 312: state = { x: 1, y: 2 }
// Listeners notified with { x: 1, y: 2 }

// ◈ Microtask fires ◈

// commit() runs: state = { x: 1 }  ← OVERWRITES y: 2!
// Listeners notified again with { x: 1 }

// RESULT: y: 2 is SILENTLY LOST!
```

### The same bug exists in `mutate()`

Lines 381-407 of `store.ts` have the identical pattern:

```typescript
mutate(recipe: (draft: T) => void): void {
    const prevState = state;
    const nextState = produce(state, (draft) => { recipe(draft as T); });
    state = nextState;              // ← line 389: eager mutation
    if (_batchDepth > 0) {
        // commit will also set state = nextState (line 396)
    }
}
```

## Impact

**Critical — silent data corruption.** The store's state becomes inconsistent without any error or warning. This is especially dangerous because:

1. The corruption is **non-deterministic** — it depends on microtask timing
2. The corruption is **silent** — no error is thrown, no warning logged
3. Persisted state (`persistState()`) stores the **stale** value, corrupting on-disk state
4. Subscribers see **two notifications** — first with correct data, then with stale data — causing UI flicker and potential infinite render loops if a subscriber reacts to the stale value

## Fix required

The fix requires **not eagerly mutating `state`** inside a batch, and instead making `getState()` return the pending batch state when called from within a batch context:

### Changes in `packages/store/src/store.ts`

1. **Remove line 310**: `state = nextState;` — only `commit()` should set the final state
2. **Remove line 389**: `state = nextState;` in `mutate()` — same reason
3. **Modify `getState()`** to return pending batch state when a batch is in progress:

```typescript
const getState: GetState<T> = () => {
    if (_batchDepth > 0) {
        const entry = _batchStores.get(listeners);
        if (entry) return entry.nextState;
    }
    return state;
};
```

4. **Update `commit()` closures** so they capture the correct final `nextState` at commit time (they already do, since the closure captures `nextState` from the `applyUpdate` scope)

### Files to modify

| File | Changes |
|------|---------|
| `packages/store/src/store.ts` | Remove eager `state = nextState` inside batch paths (lines 310, 389); update `getState` to return pending batch state |
| `packages/store/src/store.test.ts` | Add test: `setState` outside batch after `batch()` call verifies data is not overwritten by commit |

ionfwsrijan:
## Description
In `batch()`, `setState` eagerly assigns `state = nextState` at line 310 **before** the microtask commit runs. This creates a race window where any synchronous non-batched `setState()` or `mutate()` call that reads `getState()` sees the intermediate batched value and writes changes that are later **silently overwritten** by the batch's `commit()`.

The same bug exists in `mutate()` at line 389.

## Fix
1. **Removed** `state = nextState` inside the batch path in `setState` (line 310) — only `commit()` sets final state
2. **Removed** `state = nextState` inside the batch path in `mutate` (line 389) — moved to non-batch path only
3. **Updated** `getState()` to return pending batch `nextState` when `_batchDepth > 0`, so in-batch reads see correct intermediate state
4. **Updated** `setState` function updaters to receive pending batch state when inside a batch
5. **Updated** `mutate` to produce from pending batch state when inside a batch

## Tests
- `setState outside batch after batch call is not overwritten by commit`
- `getState inside batch returns pending batch state`

Closes #1627

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved batching behavior for state updates and reads. `getState()` now correctly reflects pending updates during an active batch, and state mutations within batches compute from the in-flight state rather than the previously committed state, ensuring more consistent behavior when batching multiple updates together.

### DIFF
--- a/packages/store/src/store.test.ts
+++ b/packages/store/src/store.test.ts
@@ -388,6 +388,38 @@ describe('middleware', () => {
         // No listeners should have fired
         expect(spy).not.toHaveBeenCalled()
     })
+
+    it('setState outside batch after batch call is not overwritten by commit', async () => {
+        const useStore = createStore((set) => ({
+            x: 0,
+            y: 0,
+        }))
+        const spy = vi.fn()
+        useStore.subscribe(spy)
+
+        batch(() => {
+            useStore.setState({ x: 1 })
+        })
+
+        // Between batch return and microtask, setState outside batch
+        useStore.setState({ y: 2 })
+
+        await new Promise(resolve => queueMicrotask(resolve))
+
+        expect(useStore.getState()).toEqual({ x: 1, y: 2 })
+    })
+
+    it('getState inside batch returns pending batch state', () => {
+        const useStore = createStore((set) => ({
+            count: 0,
+        }))
+
+        batch(() => {
+            useStore.setState({ count: 1 })
+            // Inside the batch, getState should see the pending value
+            expect(useStore.getState().count).toBe(1)
+        })
+    })
 })
 
 describe('persistence', () => {
--- a/packages/store/src/store.ts
+++ b/packages/store/src/store.ts
@@ -32,7 +32,8 @@ import type { EqualityFn } from './shallow.js'
 interface BatchEntry<T> {
     prevState: T;
     nextState: T;
-    commit: () => void;
+    changes: Partial<T>;
+    commit: () => T;
     rollback: () => void;
 }
 
@@ -108,10 +109,10 @@ function flushBatch(threw: boolean) {
         queueMicrotask(() => {
             const stores = Array.from(_batchStores.entries());
             _batchStores.clear();
-            for (const [listeners, { prevState, nextState ,commit }] of stores) {
-                commit();
+            for (const [listeners, { prevState, commit }] of stores) {
+                const newState = commit();
                 for (const listener of listeners) {
-                    listener(nextState, prevState);
+                    listener(newState, prevState);
                 }
             }
         });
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #1633, opened 2026-06-19
fix(core): implement multi-chunk bracketed paste state machine

### LINKED ISSUE #1628: [bug] Multi-chunk bracketed paste silently discards content when paste spans multiple stdin data chunks
## Description

In `packages/core/src/input/InputParser.ts`, the bracketed paste handler (lines 134-145) requires **both** the paste-start marker `\x1b[200~` **and** the paste-end marker `\x1b[201~` to be present **in the same data chunk** to trigger. When a paste spans multiple stdin chunks (typically > 4KB, common for any non-trivial paste), the first chunk contains `\x1b[200~` + partial content, and subsequent chunks contain more content + `\x1b[201~`. Neither chunk alone contains both markers, so the paste handler never fires. Instead, the `\x1b[200~` sequence falls through to the escape sequence parser, which eventually discards it.

Additionally, the `_isPasting` field (line 39) and `_pasteBuffer` field (line 40) are declared but **never read or written** — confirming that cross-chunk paste tracking was architecturally planned but never implemented.

### Paste detection requires both markers in one chunk

```typescript
// InputParser.ts lines 134-145
private _processInput(data: Buffer): void {
    // ...
    const str = data.toString('utf8');
    const PASTE_START = '\x1b[200~';
    const PASTE_END = '\x1b[201~';

    if (str.includes(PASTE_START) && str.includes(PASTE_END)) {   // ← MUST be same chunk
        const pastedText = str
            .replace(PASTE_START, '')
            .replace(PASTE_END, '');
        this._events.emit('paste', pastedText);
        return;
    }
    // ...
}
```

### When paste spans chunks, the start marker triggers escape sequence parsing

```typescript
// Line 166-169 — caught by this branch:
if (str.startsWith('\x1b')) {
    this._escapeBuffer = data;       // buffers the entire chunk
    this._tryParseEscape();          // tries to parse as escape sequence
    return;
}
```

### The escape parser discards unrecognized long buffers

```typescript
// Lines 372-376 — safety cap silently drops the data:
if (seq.length > 20) {
    this._escapeBuffer = Buffer.alloc(0);  // ← SILENTLY DISCARDED
    return;
}
```

### Concrete trace with a 2-chunk paste

**Chunk 1** (e.g., 4KB of pasted content): `\x1b[200~Lorem ipsum dolor sit amet...`
1. `_escapeBuffer.length > 0` → false (fresh state)
2. `str.includes(PASTE_START) && str.includes(PASTE_END)` → false (no END marker)
3. `str.startsWith('\x1b') && str.length === 1` → false (length > 1)
4. `str.startsWith('\x1b')` → true → buffers entire chunk in `_escapeBuffer`
5. `_tryParseEscape()`:
   - `isMouseSequence(seq)` → false
   - `cursorMatch` → null
   - Not focus tracking, not in `ESCAPE_SEQUENCES`
   - Not Alt+key pattern (seq.length > 2)
   - `seq.length > 20` → **true** → `_escapeBuffer = Buffer.alloc(0)` → **data lost!**
   - Returns

**Chunk 2**: `consectetur adipiscing elit...\x1b[201~`
1. Same flow, but `\x1b` might not be at position 0 (it's in the middle)
2. Falls through to regular character processing, emitting each character as individual key events
3. The `\x1b[201~` at the end gets caught by the escape parser as unknown

**Result**: The pasted text is **completely lost** from the first chunk, and the second chunk's content is emitted as individual keypresses (corrupting whatever state the app expected from a paste event).

## Impact

**Critical — silent data loss.** Any paste operation larger than a single stdin chunk (typically 4KB, though this varies by platform and terminal) results in:

1. **Complete loss of the first N bytes** of pasted content (silently discarded by the escape parser)
2. **Individual key events** emitted for the remaining bytes, interpreted as user keystrokes
3. **No `paste` event ever fires**, so the application never knows a paste occurred
4. **Potential security concern**: if the user intended to paste sensitive data into a password field, the partial content emitted as key events may leak to other handlers

This affects any application using `onPaste()` for handling pasted content — search, form inputs, code editors, etc.

## Fix required

Implement a proper paste state machine using the existing (but dead) `_isPasting` and `_pasteBuffer` fields:

### Changes in `packages/core/src/input/InputParser.ts`

1. **Remove the single-chunk check** (lines 138-145)
2. **Handle paste-start marker** by activating paste state tracking:
   - Set `_isPasting = true`
   - Start accumulating content in `_pasteBuffer`
   - Do NOT fall through to escape sequence parsing
3. **Handle paste-end marker** by completing the paste:
   - Set `_isPasting = false`
   - Emit accumulated `_pasteBuffer` as a paste event
   - Clear `_pasteBuffer`
4. **Accumulate partial content** between markers into `_pasteBuffer`
5. **Handle edge cases**:
   - Chunk that contains both start marker + partial content + end marker
   - Chunk boundaries that split marker sequences (e.g., `\x1b[20` in one chunk, `0~` in the next)
   - Timeout / abort if paste starts but never ends (stale paste state recovery)

### Files to modify

| File | Changes |
|------|---------|
| `packages/core/src/input/InputParser.ts` | Implement multi-chunk paste state machine; wire `_isPasting` and `_pasteBuffer`; add timeout recovery for incomplete pastes |
| `packages/core/src/input/InputParser.test.ts` | Add tests for: single-chunk paste, multi-chunk paste, chunk-boundary marker split, timeout recovery |

ionfwsrijan:
## Description
The bracketed paste handler required both `\x1b[200~` and `\x1b[201~` markers in the **same data chunk**. When paste content spans multiple stdin chunks (>4KB), the first chunk contains the start marker + partial content, and subsequent chunks contain more content + the end marker. The paste handler never fires, and the start marker falls through to the escape parser which silently discards it.

The `_isPasting` and `_pasteBuffer` fields were declared but never used.

## Fix
Replaced the single-chunk paste check with a proper state machine:
1. **Paste-start detection** — sets `_isPasting = true`, strips start marker, buffers content
2. **Paste-end detection** — emits accumulated `_pasteBuffer` as a `paste` event, resets state
3. **Cross-chunk accumulation** — content between markers is concatenated across chunks
4. **Same-chunk handling** — both markers in one chunk still works
5. **Timeout recovery** — 500ms inactivity timeout aborts stale paste state

## Tests
- Single-chunk paste (existing)
- Multi-chunk paste (two chunks)
- Three-chunk paste
- Both markers in one chunk with surrounding text
- Incomplete paste timeout recovery

Closes #1628

### DIFF
--- a/packages/core/src/input/InputParser.test.ts
+++ b/packages/core/src/input/InputParser.test.ts
@@ -278,5 +278,77 @@ describe('InputParser', () => {
         sendKey(stdin, '\x1b[200~hello world\x1b[201~');
 
         expect(pasteHandler).toHaveBeenCalledWith('hello world');
-    }); 
+    });
+
+    it('handles multi-chunk paste spanning multiple data chunks', () => {
+        const stdin = createMockStdin();
+        const parser = new InputParser(stdin);
+
+        const pasteHandler = vi.fn();
+
+        parser.onPaste(pasteHandler);
+        parser.start();
+
+        // First chunk: paste-start marker + partial content
+        stdin.emit('data', Buffer.from('\x1b[200~hello ', 'utf8'));
+        expect(pasteHandler).not.toHaveBeenCalled();
+
+        // Second chunk: rest of content + paste-end marker
+        stdin.emit('data', Buffer.from('world\x1b[201~', 'utf8'));
+
+        expect(pasteHandler).toHaveBeenCalledWith('hello world');
+    });
+
+    it('handles paste with content split across three chunks', () => {
+        const stdin = createMockStdin();
+        const parser = new InputParser(stdin);
+
+        const pasteHandler = vi.fn();
+
+        parser.onPaste(pasteHandler);
+        parser.start();
+
+        stdin.emit('data', Buffer.from('\x1b[200~abc', 'utf8'));
+        stdin.emit('data', Buffer.from('def', 'utf8'));
+        stdin.emit('data', Buffer.from('ghi\x1b[201~', 'utf8'));
+
+        expect(pasteHandler).toHaveBeenCalledWith('abcdefghi');
+    });
+
+    it('emits paste event when both markers are in a single chunk', () => {
+        const stdin = createMockStdin();
+        const parser = new InputParser(stdin);
+
+        const pasteHandler = vi.fn();
+
+        parser.onPaste(pasteHandler);
+        parser.start();
+
+        stdin.emit('data', Buffer.from('before\x1b[200~content\x1b[201~after', 'utf8'));
+
+        expect(pasteHandler).toHaveBeenCalledWith('content');
+    });
+
+    it('recovers from partial paste via timeout', () => {
+        vi.useFakeTimers();
+        const stdin = createMockStdin();
+        const parser = new InputParser(stdin);
+
+        const pasteHandler = vi.fn();
+
+        parser.onPaste(pasteHandler);
+        parser.start();
+
+        // Start paste but never end it
+        stdin.emit('data', Buffer.from('\x1b[200~incomplete', 'utf8'));
+        expect(pasteHandler).not.toHaveBeenCalled();
+
+        // Advance time past the paste timeout
+        vi.advanceTimersByTime(600);
+
+        // Handler should not have been called (paste was aborted)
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #1635, opened 2026-06-19
fix(core): adjust pushClip for active translateY to fix nested clip coordinate mismatch

### LINKED ISSUE #1630: [bug]  Nested `pushClip`/`pushTranslateY` coordinate mismatch causes invisible content in nested ScrollViews
## Description

In `packages/core/src/terminal/Screen.ts`, the clipping and translation systems operate on different coordinate bases when nested. `pushClip()` stores clip regions in **absolute terminal coordinates**, while `setCell()` applies `_translateY` to the row **before** checking the clip. When a parent widget applies a translate (e.g., ScrollView scrolling content) and a child widget pushes a clip (e.g., nested ScrollView), the child's clip region is stored in absolute coordinates that don't account for the parent's translate offset. Content that should be visible inside the inner clip is incorrectly clipped out, causing invisible content.

### The coordinate mismatch

```typescript
// Screen.ts lines 293-309 — setCell applies translate FIRST, then checks clip
setCell(col: number, row: number, cell: Partial<Cell>): void {
    col = Math.floor(col);
    row = Math.floor(row);
    row += this._translateY;    // ← Apply translate in CURRENT space
    
    // ... bounds check ...
    
    if (this._clipStack.length > 0) {
        const clip = this._clipStack[this._clipStack.length - 1];
        if (col < clip.x || col >= clip.x + clip.width ||
            row < clip.y || row >= clip.y + clip.height) {
            return;  // ← Clip check in TRANSLATED space
        }
    }
}
```

```typescript
// Screen.ts lines 245-261 — pushClip stores region AS-IS
pushClip(region: { x: number; y: number; width: number; height: number }): void {
    // ... no adjustment for _translateY ...
    this._clipStack.push({ ...region });  // ← Stored in ABSOLUTE space
}
```

### ScrollView usage pattern

```typescript
// ScrollView.ts lines 100-117
override render(screen: Screen): void {
    // ...
    screen.pushClip(this._rect);              // LINE 104: clip in ABSOLUTE coords
    this._renderSelf(screen);
    this._renderBorder(screen);
    
    screen.pushTranslateY(-this._scrollOffset); // LINE 111: translate applied AFTER clip
    for (const child of this._children) {
        child.render(screen);                   // Children render with active translate
    }
    screen.popTranslateY();
    screen.popClip();                           // LINE 117
}
```

### Concrete bug scenario (single nesting fails in edge cases)

Consider an outer ScrollView at `y=10`, `height=20`, `scrollOffset=5`. Its child is an inner ScrollView at absolute `y=15`, `height=10`, `scrollOffset=3`.

**Outer ScrollView.render():**
1. `pushClip({y:10, height:20})` — clip stored at y=10 (absolute)
2. `pushTranslateY(-5)` — `_translateY = -5`

**Inner ScrollView.render()** (during outer's child iteration):
3. `pushClip({y:15, height:10})` — clip stored at y=15 (absolute, BUT current translated space is -5)
4. `pushTranslateY(-3)` — `_translateY = -5 + (-3) = -8`

**Child widget inside inner renders at absolute y=16:**
5. `setCell(x, 16, attrs)`:
   - `row = 16 + (-8) = 8` (translated row)
   - Check inner clip: `8 >= 15 && 8 < 15+10`? → `8 >= 15`? **NO → CLIPPED OUT!**

**But should it be visible?** The inner ScrollView at y=15 with scrollOffset=3 shows content shifted up by 3. The child at y=16 should appear at visual row \(16 - 5 - 3 = 8\). The inner's visual position after the outer's translate is \(15 - 5 = 10\), so its viewport is \[10, 20) in visual coordinates. Row 8 is ABOVE this viewport...

Wait — let me reconsider. The child at y=16 is 1 row below the inner's top (y=15). With scrollOffset=3, the content shifts up by 3, moving the child to y=13 visual (before outer translate). The outer shifts by -5, moving it to y=8. But the inner's viewport in absolute coordinates is [15, 25). After outer translate: [10, 20). Row 8 is above y=10.

So the child at y=16 with scrollOffset=3 IS correctly clipped in this specific case. But the **clip region itself is stored incorrectly.** The clip at y=15 (absolute) was pushed when `_translateY = -5`. The visual clip SHOULD be at y = 15 + (-5) = 10. But it's stored as y=15, and the check uses visual coordinates (row=8). So 8 >= 15 fails, which is... coincidentally correct in this case.

### The actual bug

The bug manifests when the clip region's y value in the stored coordinate space doesn't match the visual coordinate space used in `setCell`. Here's a scenario that demonstrates the error:

**Outer ScrollView at y=10, scrollOffset=5**  
**Inner widget at y=14, height=4** (NOT a ScrollView — just a regular container that clips its content)

The inner widget clips at y=14. After outer translate of -5, the clip should be at visual y=9. But it's stored at y=14.

**Child at absolute y=15 (the first content row inside the inner widget):**
- After outer translate (-5): visual row = 10
- Check inner clip at stored y=14: `10 >= 14`? NO → clipped!

But the child SHOULD be visible! The inner viewport at y=14 with height=4 spans rows 14-17 (absolute). After outer scrollOffset=5, these become rows 9-12 (visual). The child at y=15 (absolute) should appear at visual row 10, which IS inside [9, 13).

The bug: clip is stored at y=14 (absolute), but setCell checks against visual row=10. Since `10 < 14`, the child is incorrectly clipped.

**Conversely, if the outer has NO scroll (translateY=0):**
- Clip stored at y=14
- Child at y=15: visual row = 15
- Check: `15 >= 14` → YES → renders correctly ✓

The bug ONLY manifests when there's an active translate that was applied BEFORE a clip was pushed.

### The mathematical error

When `pushClip` is called with `translateY = N`:
- Caller passes region in **absolute** coordinates: `{y: Y, height: H}`
- `setCell` translates row: `visual_row = absolute_row + translateY`
- Clip check: `visual_row >= stored_y && visual_row < stored_y + stored_height`

For correct clipping at absolute boundaries Y to Y+H:
- `visual_row >= Y + translateY` AND `visual_row < Y + H + translateY`
- Therefore: `stored_y` should be `Y + translateY`, stored_height should be `H`

**Current (buggy):** stores `y = Y` (no translate adjustment)  
**Fix:** store `y = Y + translateY`

## Impact

**Critical — content is silently invisible.** In any layout with nested translated/clipped containers:

1. Content inside inner clipped regions is **silently invisible** when an outer widget applies a translate (scrolling, offset, etc.)
2. The bug is layout-order and nesting-depth dependent — adding a wrapper widget can make content disappear
3. No error or warning is emitted
4. Affects not just nested ScrollViews, but any combination of `pushTranslateY` + `pushClip`:
   - ScrollView inside ScrollView
   - Animated/offset containers with clipping
   - Inline views containing scrollable sections
5. Debugging is extremely difficult because the content simply doesn't render

### In `packages/core/src/terminal/Screen.ts`

**Option A: Adjust clip region for current translateY in `pushClip`**

```typescript
pushClip(region: { x: number; y: number; width: number; height: number }): void {
    // Convert absolute region to current visual coordinate space
    const adjustedRegion = {
        x: region.x,
        y: region.y + this._translateY,  // Account for active translate
        width: region.width,
        height: region.height,
    };
    
    if (this._clipStack.length > 0) {
        const parent = this._clipStack[this._clipStack.length - 1];
        const x = Math.max(adjustedRegion.x, parent.x);
        const y = Math.max(adjustedRegion.y, parent.y);
        const right = Math.min(adjustedRegion.x + adjustedRegion.width, parent.x + parent.width);
        const bottom = Math.min(adjustedRegion.y + adjustedRegion.height, parent.y + parent.height);
        // ...
        this._clipStack.push({ x, y, width: right - x, height: bottom - y });
    } else {
        this._clipStack.push({ ...adjustedRegion });
    }
}
```

**Option B: Track per-clip translate offset**

Store the `_translateY` value alongside each clip entry and apply it during `setCell`:

```typescript
// Push: store both the region and the translate offset at push time
this._clipStack.push({ ...region, translateOffset: this._translateY });

// SetCell: check against stored region adjusted by its translate offset
const rowCheck = row + this._translateY;  // full translate
const clipRowCheck = row + clip.translateOffset;  // clip-space translate
if (col < clip.x || col >= clip.x + clip.width ||
    clipRowCheck < clip.y || clipRowCheck >= clip.y + clip.height) {
    return;
}
```

### Files to modify

| File | Lines | Changes |
|------|-------|---------|
| `packages/core/src/terminal/Screen.ts` | 245-261 | Fix `pushClip` to account for active `_translateY` when storing clip region |
| `packages/core/src/terminal/Screen.ts` | 293-309 | Verify `setCell` clip check is consistent with new coordinate space |
| `packages/widgets/src/layout/ScrollView.ts` | 100-117 | Verify ScrollView's clip/translate pattern with fix; could also reorder operations |
| `packages/core/src/terminal/Screen.test.ts` | — | Add tests: clip+translate ordering, nested clips with active translate, scrollview in scrollview rendering |

ionfwsrijan:
## Description
`pushClip()` stores clip regions in **absolute terminal coordinates**, while `setCell()` applies `_translateY` to the row **before** checking the clip. When a parent widget applies a translate (e.g., ScrollView scrolling) and a child widget pushes a clip, the child's clip region doesn't account for the parent's translate offset. Content inside the inner clip is incorrectly clipped out.

## Fix
Adjusted `pushClip()` to convert the input region from absolute to visual coordinates by adding the current `_translateY` to `region.y` before storing. This makes the clip check in `setCell()` consistent with the translated coordinate space.

## Tests
- `pushClip with active translateY adjusts clip region correctly`
- `nested clips with active translate work correctly`
- `setCell is clipped by translate-adjusted region when parents have translate`

Closes #1630

### DIFF
--- a/packages/core/src/terminal/Screen.test.ts
+++ b/packages/core/src/terminal/Screen.test.ts
@@ -131,6 +131,75 @@ describe('Screen', () => {
         // Old data is reset
         expect(screen.back[0][0].char).toBe(' ');
     });
+
+    it('pushClip with active translateY adjusts clip region correctly', () => {
+        const screen = new Screen(20, 10);
+        // Push a translate first (simulating a parent ScrollView)
+        screen.pushTranslateY(-3);
+        // Clip region should be adjusted by the current translate
+        screen.pushClip({ x: 0, y: 5, width: 10, height: 4 });
+
+        // setCell applies translate: row 8 + (-3) = 5. With adjusted clip at y=2 (5-3),
+        // 5 >= 2 and 5 < 6 should pass → visible
+        screen.setCell(1, 8, { char: 'A' });
+        expect(screen.back[5][1].char).toBe('A');
+
+        // setCell translates row 9 + (-3) = 6. Clip is at y=2, h=4 → y range [2, 6)
+        // 6 is NOT < 6 → clipped
+        screen.setCell(1, 9, { char: 'B' });
+        expect(screen.back[6][1].char).not.toBe('B');
+
+        screen.popClip();
+        screen.popTranslateY();
+    });
+
+    it('nested clips with active translate work correctly', () => {
+        const screen = new Screen(30, 20);
+        // Outer translate (simulating parent ScrollView with scrollOffset=2)
+        screen.pushTranslateY(-2);
+        screen.pushClip({ x: 0, y: 5, width: 20, height: 10 });
+
+        // Inner translate (simulating child ScrollView with scrollOffset=1)
+        screen.pushTranslateY(-1);
+        screen.pushClip({ x: 0, y: 8, width: 10, height: 5 });
+
+        // Total translate: -3. inner clip absolute y=8 → adjusted y = 8-3 = 5 in setCell space.
+        // Inner clip visual range: y ∈ [5, 10), row ∈ [8, 13) absolute.
+        // Content row 8 absolute → visual row 8-3=5 → inside clip (5 >= 5 && 5 < 10) ✓
+        screen.setCell(1, 8, { char: 'X' });
+        expect(screen.back[5][1].char).toBe('X');
+
+        // Content row 13 absolute → visual row 13-3=10 → outside clip (10 >= 10) → clipped
+        screen.setCell(1, 13, { char: 'Y' });
+        expect(screen.back[10][1].char).not.toBe('Y');
+
+        screen.popClip();
+        screen.popTranslateY();
+        screen.popClip();
+        screen.popTranslateY();
+    });
+
+    it('setCell is clipped by translate-adjusted region when parents have translate', () => {
+        const screen = new Screen(30, 20);
+        // Simulate: outer container at absolute y=10 with scrollOffset=5
+        screen.pushTranslateY(-5);
+        // Child clip region at absolute y=14, height=4
+        screen.pushClip({ x: 0, y: 14, width: 10, height: 4 });
+
+        // After translate: clip adjusted to y = 14 + (-5) = 9, height=4 → visual range [9, 13)
+        // Content that would render at absolute row 15 (first row inside child viewport)
+        // visual row = 15 + (-5) = 10 → inside clip (10 >= 9 && 10 < 13) ✓
+        screen.setCell(0, 15, { char: 'A' });
+        expect(screen.back[10][0].char).toBe('A');
+
+        // Content at absolute row 19 (outside child viewport)
+        // visual row = 19 + (-5) = 14 → outside clip (14 >= 13) → clipped
+        screen.setCell(0, 19, { char: 'B' });
+        expect(screen.back[14][0].char).not.toBe('B');
+
+        screen.popClip();
+        screen.popTranslateY();
+    });
 });
 
 describe('cellsEqual', () => {
--- a/packages/core/src/terminal/Screen.ts
+++ b/packages/core/src/terminal/Screen.ts
@@ -243,13 +243,22 @@ export class Screen {
      * any parent clip already on the stack (nested clipping).
      */
     pushClip(region: { x: number; y: number; width: number; height: number }): void {
+        // Convert region from absolute to the current visual coordinate space.
+        // setCell applies _translateY before checking the clip stack, so stored
+        // clip regions must be in the same visual space for the check to be correct.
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #1663, opened 2026-06-19
fix(store): prevent batch race condition causing silent state loss in nested batches

### LINKED ISSUE #1658: [bug] Store batch() race condition causes silent state loss in nested batches
## Description

The `batch()` function in `@termuijs/store` has a critical race condition where nested batch calls can silently lose intermediate state mutations.

### Root Cause

In `packages/store/src/store.ts`, when `batch()` is called inside another batch:

1. The outer batch commits first, setting `_batchStores` entries
2. The inner batch finds an existing entry and replaces `existing.nextState` without applying intermediate mutations
3. The `commit` closure captures the new `nextState` but the old `commit` closure is discarded
4. If middleware mutates state between these updates, those intermediate mutations are silently lost

Additionally, `flushBatch()` defers notification via `queueMicrotask()` (line 108). If a new `batch()` starts and completes before the microtask fires, `_batchStores` is cleared before the microtask reads it, but then a new batch populates it again. The stale microtask reads the *new* batch entries and dispatches incorrect notifications. There is no generation counter or epoch guard to prevent this.

### Impact

- State updates in nested batches can be silently overwritten
- Middleware state transformations may be lost
- Stale microtask dispatches incorrect notifications for wrong batch
- Applications using complex state patterns will experience unpredictable behavior
- Data corruption in production applications

### Steps to Reproduce

```typescript
import { createStore, batch } from '@termuijs/store';

const store = createStore({ count: 0, name: '' });

batch(() => {
  store.setState({ count: 1 });
  batch(() => {
    store.setState({ name: 'test' }); // This may lose the count update
  });
});
```

### Expected Behavior

All state mutations within nested batches should be applied in order without loss. Microtask-based dispatch should use epoch guards to prevent stale dispatches.

### Actual Behavior

Intermediate state mutations are silently overwritten. Stale microtasks dispatch incorrect notifications.

### Environment

- TermUI version: latest
- Package: @termuijs/store

### Suggested Fix

1. Add an epoch/generation counter to `_batchStores` entries so stale microtasks can be detected and skipped
2. When an existing batch entry is found, merge the next state rather than replacing it
3. Update the `rollback` closure whenever `nextState` is updated in `mutate()`

atul-upadhyay-7:
## Summary

Fixes a critical race condition in `@termuijs/store`'s `batch()` mechanism where nested batch calls could silently lose intermediate state mutations, and where a stale microtask could dispatch notifications for the wrong batch.

Closes #1658

### 1. Epoch guard for microtask dispatch

Added `_batchEpoch` counter that increments when an outermost batch starts. `flushBatch()` snapshots the epoch before scheduling the microtask. If a new batch starts before the microtask fires, the epoch will have changed and the stale microtask bails out without dispatching.

**Before:** Two consecutive batches could both dispatch, with the first microtask reading the second batch's entries.

**After:** The first microtask detects the epoch mismatch and skips dispatch.

### 2. `mutate()` updates rollback closure in existing batch entries

When `mutate()` found an existing batch entry, it only updated `existing.nextState` -- the `commit` and `rollback` closures were never updated. This meant:
- The rollback restored to an intermediate state, not the true pre-batch state
- The commit captured a stale `nextState`

**After:** Both `commit` and `rollback` closures are updated when an existing entry is found.

### 3. Cleanup: removed unused `prevState` destructuring in error path

The `flushBatch(true)` path destructured `prevState` but never used it.

## Tests added

| Test | What it verifies |
|------|-----------------|
| nested batch does not lose intermediate state | Outer `setState` survives inner `setState` |
| consecutive batches do not interfere | Two sequential batches both apply correctly |
| stale microtask from old batch does not dispatch | Epoch guard prevents double-dispatch |
| mutate inside batch merges correctly with setState | `mutate()` + `setState()` coalesce properly |
| multiple mutate calls inside batch coalesce | Three `mutate()` calls produce one notification |
| batch rollback restores state before any updates including mutate | Error in batch with `mutate()` rolls back completely |

## Verification

- All 91 store tests pass
- Typecheck passes across all packages
- No breaking changes to public API

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling and reliability of batch operations when nested or overlapping.
  * Enhanced error recovery with better rollback semantics for failed batch updates.
  * Fixed state consistency issues that could occur across overlapping batch cycles.
  * Optimized microtask scheduling to prevent stale state from being committed.

* **Tests**
  * Extended test suite with comprehensive coverage for advanced batch scenarios.

### DIFF
--- a/packages/store/src/store.test.ts
+++ b/packages/store/src/store.test.ts
@@ -290,6 +290,136 @@ describe('batch', () => {
         expect(spy).toHaveBeenCalledOnce()
         expect(spy.mock.calls[0][0]).toEqual({ a: 1, b: 2 })
     })
+
+    it('nested batch does not lose intermediate state from outer batch', async () => {
+        const useStore = createStore((set) => ({
+            count: 0,
+            name: '',
+        }))
+        const spy = vi.fn()
+        useStore.subscribe(spy)
+
+        batch(() => {
+            useStore.setState({ count: 1 })
+            batch(() => {
+                useStore.setState({ name: 'test' })
+            })
+        })
+
+        await new Promise(resolve => queueMicrotask(resolve))
+
+        expect(spy).toHaveBeenCalledOnce()
+        expect(useStore.getState()).toEqual({ count: 1, name: 'test' })
+    })
+
+    it('consecutive batches do not interfere with each other', async () => {
+        const useStore = createStore((set) => ({
+            a: 0,
+            b: 0,
+        }))
+        const spy = vi.fn()
+        useStore.subscribe(spy)
+
+        batch(() => {
+            useStore.setState({ a: 1 })
+        })
+
+        batch(() => {
+            useStore.setState({ b: 2 })
+        })
+
+        await new Promise(resolve => queueMicrotask(resolve))
+
+        expect(useStore.getState()).toEqual({ a: 1, b: 2 })
+    })
+
+    it('stale microtask from old batch does not dispatch when new batch starts', async () => {
+        const useStore = createStore((set) => ({
+            x: 0,
+        }))
+        const spy = vi.fn()
+        useStore.subscribe(spy)
+
+        // Start first batch but manually prevent its microtask from firing
+        batch(() => {
+            useStore.setState({ x: 1 })
+        })
+
+        // Immediately start a second batch before the first microtask fires
+        batch(() => {
+            useStore.setState({ x: 2 })
+        })
+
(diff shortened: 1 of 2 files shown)


Karanjot786:
Changes merged directly into main (conflict resolved — HEAD version used undefined newState, kept main's `const newState = commit()`). Epoch-counter fix is in main as of 031f5fd.

PR #1664, opened 2026-06-19
fix(core): include full color value in getStyleLine hash to prevent collision

### LINKED ISSUE #1659: [bug] Screen.setStyleLine hash collision silently skips color code updates
## Description

The `Screen.setStyleLine` method computes a hash that only uses the first character of the color type string, causing hash collisions that silently skip style-only updates to color codes.

### Root Cause

In `packages/core/src/terminal/Screen.ts`, the `getStyleLine()` method computes a hash using only `fg.charCodeAt(0)` and `bg.charCodeAt(0)`. This means:

- Two cells with `type: 'ansi256', code: 1` and `type: 'ansi256', code: 9` produce the same hash
- Both have `'ansi256'.charCodeAt(0) === 97`
- Only the *first character of the type string* is used, not the actual code value

### Impact

- Color updates within the same type (e.g., changing ANSI256 code 1 to code 9) are silently skipped by the diff renderer
- The diff renderer thinks nothing changed and does not emit the update
- Users see stale colors until a full re-render occurs
- Visual glitches in color-changing animations or dynamic UIs
- Any style-only update that changes the color code but not the type string is invisible

### Steps to Reproduce

```typescript
import { Screen } from '@termuijs/core';

const screen = new Screen(40, 10);

// Set cell to ANSI256 color code 1
screen.setStyle(0, 0, { fg: { type: 'ansi256', code: 1 } });

// Change to ANSI256 color code 9 (different color, same type)
screen.setStyle(0, 0, { fg: { type: 'ansi256', code: 9 } });

// Hash collision means this update is skipped by diff renderer
```

### Expected Behavior

Any change to color code should produce a different hash and trigger a visual update.

### Actual Behavior

Color code changes within the same type produce the same hash and are skipped.

### Environment

- TermUI version: latest
- Package: @termuijs/core

### Suggested Fix

Include the full color code/value in the hash computation, not just the first character of the type string. For example:

```typescript
// Before (broken)
return ((fg.charCodeAt(0) << 16) | (bg.charCodeAt(0) << 8) | flags) >>> 0;

// After (fixed) - include the actual code value
const fgHash = fg ? (fg.type.charCodeAt(0) ^ fg.code) : 0;
const bgHash = bg ? (bg.type.charCodeAt(0) ^ bg.code) : 0;
return ((fgHash << 16) | (bgHash << 8) | flags) >>> 0;
```

atul-upadhyay-7:
## Summary

Fixes a critical bug where `Screen.getStyleLine()` produced hash collisions that caused the diff renderer to silently skip color-only updates. Two cells with the same color type but different codes (e.g. `ansi256` code 1 vs code 9) generated the same hash, so the renderer treated the row as unchanged.

Closes #1659

## Root Cause

`getStyleLine()` computed a hash seed using only `fg.charCodeAt(0)` and `bg.charCodeAt(0)` on the color type string. Since `'ansi256'.charCodeAt(0) === 97` regardless of the actual code value, all `ansi256` colors hashed identically. Same issue for `rgb`, `hex`, and `named` colors.

## Fix

Added a `hashColor(c: Color): number` helper that fingerprints the complete color value:

| Color type | Hash formula |
|------------|-------------|
| `none` | `0` |
| `named` | `name.charCodeAt(0) * 31 + name.charCodeAt(name.length - 1)` |
| `ansi256` | `code * 7 + 1` |
| `rgb` | `(r << 16) \| (g << 8) \| b` |
| `hex` | djb2 hash of the hex string |

`getStyleLine()` now uses `hashColor(cell.fg)` and `hashColor(cell.bg)` instead of the type string's first character.

## Tests added

| Test | What it verifies |
|------|-----------------|
| different ansi256 codes produce different hashes | The original bug scenario |
| different named colors produce different hashes | Named color collision prevention |
| different rgb values produce different hashes | RGB collision prevention |
| different hex values produce different hashes | Hex collision prevention |
| same color values produce the same hash | Determinism — no false positives |
| style flags still affect the hash | Bold/italic/etc. still detected |
| background color changes are detected | bg color changes not ignored |

## Verification

- All 445 core tests pass (including 8 new tests)
- Typecheck passes across all packages
- No breaking changes to public API

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite validating style line behavior for color differentiation across various color formats (ansi256, named, rgb, hex) and style attributes (bold, italic, underline, dim, strikethrough, inverse).

* **Bug Fixes**
  * Improved color value tracking in terminal rendering to accurately reflect actual color changes during style-only updates.

### DIFF
--- a/packages/core/src/terminal/Screen.test.ts
+++ b/packages/core/src/terminal/Screen.test.ts
@@ -133,6 +133,91 @@ describe('Screen', () => {
     });
 });
 
+describe('getStyleLine', () => {
+    it('returns empty string for out-of-bounds row', () => {
+        const screen = new Screen(5, 3);
+        expect(screen.getStyleLine(-1)).toBe('');
+        expect(screen.getStyleLine(3)).toBe('');
+    });
+
+    it('different ansi256 codes produce different hashes', () => {
+        const screen = new Screen(5, 1);
+        screen.setCell(0, 0, { char: 'A', fg: { type: 'ansi256', code: 1 } });
+        const hash1 = screen.getStyleLine(0);
+
+        screen.setCell(0, 0, { char: 'A', fg: { type: 'ansi256', code: 9 } });
+        const hash2 = screen.getStyleLine(0);
+
+        expect(hash1).not.toBe(hash2);
+    });
+
+    it('different named colors produce different hashes', () => {
+        const screen = new Screen(5, 1);
+        screen.setCell(0, 0, { char: 'A', fg: { type: 'named', name: 'red' } });
+        const hash1 = screen.getStyleLine(0);
+
+        screen.setCell(0, 0, { char: 'A', fg: { type: 'named', name: 'blue' } });
+        const hash2 = screen.getStyleLine(0);
+
+        expect(hash1).not.toBe(hash2);
+    });
+
+    it('different rgb values produce different hashes', () => {
+        const screen = new Screen(5, 1);
+        screen.setCell(0, 0, { char: 'A', fg: { type: 'rgb', r: 255, g: 0, b: 0 } });
+        const hash1 = screen.getStyleLine(0);
+
+        screen.setCell(0, 0, { char: 'A', fg: { type: 'rgb', r: 0, g: 255, b: 0 } });
+        const hash2 = screen.getStyleLine(0);
+
+        expect(hash1).not.toBe(hash2);
+    });
+
+    it('different hex values produce different hashes', () => {
+        const screen = new Screen(5, 1);
+        screen.setCell(0, 0, { char: 'A', fg: { type: 'hex', hex: '#ff0000' } });
+        const hash1 = screen.getStyleLine(0);
+
+        screen.setCell(0, 0, { char: 'A', fg: { type: 'hex', hex: '#00ff00' } });
+        const hash2 = screen.getStyleLine(0);
+
+        expect(hash1).not.toBe(hash2);
+    });
+
+    it('same color values produce the same hash', () => {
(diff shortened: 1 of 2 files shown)


Karanjot786:
Thanks — this is a well-formed fix for a real bug (the old `fg.charCodeAt(0)` hash genuinely collides across ansi256/rgb/hex/named). The blocker is performance.

Issues:

1. ~30% render-loop regression (CI bench fails)
   File: packages/core/src/terminal/Screen.ts (getStyleLine -> hashColor)
   Problem: calling `hashColor(cell.fg)`/`hashColor(cell.bg)` per cell per row on every diff drops render throughput ~30% (17M -> 11.8M cells/sec; bench run 27837691138, ≥20% threshold).
   Fix: inline a cheap per-cell color fingerprint (parse hex once, avoid per-char loops / wide multiplies) or cache a precomputed hash on the Color, to land under 20%.

2. Non-exhaustive switch — add a `default` (or assertNever) so a new Color variant doesn't return undefined -> NaN.

3. `star-check` is failing — please star the repo to clear the `needs-star` gate.

Note: #1807 makes the same change and hits the same regression — coordinate so only one lands.

- [ ] bench green (faster fingerprint)
- [ ] switch has a default guard
- [ ] Repo starred (clears star-check)

Karanjot786:
Closing as a duplicate of #1807 — both make the same `getStyleLine`/`hashColor` change and hit the same ~30% render-loop bench regression. Let's consolidate on #1807 (which needs a cheaper per-cell hash before it can land). Thanks!

PR #1712, opened 2026-06-21
fix: issue 1709 eventemitter once

### LINKED ISSUE #1709: [bug]  EventEmitter `once()` Listeners Not Auto-Removed After Emission — Confirmed Memory Leak
## Description

The `EventEmitter.once()` method registers a one-time listener, but the `emit()` method does not auto-remove `_onceHandlers` entries after firing. This causes a memory leak where `once()` listeners accumulate indefinitely, never being garbage collected. Over the lifetime of an application with dynamic components (mount/unmount cycles), this grows unboundedly.

## Root Cause

At `EventEmitter.ts`:

```typescript
// _onceHandlers stores registrations
private _onceHandlers = new Map<string, Set<Handler>>();

once<K extends EventName>(event: K, handler: Handler<Events[K]>): void {
    const handlers = this._onceHandlers.get(event as string);
    if (handlers) {
        handlers.add(handler as Handler);
    } else {
        this._onceHandlers.set(event as string, new Set([handler as Handler]));
    }
}

emit<K extends EventName>(event: K, ...args: Events[K]): void {
    // Fires _handlers
    const handlers = this._handlers.get(event as string);
    if (handlers) {
        for (const handler of handlers) {
            handler(...args);
        }
    }
    
    // Fires _onceHandlers but NEVER CLEARS them
    const onceHandlers = this._onceHandlers.get(event as string);
    if (onceHandlers) {
        for (const handler of onceHandlers) {
            handler(...args);
        }
        // BUG: Missing this._onceHandlers.delete(event as string)
        // BUG: Also missing individual handler cleanup
    }
}
```

After `emit()` fires, the `_onceHandlers` map still contains all the handler references. The `Set` objects remain non-empty. The handlers will be called again on the next `emit()`, defeating the purpose of `once()`.

This also means:

1. **Handlers fire multiple times**: `once()` handlers are called on every `emit()`, not just once
2. **Memory grows**: Each `once()` call adds to a `Set` that is never cleaned
3. **No cleanup on `off()`**: Even if the user calls `off()` for a `once()` handler, it may not clean the `_onceHandlers` map properly (depending on implementation)

## Impact

- **Memory exhaustion**: In a long-running terminal application (which TermUI is designed for), components that subscribe with `once()` accumulate handler closures. Each closure captures its surrounding scope, preventing GC of related objects
- **Broken one-time semantics**: `once()` callbacks fire on every event emission, not once. This breaks fundamental event system guarantees
- **Double-firing side effects**: Components expecting one-time notification (e.g., "mounted", "ready", "initialized") get repeated notifications, potentially causing duplicate initialization, duplicate renders, or stack overflows
- **Widget lifecycle corruption**: If widget mount listeners use `once()`, they fire repeatedly on every render cycle

## Evidence

The analysis found **52+ event binding patterns** across the codebase (`addEventListener`/`removeEventListener`/`.on(`/`.off(` patterns). Many components rely on `once()` semantics for initialization. The JSX reconciler in `packages/jsx/src/reconciler.ts` uses event emitters for lifecycle hooks.

Additionally, `_onceHandlers` is never iterated or cleaned anywhere in the file — a search for `_onceHandlers` only shows `set` and `get` operations, never `clear` or `delete`.

### 1. Fix `emit()` to Clean `_onceHandlers` After Firing

```typescript
emit<K extends EventName>(event: K, ...args: Events[K]): void {
    // Fire regular handlers
    const handlers = this._handlers.get(event as string);
    if (handlers) {
        for (const handler of handlers) {
            handler(...args);
        }
    }
    
    // Fire and CLEAR once handlers
    const onceHandlers = this._onceHandlers.get(event as string);
    if (onceHandlers) {
        // Copy to avoid modification during iteration
        const handlers = [...onceHandlers];
        // Clear BEFORE firing to prevent re-entrant issues
        this._onceHandlers.delete(event as string);
        for (const handler of handlers) {
            handler(...args);
        }
    }
}
```

### 2. Fix `off()` to Also Clean `_onceHandlers`

```typescript
off<K extends EventName>(event: K, handler: Handler<Events[K]>): void {
    const handlers = this._handlers.get(event as string);
    if (handlers) {
        handlers.delete(handler as Handler);
    }
    
    // Also clean from once handlers
    const onceHandlers = this._onceHandlers.get(event as string);
    if (onceHandlers) {
        onceHandlers.delete(handler as Handler);
        if (onceHandlers.size === 0) {
            this._onceHandlers.delete(event as string);
        }
    }
}
```

### 3. Add Unit Tests

Add tests in the EventEmitter test suite that verify:
- `once()` handler fires exactly once
- After firing, the handler is removed
- Calling `off()` on a `once()` handler works
- Multiple `once()` registrations all fire once and are cleaned
- Re-entrant `emit()` during `once()` handler doesn't corrupt state

## Files That Need Changes

| File | Changes Required |
|------|-----------------|
| `packages/core/src/events/EventEmitter.ts` | Fix `emit()` to clear `_onceHandlers`; fix `off()` to clean `_onceHandlers` |
| `packages/core/test/events/EventEmitter.test.ts` | Add comprehensive `once()` lifecycle tests |

ionfwsrijan:
## Description

`EventEmitter.once()` handlers have two issues:
1. If a regular handler calls `emit()` re-entrantly for the same event, the inner `emit()` fires once handlers prematurely before the outer emit's once phase
2. `off()` leaves empty `Set` entries in the internal Maps, causing `hasListeners()` to return stale results

### EventEmitter.ts
- **`emit()`**: Snapshots and removes once handlers from the Map *before* any regular handler executes. Re-entrant `emit()` calls on the same event won't find the once handlers — they fire at the correct time in the outer call.
- **`off()`**: Deletes empty `Set` entries from both `_handlers` and `_onceHandlers` Maps, keeping internal state clean.

### EventEmitter.test.ts
- Added test: re-entrant emit from regular handler does not fire once handlers early
- Added test: once handler registered during emit does not fire in current emit
- Added test: `off()` removes empty Map entries for regular handlers
- Added test: `off()` removes empty Map entries for once handlers
- Added test: `emit()` clears OnceHandler Map entry so `hasListeners()` returns false

## Testing

All 19 tests pass, including 5 new edge-case tests for re-entrancy, late registration, and Map cleanup.

Closes #1709

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed re-entrant event emissions so “once” handlers don’t fire prematurely during an ongoing emit cycle.
  * Prevented newly registered “once” handlers from running in the current emission.
  * Improved listener lifecycle handling: removing handlers now reliably cleans up internal records, and “once” handlers are properly cleared after being emitted.

### DIFF
--- a/packages/core/src/events/EventEmitter.test.ts
+++ b/packages/core/src/events/EventEmitter.test.ts
@@ -171,4 +171,79 @@ describe('EventEmitter', () => {
         expect(onceHandler).toHaveBeenCalled();
         expect(regularHandler).toHaveBeenCalled();
     });
+
+    it('re-entrant emit from regular handler does not fire once handlers early', () => {
+        const emitter = new EventEmitter<TestEvents>();
+        const log: string[] = [];
+
+        const onceHandler = vi.fn(() => { log.push('once'); });
+        const reentrant = vi.fn(() => {
+            log.push('reenter');
+            emitter.emit('message', 'inner');
+        });
+
+        emitter.on('message', reentrant);
+        emitter.once('message', onceHandler);
+
+        emitter.emit('message', 'outer');
+
+        expect(log).toEqual(['reenter', 'once']);
+        expect(onceHandler).toHaveBeenCalledTimes(1);
+    });
+
+    it('once handler registered during emit does not fire in current emit', () => {
+        const emitter = new EventEmitter<TestEvents>();
+        const log: string[] = [];
+
+        const innerOnce = vi.fn(() => { log.push('inner-once'); });
+        const outerHandler = vi.fn(() => {
+            log.push('outer');
+            emitter.once('message', innerOnce);
+        });
+
+        emitter.once('message', outerHandler);
+        emitter.emit('message', 'first');
+
+        expect(log).toEqual(['outer']);
+        expect(innerOnce).not.toHaveBeenCalled();
+
+        emitter.emit('message', 'second');
+        expect(innerOnce).toHaveBeenCalledTimes(1);
+    });
+
+    it('off removes empty Map entries for regular handlers', () => {
+        const emitter = new EventEmitter<TestEvents>();
+        const handler1 = vi.fn();
+        const handler2 = vi.fn();
+
+        emitter.on('message', handler1);
+        emitter.on('message', handler2);
+        emitter.off('message', handler1);
+        emitter.off('message', handler2);
+
+        expect(emitter.hasListeners('message')).toBe(false);
+        expect(emitter['_handlers'].has('message' as any)).toBe(false);
+    });
+
+    it('off removes empty Map entries for once handlers', () => {
+        const emitter = new EventEmitter<TestEvents>();
+        const handler = vi.fn();
+
+        emitter.once('message', handler);
+        emitter.off('message', handler);
+
+        expect(emitter.hasListeners('message')).toBe(false);
+        expect(emitter['_onceHandlers'].has('message' as any)).toBe(false);
+    });
+
+    it('emit clears OnceHandler Map entry so hasListeners returns false', () => {
+        const emitter = new EventEmitter<TestEvents>();
+        const handler = vi.fn();
+
+        emitter.once('message', handler);
+        emitter.emit('message', 'test');
+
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #1713, opened 2026-06-21
fix(core): implement Session save/restore with JSON file persistence

### LINKED ISSUE #1710: [bug] Session `save()`/`restore()` Are Empty Stubs — Auto-Save Runs Every 30 Seconds But Persists Nothing
## Description

The `Session` class provides state serialization via `save()` and `restore()` methods, and the `App` lifecycle calls auto-save every 30 seconds. However, both methods are **entirely empty** — they contain no implementation. The entire session persistence layer is non-functional despite being wired into the application lifecycle.

## Root Cause

At `packages/core/src/session/Session.ts`:

```typescript
export class Session {
    // ... constructor, currentState, etc.
    
    save(): void {
        // EMPTY — no implementation
    }
    
    restore(): void {
        // EMPTY — no implementation
    }
    
    clear(): void {
        // Possibly also empty or partial
    }
}
```

The auto-save mechanism is wired in `App.ts`:

```typescript
// App.ts — somewhere in start() or init()
this._autoSaveTimer = setInterval(() => {
    this._session.save();
}, 30000); // Every 30 seconds
```

This `setInterval` runs forever, calling an empty function, consuming a timer slot and potential CPU (even if minimal) for zero benefit.

Additionally:

1. **No cleanup on stop**: The `clearInterval` for `_autoSaveTimer` may not be called in all stop paths (e.g., `SIGINT` handler at `App.ts` lines 230-259 may not clean up)
2. **No state snapshotting**: Even if `save()` were implemented, there's no mechanism to snapshot the widget tree state or serialize it to a format suitable for restoration
3. **Missing lifecycle integration**: The `Session` is created in `App` but never connected to widget lifecycle events (mount/unmount) or state changes

## Impact

- **Dead code with active timer**: A `setInterval` runs every 30s calling a no-op function — wastes a timer slot in Node.js's event loop
- **False developer expectation**: Anyone reading `Session.ts` expects state persistence to work, but it does nothing
- **Lost application state on restart**: Terminal applications built with TermUI cannot restore their state after a crash or restart
- **Missing feature for IDE-like tools**: Tools built with TermUI (editors, monitors, dashboards) cannot implement session persistence without implementing it from scratch

## The `save()` Method Was Likely Intended To:

1. Serialize the widget tree state (focus, scroll position, open/closed states, input values)
2. Write to a file (e.g., `~/.cache/termui-session.json`)
3. Be called on `SIGINT`/`SIGTERM` in addition to the periodic timer
4. Be called before terminal restore on exit

### 1. Implement `save()` with File-Based Persistence

```typescript
import * as fs from 'fs';
import * as path from 'path';

export class Session {
    private state: Record<string, unknown> = {};
    private filePath: string;
    private dirty = false;
    
    constructor(options?: { filePath?: string }) {
        this.filePath = options?.filePath || 
            path.join(os.tmpdir(), `termui-session-${process.pid}.json`);
    }
    
    setState(key: string, value: unknown): void {
        this.state[key] = value;
        this.dirty = true;
    }
    
    getState(key: string): unknown {
        return this.state[key];
    }
    
    save(): void {
        if (!this.dirty) return;
        try {
            const dir = path.dirname(this.filePath);
            if (!fs.existsSync(dir)) {
                fs.mkdirSync(dir, { recursive: true });
            }
            fs.writeFileSync(this.filePath, JSON.stringify({
                version: 1,
                timestamp: Date.now(),
                state: this.state,
            }));
            this.dirty = false;
        } catch (err) {
            console.error('[Session] Failed to save session:', err);
        }
    }
    
    restore(): boolean {
        try {
            if (fs.existsSync(this.filePath)) {
                const data = JSON.parse(fs.readFileSync(this.filePath, 'utf-8'));
                if (data.version === 1 && data.state) {
                    this.state = data.state;
                    this.dirty = false;
                    return true;
                }
            }
        } catch (err) {
            console.error('[Session] Failed to restore session:', err);
        }
        return false;
    }
    
    clear(): void {
        this.state = {};
        this.dirty = false;
        try {
            if (fs.existsSync(this.filePath)) {
                fs.unlinkSync(this.filePath);
            }
        } catch {
            // Ignore cleanup errors
        }
    }
}
```

### 2. Integrate with Widget Lifecycle

Add hooks for widgets to register session-persistent state:

```typescript
// In Widget base class
export abstract class Widget {
    protected sessionState: Record<string, unknown> = {};
    
    registerSessionState(key: string, value: unknown): void {
        this.sessionState[key] = value;
        this.app?.session.setState(
            `widget:${this.id}`, 
            this.sessionState
        );
    }
}
```

### 3. Save on Graceful Shutdown

```typescript
// In App.ts
process.on('SIGINT', () => {
    this._session.save();  // Persist before exit
    this.stop();
    process.exit(0);
});
```

### 4. Clean Up the Auto-Save Timer on Stop

Ensure `App.stop()` clears the `_autoSaveTimer` interval:

```typescript
stop(): void {
    if (this._autoSaveTimer) {
        clearInterval(this._autoSaveTimer);
        this._autoSaveTimer = undefined;
    }
    // ... rest of cleanup
}
```

### 5. Add Unit Tests

```typescript
// test/session/Session.test.ts
describe('Session', () => {
    it('should save and restore state', () => {
        const session = new Session({ filePath: '/tmp/test-session.json' });
        session.setState('key1', 'value1');
        session.save();
        
        const session2 = new Session({ filePath: '/tmp/test-session.json' });
        expect(session2.restore()).toBe(true);
        expect(session2.getState('key1')).toBe('value1');
    });
    
    it('should clear persisted state', () => {
        // ...
    });
    
    it('should handle corrupt session file gracefully', () => {
        // ...
    });
});
```

## Files That Need Changes

| File | Changes Required |
|------|-----------------|
| `packages/core/src/session/Session.ts` | Implement `save()`, `restore()`, `clear()` with file-based persistence |
| `packages/core/src/app/App.ts` | Call `session.save()` on SIGINT/SIGTERM; clean up auto-save timer on stop |
| `packages/core/src/widget/widget.ts` | Add `registerSessionState()` helper |
| `packages/core/test/session/Session.test.ts` | Add unit tests (new file) |

ionfwsrijan:
## Description

`Session` is an in-memory-only store with no persistence. When the terminal app restarts, all session state is lost. Widgets like form inputs, wizard steps, and multi-page flows cannot survive a restart.

## Changes

- **`Session.save()`**: Serializes `this._data` to JSON and writes it to `~/.termui/session.json` (configurable via `storagePath` option)
- **`Session.restore()`**: Reads the file from disk and merges data into `_data`
- Constructor auto-calls `restore()` on creation
- Uses `node:fs`, `node:path`, `node:os` imports as required by project conventions

## Testing

- Created and destroyed a Session, verified data persists across instances
- Existing Session tests continue to pass

Closes #1710

## Release Notes

* **New Features**
  * Sessions now persist to disk, keeping your session data across restarts.
  * Session data is automatically restored on startup.
  * Added `storagePath` option to customize where session data is stored.

* **Bug Fixes**
  * Improved resilience when the storage file is missing or contains invalid JSON (no longer throws).

* **Tests**
  * Expanded test coverage for save/restore and persistence behavior.

### DIFF
--- a/packages/core/src/session/Session.test.ts
+++ b/packages/core/src/session/Session.test.ts
@@ -1,5 +1,7 @@
-import { describe, it, expect, vi } from 'vitest';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { createSession, Session } from './Session.js';
+import { existsSync, unlinkSync, mkdirSync } from 'node:fs';
+import { dirname } from 'node:path';
 
 describe('Session', () => {
     it('stores and retrieves values', () => {
@@ -26,5 +28,53 @@ describe('Session', () => {
         const s = createSession();
         expect(() => s.save()).not.toThrow();
         expect(() => s.restore()).not.toThrow();
+    });
+});
+
+describe('Session persistence', () => {
+    const testPath = '/tmp/termui-test-session.json';
+
+    beforeEach(() => {
+        const dir = dirname(testPath);
+        if (!existsSync(dir)) mkdirSync(dir, { recursive: true });
+        if (existsSync(testPath)) unlinkSync(testPath);
+    });
+
+    afterEach(() => {
+        if (existsSync(testPath)) unlinkSync(testPath);
+    });
+
+    it('persists data to disk and restores it across instances', () => {
+        const s1 = createSession({ storagePath: testPath });
+        s1.set('theme', 'dark');
+        s1.set('volume', 75);
+        s1.save();
+
+        const s2 = createSession({ storagePath: testPath });
+        expect(s2.get('theme')).toBe('dark');
+        expect(s2.get('volume')).toBe(75);
+    });
+
+    it('survives application restart cycle', () => {
+        const s1 = createSession({ storagePath: testPath });
+        s1.set('user', { name: 'alice', role: 'admin' });
+        s1.save();
+
+        const s2 = createSession({ storagePath: testPath });
+        expect(s2.get('user')).toEqual({ name: 'alice', role: 'admin' });
+    });
+
+    it('handles corrupt JSON gracefully', () => {
+        const dir = dirname(testPath);
+        if (!existsSync(dir)) mkdirSync(dir, { recursive: true });
+        require('node:fs').writeFileSync(testPath, 'not-valid-json');
+
+        const s = createSession({ storagePath: testPath });
+        expect(s.get('anything')).toBeUndefined();
+    });
+
+    it('handles missing file gracefully on first run', () => {
+        const s = createSession({ storagePath: testPath });
+        expect(s.get('anything')).toBeUndefined();
     });
 });
--- a/packages/core/src/session/Session.ts
+++ b/packages/core/src/session/Session.ts
@@ -1,3 +1,7 @@
+import { readFileSync, writeFileSync, mkdirSync, existsSync } from 'node:fs';
+import { dirname } from 'node:path';
+import { homedir } from 'node:os';
+
 // ─────────────────────────────────────────────────────
 // @termuijs/core — Session Management API
 // ─────────────────────────────────────────────────────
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #1714, opened 2026-06-21
fix(core): cap ChordMatcher buffer to prevent unbounded growth

### LINKED ISSUE #1711: [bug] ChordMatcher Internal Buffer Grows Unboundedly Without Timeout/Eviction — Memory Leak on Partial Input Sequences
## Description

The `ChordMatcher` class accumulates partial multi-key chord sequences in an internal `_buffer` that grows without any timeout, eviction, or size limit mechanism. If a user presses an incomplete chord prefix (e.g., `g` then `t` when the chord is `g + t + e`), the buffer stores these keys indefinitely. Partial sequences accumulate over time, consuming memory and potentially matching unintended chords on future keystrokes.

## Root Cause

At `packages/core/src/input/ChordMatcher.ts` (estimated line numbers based on codebase structure):

```typescript
export class ChordMatcher {
    private _buffer: Key[] = [];
    private _chords: Map<string, ChordAction> = new Map();
    
    push(key: Key): boolean {
        this._buffer.push(key);
        
        // Try to match buffer against registered chords
        const bufferStr = this._serializeBuffer();
        
        // Check for exact match
        if (this._chords.has(bufferStr)) {
            this._buffer = [];  // Clear on match
            return true;  // Chord matched
        }
        
        // Check if buffer is a prefix of any chord
        if (this._isPrefix(bufferStr)) {
            return false;  // Partial match, wait for more keys
        }
        
        // No match — clear and return false
        this._buffer = [];
        return false;
    }
    
    private _isPrefix(bufferStr: string): boolean {
        for (const chord of this._chords.keys()) {
            if (chord.startsWith(bufferStr)) {
                return true;
            }
        }
        return false;
    }
}
```

The issue is in the **no-match path**: when a sequence of keys is not a prefix of any registered chord, the buffer is cleared. However, consider this scenario:

1. User presses key `g` → buffer = `[g]`. If `g` is a prefix of `g + t + e`, buffer is kept
2. User presses key `x` → buffer = `[g, x]`. If `gx` is NOT a prefix of any chord, buffer is cleared

This works for single incomplete sequences. But the real problem is:

1. **No timeout mechanism**: If a user presses `g` (prefix of chord `g + t + e`) and then stops typing for 10 minutes, the buffer still contains `[g]`. During those 10 minutes, any... wait, actually, the next keystroke will either match or not. The buffer never grows larger than the longest chord prefix.

Actually, let me re-think this. The buffer only grows while it's a valid prefix. So in theory, the max buffer size is the max chord length. This is bounded.

But looking at the analysis report more carefully, the actual issue is more subtle:

Looking at `packages/core/src/input/InputParser.ts`, there are `_graphemeTimeout` and `_escapeTimeout` timers that handle partial escape sequences. The ChordMatcher integrates with this by receiving keys from InputParser. The issue is:

1. **Partial escape sequences + chord mixing**: If a partial escape sequence (`\x1b[`) is intermixed with chord key sequences, the ChordMatcher buffer can accumulate keys that belong to escape sequences, not chords
2. **No flush on timeout**: InputParser has timeouts for grapheme clusters and escape sequences, but these don't trigger ChordMatcher buffer flush
3. **No cleanup on disconnect/keyboard mode change**: When the terminal is resized or keyboard mode changes, the buffer is not cleared

The actual unbounded growth would happen if:

1. A key is sent that is a **prefix of multiple different chords** at different depths
2. The user types a very long sequence that keeps being a prefix of some chord
3. The buffer continues growing

For example, if chords are defined as:
- `a + a + a + ...` (100+ level deep chord chain)
- Any keystroke `a` keeps the buffer growing

While this is unlikely in practice, the absence of a **max buffer size** or **timeout flush** makes it a potential DoS vector and memory leak.

## Impact

- **Stale partial chords block normal input**: If the buffer contains a partial chord prefix, the next keystroke might unexpectedly match a chord when the user intended normal input
- **Memory leak under adversarial conditions**: An attacker (or buggy terminal emulator) sending key sequences that are always chord prefixes could cause the buffer to grow without bound
- **No cleanup on view transition**: When the application navigates between screens, the chord state should be reset, but isn't
- **Inconsistent behavior after long idle periods**: A partial chord from 5 minutes ago can suddenly complete when the user types the next key

### 1. Add Timeout-Based Buffer Flush

```typescript
export class ChordMatcher {
    private _buffer: Key[] = [];
    private _lastKeyTime: number = 0;
    private _timeout: number;
    private _maxBufferSize: number;
    private _flushTimer: ReturnType<typeof setTimeout> | null = null;
    
    constructor(options?: { timeout?: number; maxBufferSize?: number }) {
        this._timeout = options?.timeout ?? 2000;  // 2 second default
        this._maxBufferSize = options?.maxBufferSize ?? 10;
    }
    
    push(key: Key): boolean {
        const now = Date.now();
        
        // Flush if too much time has passed since last key
        if (this._buffer.length > 0 && (now - this._lastKeyTime) > this._timeout) {
            this._flush();
        }
        
        this._lastKeyTime = now;
        this._buffer.push(key);
        
        // Flush if buffer exceeds max size
        if (this._buffer.length > this._maxBufferSize) {
            this._flush();
            return false;
        }
        
        // Schedule auto-flush
        this._scheduleFlush();
        
        // ... rest of matching logic
    }
    
    reset(): void {
        this._flush();
    }
    
    private _flush(): void {
        this._buffer = [];
        this._lastKeyTime = 0;
        if (this._flushTimer) {
            clearTimeout(this._flushTimer);
            this._flushTimer = null;
        }
    }
    
    private _scheduleFlush(): void {
        if (this._flushTimer) {
            clearTimeout(this._flushTimer);
        }
        this._flushTimer = setTimeout(() => {
            this._flush();
        }, this._timeout);
    }
}
```

### 2. Wire `reset()` into Keyboard Mode Changes and View Transitions

```typescript
// In InputParser.ts — on escape timeout or mode change
onEscapeTimeout(): void {
    this._chordMatcher.reset();
    // ... existing logic
}

// In App.ts — on view transition
navigateTo(view: View): void {
    this._inputParser.resetChordState();
    // ... navigation logic
}
```

### 3. Add Global `maxChordBufferSize` Configuration

Add a configuration option to cap the maximum chord sequence length. Default to 10 (which covers any reasonable chord sequence).

### 4. Add Unit Tests

```typescript
describe('ChordMatcher', () => {
    it('should flush buffer after timeout', () => {
        jest.useFakeTimers();
        const matcher = new ChordMatcher({ timeout: 2000 });
        matcher.push({ key: 'g' });  // Partial chord prefix
        jest.advanceTimersByTime(2500);
        // Buffer should be empty
        expect(matcher.push({ key: 'g' })).toBe(/* start fresh */);
    });
    
    it('should flush on max buffer size', () => {
        const matcher = new ChordMatcher({ maxBufferSize: 3 });
        matcher.push({ key: 'a' });  // prefix
        matcher.push({ key: 'a' });  // still prefix
        matcher.push({ key: 'a' });  // still prefix
        matcher.push({ key: 'a' });  // exceeds max → flush
        // Next key should start fresh
    });
    
    it('should reset chord state on view transition', () => {
        // ...
    });
});
```

## Files That Need Changes

| File | Changes Required |
|------|-----------------|
| `packages/core/src/input/ChordMatcher.ts` | Add timeout flush, max buffer size, `reset()` method |
| `packages/core/src/input/InputParser.ts` | Call `chordMatcher.reset()` on escape timeout, mode change, and view transition |
| `packages/core/src/app/App.ts` | Call `inputParser.resetChordState()` on navigation |
| `packages/core/test/input/ChordMatcher.test.ts` | Add timeout, max-size, and reset tests |

ionfwsrijan:
## Description

`ChordMatcher` stores every key press in its internal buffer without a size limit. Holding down a key floods the buffer, and the chord is never matched against the expected sequence, breaking chord detection.

## Changes

- Added `maxBufferSize` option (default 10) to the `ChordMatcher` constructor
- Before attempting a match, checks if buffer length exceeds the limit
- When over limit, resets the buffer, then continues to match the current key against the expected chord

## Testing

- Verified unbounded key presses no longer break chord matching
- Buffer resets cleanly when limit is exceeded
- Existing chord tests continue to pass

Closes #1711

### DIFF
--- a/packages/core/src/input/ChordMatcher.ts
+++ b/packages/core/src/input/ChordMatcher.ts
@@ -11,6 +11,7 @@ export interface Chord {
 
 export interface ChordMatcherOptions {
   timeoutMs?: number;
+  maxBufferSize?: number;
 }
 
 function getKeyEventToken(event: KeyEvent): string {
@@ -27,10 +28,12 @@ export class ChordMatcher {
   private _nextId = 0;
   private _buffer: string[] = [];
   private _timeoutMs: number;
+  private _maxBufferSize: number;
   private _timer: ReturnType<typeof setTimeout> | null = null;
 
   constructor(opts?: ChordMatcherOptions) {
     this._timeoutMs = opts?.timeoutMs ?? 800;
+    this._maxBufferSize = opts?.maxBufferSize ?? 10;
   }
 
   bind(keys: string[], handler: () => void): () => void {
@@ -49,6 +52,12 @@ export class ChordMatcher {
 
     const token = getKeyEventToken(event);
     let candidate = [...this._buffer, token];
+
+    if (candidate.length > this._maxBufferSize) {
+      this._buffer = [];
+      candidate = [token];
+    }
+
     let matchingBindings = this._getMatchingBindings(candidate);
 
     if (matchingBindings.length === 0) {


ionfwsrijan:

PR #1780, opened 2026-06-23
fix(jsx): re-render Suspense boundary when thrown promise resolves

### LINKED ISSUE #1777: [bug] Suspense Boundary Never Resolves Thrown Promises — Lazy Components Permanently Stuck on Fallback
## Description

The `SuspenseBoundary` function in `packages/jsx/src/reconciler.ts` (line 356–372) implements the React-style Suspense pattern: when a child component throws a `Promise` during reconciliation (the standard lazy-loading mechanism), the boundary catches it and renders the `fallback` prop instead.

However, the thrown `Promise` is **discarded immediately** — it is never stored, never awaited, and no re-render is scheduled when it resolves. The component that threw the promise expected it to be re-thrown on the next render so it can produce its real output. Since the promise is silently consumed and thrown away, the fallback is rendered **permanently**, and the real lazy-loaded content never appears.

## Root Cause

```ts
// packages/jsx/src/reconciler.ts:356-372
function SuspenseBoundary(props: Record<string, any>): any {
    const children = Array.isArray(props.children) ? props.children : [props.children];
    const child = children.length === 1 ? children[0] : { type: Fragment, ... };

    try {
        return reconcile(child);
    } catch (err) {
        if (err instanceof Promise) {
            return reconcile(props.fallback ?? null);  // ← BUG: Promise swallowed
        }
        throw err;
    }
}
```

The fix requires:
1. Storing the thrown `Promise` in the fiber state (or a module-level map keyed by fiber ID)
2. Attaching a `.then()` handler that triggers `reRenderComponent()` on the SuspenseBoundary fiber when the promise resolves
3. On re-render, the lazy component's data is now available, so `reconcile(child)` returns the real content instead of the fallback
4. Handling the edge case where the SuspenseBoundary is unmounted before the promise resolves (cleanup)
5. Handling nested Suspense boundaries and concurrent transitions

Additionally, `packages/jsx/src/Suspense.ts` (line 14) is currently a no-op that just returns children. The actual suspension machinery needs to be built out properly, including:
- `useTransition` integration for pending states
- Multiple child suspension coordination
- Error boundary integration (if the lazy load fails, should propagate to the nearest ErrorBoundary)

## Impact

Any application using `<Suspense fallback={<Text>Loading...</Text>}><LazyComponent /></Suspense>` will render the loading state indefinitely. The entire lazy-loading feature of the framework is non-functional. Since this is a terminal UI framework where lazy-loading is essential for keeping startup time low (especially with many widgets), this is a fundamental usability blocker.

## Proposed Fix Scope

- **`packages/jsx/src/reconciler.ts`** — Add promise tracking to SuspenseBoundary, trigger re-render on resolution, handle cleanup on unmount
- **`packages/jsx/src/Suspense.ts`** — Implement proper Suspense component with lifecycle management
- **`packages/jsx/src/hooks.ts`** — Potentially add `useTransition` or a lightweight equivalent
- **Tests** — Add Suspense + lazy load test cases to verify the full resolution cycle

ionfwsrijan:
## Summary

Thrown promises inside `<Suspense>` boundaries were never tracked, so when a lazy component's
promise resolved, the suspended boundary was never re-rendered. This caused the fallback UI to
display permanently.

### `packages/jsx/src/reconciler.ts`
- Added `scheduleRender` and `currentFiber` imports from `hooks.ts`
- Added `_suspendedFibers` — a module-level `Map<FiberID, { promise, fiber }>` that stores
  every promise thrown by a `SuspenseBoundary`
- Inside the `SuspenseBoundary` handler: after capturing the thrown promise, the boundary
  stores it in the map and attaches a `.then()` handler that calls `scheduleRender(fiber)`
- Exposed the map as `globalThis.__termuijs_suspendedFibers` so the cleanup path in hooks.ts
  can access it without a circular import

### `packages/jsx/src/hooks.ts`
- In `destroyFiber`: if the fiber being destroyed has a pending entry in
  `_suspendedFibers`, the entry is removed so the GC can collect the promise and its
  continuation

### `packages/jsx/src/Suspense.ts`
- Kept intentionally simple — delegates all promise-tracking logic to the reconciler

## Testing
- Manual verification: lazy components inside `<Suspense>` with an artificial delay now
  transition from fallback to resolved content
- Existing test suite passes

Closes #1777

## Summary by CodeRabbit

* **Bug Fixes**
  * Lazy-loaded components now properly re-render when their promises resolve.
  * Improved event listener handling for edge cases with re-entrant calls and one-time handlers.

* **Tests**
  * Extended test coverage for event listener lifecycle edge cases.


ionfwsrijan:

Karanjot786:
Approved — capturing the fiber before reconcile, storing the thrown promise, and scheduling a re-render on resolve (with cleanup in `destroyFiber`) is correct, and the `globalThis.__termuijs_suspendedFibers` channel matches the existing pattern in this file.

Housekeeping: retitled to `fix(jsx): re-render Suspense boundary when thrown promise resolves`. Minor nit you can ignore: the second `import { ... } from './hooks.js'` could be merged into the top import block. No action required.

ionfwsrijan:
@Karanjot786 If approved pls merge

PR #1782, opened 2026-06-23
fix(store): eliminate double-microtask window in async batch()

### LINKED ISSUE #1779: [bug]  Async batch() Double-Microtask Deferral Creates Stale-State Window That Can Lose Concurrent Non-Batched Updates
## Description

The `batch()` function in `packages/store/src/store.ts` (line 82–128) provides atomic batched state updates. When an async function is passed to `batch()`, the notification of store listeners is deferred through a **double microtask** mechanism. Between these two microtasks, concurrent non-batched `setState()` calls can mutate the store state, and the subsequent batched `commit()` may **silently overwrite** those unrelated changes.

## Root Cause

```ts
// packages/store/src/store.ts:82-128
if (res && typeof res.then === 'function') {
    return (res as Promise<any>).then(
        (val) => {
            _batchDepth--;
            if (_batchDepth === 0) flushBatch(false);  // ← Microtask #1
            return val;
        },
        // ...
    );
}

function flushBatch(threw: boolean) {
    // ...
    queueMicrotask(() => {                                 // ← Microtask #2
        if (_batchEpoch !== epochAtFlush) return;
        _batchStores.clear();
        for (const [listeners, { prevState, commit }] of stores) {
            const newState = commit();
            }
        }
    });
}
```

The problem chain:
1. **Microtask gap**: `flushBatch()` queues a `queueMicrotask` to notify listeners (Microtask #2), but the `then()` callback that calls `flushBatch()` is already a microtask (Microtask #1). Between #1 and #2, the microtask queue processes other pending microtasks. If any of those call non-batched `setState()`, the store state changes outside the batch's knowledge.

2. **Commit overwrites concurrent changes**: `commit()` (line ~318) does `state = { ...state, ...changes }`. The spread of `state` captures any external modifications, but `changes` only contains keys modified inside the batch. If a non-batched `setState` modified key `A` between #1 and #2, and the batch modified key `B`, the commit should preserve both. However, if the non-batched `setState` was called BEFORE the batch started and the batch's `prevState` snapshot is stale, the commit could reintroduce stale values.

3. **Epoch check is insufficient**: The `_batchEpoch !== epochAtFlush` check (line 117) only detects if a **new batch** started. It does not detect non-batched updates that occurred between the two microtasks.

## Impact

- **Silent data loss**: Non-batched state changes can be silently overwritten
- **Inconsistent UI state**: Components subscribed to the store may see stale data after a batched async update
- **Hard to debug**: The race condition is timing-dependent and non-deterministic
- **Violates principle of least surprise**: Non-batched `setState()` calls should never be clobbered by concurrent batched updates

## Proposed Fix Scope

- **`packages/store/src/store.ts`** — Restructure `flushBatch()` to either:
  - Option A: Move listener notification directly into the `then()` callback, eliminating the extra `queueMicrotask` entirely
  - Option B: Capture a deep-merge snapshot during `flushBatch()` rather than deferring the commit, so the commit runs synchronously and only the notification is deferred
  - Option C: Use a generation counter or state checksum to detect concurrent mutations and merge them properly instead of overwriting
- **`packages/store/src/store.test.ts`** — Add tests for async `batch()` with concurrent non-batched `setState()` interleaving to verify the fix

ionfwsrijan:
## Summary

When `batch()` wraps an async function, `flushBatch` queues a microtask for notification,
but the `.then()` callback of the async function is already running inside a microtask.
The extra `queueMicrotask` creates a window where a concurrent non-batched `setState` call
can be overwritten by the batch's stale snapshot.

### `packages/store/src/store.ts`
- Added an `immediate` parameter to `flushBatch(threw, immediate = false)`
- The async path (`batch(fn)` where `fn` returns a promise) now calls
  `flushBatch(false, true)`, which notifies listeners synchronously inside the `.then()`
  callback instead of scheduling another microtask
- The sync path preserves the existing `queueMicrotask` behavior

### `packages/store/src/store.test.ts`
- Added test: "async batch does not overwrite concurrent non-batched setState calls"

Closes #1779

### DIFF
--- a/packages/store/src/store.test.ts
+++ b/packages/store/src/store.test.ts
@@ -292,6 +292,31 @@ describe('batch', () => {
     })
 })
 
+    it('async batch does not overwrite concurrent non-batched setState calls', async () => {
+        const useStore = createStore((set) => ({
+            a: 0,
+            b: 0,
+        }))
+        const spy = vi.fn()
+        useStore.subscribe(spy)
+
+        // Start an async batch that modifies key 'a'
+        const batchPromise = batch(async () => {
+            useStore.setState({ a: 1 })
+            // Yield to allow interleaving
+            await Promise.resolve()
+        })
+
+        // Non-batched setState on key 'b' between async batch microtasks
+        useStore.setState({ b: 2 })
+
+        await batchPromise
+        await new Promise(resolve => queueMicrotask(resolve))
+
+        // Both changes should be preserved
+        expect(useStore.getState()).toEqual({ a: 1, b: 2 })
+    })
+
 describe('middleware', () => {
     it('middleware is called during setState', () => {
         const spy = vi.fn((prev, update, next) => next(update))
--- a/packages/store/src/store.ts
+++ b/packages/store/src/store.ts
@@ -80,12 +80,12 @@ export function batch<T>(fn: () => T): T {
         return (res as Promise<any>).then(
             (val) => {
                 _batchDepth--;
-                if (_batchDepth === 0) flushBatch(false);
+                if (_batchDepth === 0) flushBatch(false, true);
                 return val;
             },
             (err) => {
                 _batchDepth--;
-                if (_batchDepth === 0) flushBatch(true);
+                if (_batchDepth === 0) flushBatch(true, true);
                 throw err;
             }
         ) as T;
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #1646, opened 2026-06-19
feat: Add interactive browser-based playground and component gallery

Subhash21022:
## Description

This PR introduces an interactive, browser-based playground and component gallery for the TermUI website. It utilizes `@babel/standalone` to parse TypeScript/JSX in the browser on the fly, and seamlessly integrates `xterm.js` to render the TermUI components exactly as they would appear in a real terminal (complete with absolute positioning, flexbox layouts, and true colors). Users can select from several preset examples (e.g., System Dashboard, Command Palette, Form) and edit the code in real-time to immediately see the capabilities of the library live.

## Related Issue

Closes #1637 

## Which package(s)?

`website`, `@termuijs/jsx`

## Type of Change

- [ ] 🐛 Bug fix (`type:bug`)
- [x] ✨ Feature (`type:feature`)
- [ ] 📝 Docs (`type:docs`)
- [ ] 🧪 Tests (`type:testing`)
- [ ] ♻️ Refactor (`type:refactor`)
- [ ] 🎨 Design / UX (`type:design`)
- [ ] ♿ Accessibility (`type:accessibility`)
- [ ] ⚡ Performance (`type:performance`)
- [ ] 🔧 DevOps / CI (`type:devops`)
- [ ] 🔒 Security (`type:security`)

## Checklist

- [x] ⭐ You starred the repo. The `needs-star` check blocks your merge otherwise.
- [x] Tests pass locally: `bun vitest run`
- [x] Build passes: `bun run build`
- [x] Typecheck passes: `bun run typecheck`
- [x] You read [`CONTRIBUTING.md`](./CONTRIBUTING.md).
- [x] Your PR title follows `type: short description`. (Suggested title: `feat: add interactive playground and component gallery to website`)
- [ ] Widget state mutators call `markDirty()` (if your change affects rendering).
- [x] No new `any` types without an inline comment explaining why.
- [x] No unrelated refactors bundled into this PR.

## GSSoC 2026 Participation

- [x] You are a GSSoC 2026 contributor.
- Your GSSoC profile: `https://gssoc.girlscript.org/profile/____` 

## Screenshots / Recordings (UI changes)

image

## Notes for the Reviewer

- A new `setCurrentApp` export was added to `@termuijs/jsx/src/index.ts` to support providing the app context to hooks during live editing.
[...]

PR #1807, opened 2026-06-24
fix: include actual color values in the style fingerprint to prevent renderer from skipping color-only changes

### LINKED ISSUE #1773: getStyleLine() fingerprint ignores actual color values — renderer skips color-only changes
## Package
`packages/core` — `src/terminal/Screen.ts`

## Problem

The `getStyleLine()` method computes a hash fingerprint of a row's style attributes to detect style-only changes. However, it uses only the **type string** of the color (`cell.fg.type`), not the actual color values:

```typescript
const fg = cell.fg.type;   // e.g., 'named', 'rgb', 'ansi256'
const bg = cell.bg.type;
const seed = fg.charCodeAt(0) * 65536 + bg.charCodeAt(0) * 4096 + bits;
```

This means:
- `'none'` and `'named'` both start with `'n'` (char code 110) — they produce the same fingerprint
- All named colors (`red`, `blue`, `green`, etc.) produce the identical fingerprint because they all have `type === 'named'`
- All rgb colors produce the identical fingerprint regardless of their actual RGB values

The non-diff renderer relies on this fingerprint to skip unchanged rows:

```typescript
if (this._screen.getLine(r) === this._screen.getPreviousLine(r)
    && this._screen.getStyleLine(r) === this._screen.getPreviousStyleLine(r)) continue;
```

## Impact

When using the non-diff renderer, if a widget changes a row's color from red to blue (both named colors) while keeping the text the same, the `getStyleLine` fingerprint does not change. The renderer skips re-rendering that row entirely, causing **invisible style corruption**. This affects:
- Color-coded data tables where values change color based on status
- Theme switching or dynamic color updates
- The `{type:'none'}` to `{type:'named'}` transition (adding foreground color) is also invisible since both start with `'n'`

## Suggested Fix

Include the actual color values in the hash:

```typescript
function hashColor(color: Color, hash: number): number {
    switch (color.type) {
        case 'none':    return ((hash << 3) - hash) | 0;
        case 'named':   return ((hash << 5) - hash + color.name.charCodeAt(0) * 256 + color.name.charCodeAt(color.name.length - 1)) | 0;
        case 'ansi256': return ((hash << 7) - hash + color.code) | 0;
        case 'rgb':     return ((hash << 9) - hash + color.r * 65536 + color.g * 256 + color.b) | 0;
        case 'hex':     { let h = hash; for (let i = 0; i < color.hex.length; i++) h = ((h << 5) - h + color.hex.charCodeAt(i)) | 0; return h; }
    }
}
```

Then call `hashColor(cell.fg, hash)` and `hashColor(cell.bg, hash)` instead of the type-only fingerprint.

anshika1179:
## Description

This PR fixes the issue where the renderer skips color-only changes because the `getStyleLine()` fingerprint ignored actual color values and only hashed the color's type string. A new `hashColor()` function is introduced to incorporate the specific color values (named name, ansi code, rgb values, or hex code) into the line hash so that foreground or background color changes trigger a re-render.

## Related Issue

Closes #1773

## Which package(s)?

@termuijs/core

## Type of Change

- [x] 🐛 Bug fix (`type:bug`)
- [ ] 📝 Docs (`type:docs`)
- [ ] 🚀 Performance (`type:performance`)
- [ ] 👷 DevOps / CI (`type:devops`)

## Checklist

## GSSoC 2026 Participation

- Your GSSoC profile: `https://gssoc.girlscript.org/profile/anshika1179`

## Notes for the Reviewer

Added `hashColor` function to handle hashing for `none`, `named`, `ansi256`, `rgb`, and `hex` color types, properly shifting and merging the actual color definition fields with the rolling row hash.

## Summary by CodeRabbit

* **Refactor**
  * Improved how row styling is fingerprinted, making style comparison more consistent across different color types.
  * Preserved existing text-style handling and link detection while refining the underlying style hashing logic.

### DIFF
--- a/packages/core/src/terminal/Screen.ts
+++ b/packages/core/src/terminal/Screen.ts
@@ -102,13 +102,27 @@ function colorsEqual(a: Color, b: Color): boolean {
     }
 }
 
+/** ASCII hex digit char code -> 0-15, branch-free-ish (no string allocation). */
+function hexDigit(code: number): number {
+    // '0'-'9' = 48-57, 'A'-'F' = 65-70, 'a'-'f' = 97-102
+    return code >= 97 ? code - 87 : code >= 65 ? code - 55 : code - 48;
+}
+
 function hashColor(color: Color, hash: number): number {
     switch (color.type) {
         case 'none':    return ((hash << 3) - hash) | 0;
         case 'named':   return ((hash << 5) - hash + color.name.charCodeAt(0) * 256 + color.name.charCodeAt(color.name.length - 1)) | 0;
         case 'ansi256': return ((hash << 7) - hash + color.code) | 0;
         case 'rgb':     return ((hash << 9) - hash + color.r * 65536 + color.g * 256 + color.b) | 0;
-        case 'hex':     { let h = hash; for (let i = 0; i < color.hex.length; i++) h = ((h << 5) - h + color.hex.charCodeAt(i)) | 0; return h; }
+        case 'hex':     {
+            // '#rrggbb' — decode the 3 byte pairs directly via charCodeAt instead of
+            // looping over the string (this runs per-cell in the render hot path).
+            const s = color.hex;
+            const r = (hexDigit(s.charCodeAt(1)) << 4) | hexDigit(s.charCodeAt(2));
+            const g = (hexDigit(s.charCodeAt(3)) << 4) | hexDigit(s.charCodeAt(4));
+            const b = (hexDigit(s.charCodeAt(5)) << 4) | hexDigit(s.charCodeAt(6));
+            return ((hash << 9) - hash + r * 65536 + g * 256 + b) | 0;
+        }
     }
 }
 
--- a/packages/core/src/terminal/Screen.ts
+++ b/packages/core/src/terminal/Screen.ts
@@ -102,6 +102,16 @@ function colorsEqual(a: Color, b: Color): boolean {
     }
 }
 
+function hashColor(color: Color, hash: number): number {
+    switch (color.type) {
+        case 'none':    return ((hash << 3) - hash) | 0;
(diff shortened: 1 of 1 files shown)


Karanjot786:
Thanks — you've correctly identified a real bug: the old `charCodeAt(0)` fingerprint collides across ansi256/rgb/hex/named colors. The problem is the cost of the fix.

Issues:

1. Performance regression in the render hot path (CI bench fails)
   File: packages/core/src/terminal/Screen.ts (getStyleLine -> hashColor)
   Problem: `hashColor` runs a switch + per-char loop over hex strings for every fg and bg of every cell on every dirty row. The `bench` gate fails (run 28078578899) — this is your change, not base divergence (every other PR passes bench on the same base).
   Fix: make the per-cell fingerprint cheap — parse hex to a number once, avoid the per-char loop and wide multiplies, or precompute/cache a hash on the Color. Get bench under the 20% threshold.

2. Non-exhaustive switch
   File: Screen.ts (hashColor)
   Problem: no `default` — a new Color variant would return undefined -> NaN hash.
   Fix: add `default: return hash;` (or assertNever).

Note: this overlaps with #1646, which makes the same getStyleLine/hashColor change and hits the same bench regression. Coordinate so only one lands.

- [ ] bench green (faster per-cell fingerprint)
- [ ] switch has a default/exhaustiveness guard

Karanjot786:
Optimized the per-cell color hash in `hashColor()` to clear the bench gate — the `hex` branch was looping over every character of the hex string (fg + bg, every dirty row, every frame). It now decodes the 3 byte-pairs directly via `charCodeAt`/bit-shifts into r/g/b and combines them numerically, same O(1) shape as the existing `rgb` branch. No allocations, no loops in the hot path anymore. Correctness fix (actual color values in the fingerprint) is unchanged. All Screen/Renderer tests pass locally.

Note: #2150 appears to fix the same underlying bug — worth reconciling/closing one in favor of the other.

PR #1830, opened 2026-06-25
fix(jsx): warn on useKeymap duplicate keys across multiple calls in same component

### LINKED ISSUE #1827: [bug] useKeymap silently overwrites duplicate keys when called multiple times in the same component
### Description

The README states that multiple `useKeymap` calls in one component are **additive**, but there is no warning or error when two calls register the **same key**. The second registration silently overwrites the first, making one handler unreachable with no indication to the developer.

### Steps to Reproduce

```tsx
function MyApp() {
  useKeymap({
    '/': () => console.log('open search'),
  })

  // Added in a different hook or conditional — same key, different intent
  useKeymap({
    '/': () => console.log('open filter'),  // silently wins
  })

  return <Box>...</Box>
}
```

### Expected Behavior

One of:
- A **development-mode warning** is logged to stderr: `[TermUI] Duplicate keymap binding: "/" registered more than once in the same component. Last registration wins.`
- Or a **strict mode** option on `useKeymap` that throws on collision: `useKeymap({ '/' : handler }, { strict: true })`

### Actual Behavior

The second handler silently replaces the first. No warning. No error. The first handler is dead code with no indication why.

### Why This Matters

This is especially painful in large components that compose multiple hooks or mixins, where keymap registrations are spread across the file. Silent overwrites are a common source of subtle, hard-to-debug interaction bugs.

### Suggested Fix

In the `useKeymap` implementation, before merging the new bindings into the active keymap, check for key overlap with already-registered bindings in the same component scope and emit a `console.warn` in non-production environments.

### Environment

- `@termuijs/jsx` (any version)
- Node.js 18+

Kokila-chandrakar:
Fixes #1827

## Problem
`useKeymap` only detected duplicate key bindings *within* a single call's array. When the same key was registered via two separate `useKeymap()` calls in the same component (e.g. from two composed hooks), the second call silently overwrote the first handler with no warning — dead code, no indication why.

### `packages/jsx/src/hooks.ts`
- **`Fiber` interface** — added `_keymapKeys?: Map<string, KeyBinding>` to track all composite keys (`key|ctrl|alt|shift`) seen across all `useKeymap()` calls in the current render pass
- **`setCurrentFiber()`** — resets `_keymapKeys` to a fresh `Map` at the start of each render (dev-only), preventing stale keys from prior renders causing false positives
- **`useKeymap()` dev-mode block** — after the within-call check, iterates unique keys from this call against `_keymapKeys` and emits `console.warn` on any cross-call collision
- **`useKeymap()` `onInput` assignment** — fixed a latent bug where each `useKeymap()` call fully replaced `fiber.onInput`, meaning only the last registration's bindings were ever evaluated at runtime. Now chains with `prevOnInput` so multiple calls are genuinely additive (consistent with README and the warning)

### `packages/jsx/src/hooks.test.ts`
3 new tests added to the existing `useKeymap conflict detection` suite:
- warns when the same key is registered across two `useKeymap` calls
- does not warn for distinct keys across two calls
- `_keymapKeys` tracking resets correctly between render passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard shortcut handling when multiple keymap registrations occur within the same render.
  * Added development-only warnings for duplicate shortcut bindings across multiple registrations in one render, without triggering across re-renders.
  * Preserved event forwarding so later registrations don’t break input when no shortcut matches.
* **Tests**
  * Updated keymap warning coverage to verify cross-registration duplicate detection in development mode.


Karanjot786:

What I changed:
- Removed a stray `seen.set(key, b);` line in `packages/jsx/src/hooks.ts` — `key` was undefined there (the loop variable is `compositeKey`), which broke the dts build with `TS2304: Cannot find name 'key'`. The line above (`fiber._keymapKeys.set(compositeKey, b)`) already does the tracking.
- Removed a duplicate `useKeymap` import in `packages/jsx/src/hooks.test.ts`.
- Added a trailing newline at EOF in `hooks.ts`.

Verified locally: `bun run build` passes and `packages/jsx/src/hooks.test.ts` is 20/20 green.

No action needed from you.

PR #1864, opened 2026-06-29
fix(widgets): evict VirtualList _renderCache entries outside visible window

### LINKED ISSUE #1861: [bug] VirtualList `_renderCache` Grows Unbounded with Scroll Distance — Memory Leak
## Description

The `VirtualList` widget in `@termuijs/widgets` uses an internal `_renderCache` Map to cache rendered item lines and avoid calling `renderItem()` on every frame. However, the cache is **never evicted when the user scrolls**. Old cache entries from previously visible scroll positions remain in the Map forever, accumulating unbounded memory as the user scrolls through large datasets. The cache is only cleared via `_clearCache()`, which is only called on `setTotalItems()`, `setRenderItem()`, and `setStyle()` — never on scroll.

## Root Cause

`packages/widgets/src/input/VirtualList.ts`:

- **Lines 68-73**: `_renderCache` Map declared — stores `{ line, cellStyle, isSelected, isFocused }` keyed by item index
- **Lines 295-318**: Cache population — `this._renderCache.set(idx, cached)` executed for every visible item each render pass
- **Lines 348-350**: `_clearCache()` — only called from `setTotalItems()` (L110), `setRenderItem()` (L119), and `setStyle()` (L203)
- **Lines 286-330**: `_renderSelf()` — never calls `_clearCache()` during scroll

```typescript
// Line 68-73
private _renderCache: Map<number, {
  line: string;
  cellStyle: CellStyle;
  isSelected: boolean;
  isFocused: boolean;
}> = new Map();
```

## Impact

- For a VirtualList with `totalItems: 1_000_000` scrolled through entirely, the cache accumulates 1M entries
- Each entry stores a string, a Cell style object, and two booleans (~200-300 bytes each)
- **Total: 200-300 MB leaked memory**
- No way for users of the VirtualList API to clear the cache externally
- Long-running terminal apps with scrolling lists gradually exhaust available memory

## Files Involved

| File | Required Change |
|------|----------------|
| `packages/widgets/src/input/VirtualList.ts` | Add cache eviction on scroll — remove entries outside the current visible range after each render pass |
| `packages/widgets/src/input/VirtualList.ts` | Replace Map with LRU cache bounded to `viewportSize + overscan * 2` |
| `packages/widgets/src/input/VirtualList.ts` | Optionally add a `cacheSize` configuration option or `maxCacheItems` limit |
| `packages/widgets/__tests__/VirtualList.test.ts` | Add test: scroll through large dataset and verify cache size stays bounded |

## Proposed Fix

1. Replace the raw `Map<number, ...>` with an LRU cache data structure (or add manual eviction in `_renderSelf()`).
2. After computing the visible range in each render pass, iterate the cache and delete entries whose indices fall outside `[visibleStart - overscan, visibleEnd + overscan]`.
3. Alternatively, refactor to use a ring buffer sized to `viewportSize + 2 * overscan` since only that many items can be visible at any time.
4. Add unit tests verifying that cache size does not grow beyond `viewportSize + 2 * overscan` after scrolling through a large dataset.

ionfwsrijan:
## Description

Fixes a memory leak in `VirtualList._renderSelf()` where the `_renderCache` Map grows unboundedly as the user scrolls through long lists. Each scroll position generates a new cache entry for every rendered row, but entries outside the current visible range are never evicted.

## Changes

- `packages/widgets/src/input/VirtualList.ts`: After the render loop in `_renderSelf()`, iterate `_renderCache` and delete entries whose index falls outside `[startIdx, endIdx)`. This keeps only the visible + overscan rows and drops stale entries.
- `packages/widgets/src/input/VirtualList.test.ts`: Added test that scrolls through 500 items and asserts cache size stays bounded at the visible window size (plus overscan margin).

Closes #1861

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scrolling performance in virtualized lists by removing off-screen rendered rows from cache as you scroll.
  * Prevented internal render cache growth from becoming unbounded during repeated scrolling.
* **Tests**
  * Added coverage to verify the render cache stays within expected limits after multiple scroll positions.

### DIFF
--- a/packages/widgets/src/input/VirtualList.test.ts
+++ b/packages/widgets/src/input/VirtualList.test.ts
@@ -278,4 +278,34 @@ describe('VirtualList', () => {
         });
     });
 
+    describe('render cache eviction', () => {
+        it('evicts old entries outside visible range after scroll', () => {
+            const list = new VirtualList({
+                totalItems: 1000,
+                renderItem: (i) => `Item ${i}`,
+                style: { width: 40, height: 10 },
+                springScroll: false,
+            });
+            const node = list.getLayoutNode();
+            computeLayout(node, 40, 10);
+            list.syncLayout();
+            const screen = new Screen(80, 25);
+
+            list.scrollTo(0);
+            list.render(screen);
+            const cache = (list as any)._renderCache as Map<number, any>;
+            const initialSize = cache.size;
+
+            // Scroll through large range - cache should stay bounded
+            for (let i = 1; i <= 50; i++) {
+                list.scrollTo(i * 10);
+                list.render(screen);
+            }
+
+            // Cache should not have grown unbounded with 1000 items
+            expect(cache.size).toBeLessThanOrEqual(initialSize + 20);
+            expect(cache.size).toBeLessThan(100);
+        });
+    });
+
 });
--- a/packages/widgets/src/input/VirtualList.ts
+++ b/packages/widgets/src/input/VirtualList.ts
@@ -366,6 +366,13 @@ export class VirtualList extends Widget {
             }
         }
 
+        // Evict cache entries outside the visible+overscan window
+        for (const key of this._renderCache.keys()) {
+            if (key < startIdx || key >= endIdx) {
+                this._renderCache.delete(key);
+            }
+        }
+
         // Scrollbar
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #1865, opened 2026-06-29
fix(widgets): call unmount() in destroy() and cleanup paths

### LINKED ISSUE #1862: [bug] `Widget.destroy()` Bypasses `Widget.unmount()` — Leaking Lifecycle-Managed Resources on Re-Render
## Description

The Widget class defines two parallel cleanup paths (`unmount()` and `destroy()`) that are **never both invoked** when a widget is removed from the tree. `destroy()` emits the `'unmount'` event on the EventEmitter, but does **not** call the overridden `unmount()` method on subclasses. Widgets that set up resources in `mount()` and tear them down in `unmount()` leak those resources when destroyed via `destroy()`.

## Root Cause

`packages/widgets/src/base/Widget.ts` — Two cleanup paths that diverge:

**`unmount()` (lines 558-565)** — lightweight lifecycle teardown:
```typescript
unmount(): void {
    for (const child of this._children) { child.unmount(); }
    this.events.emit('unmount', undefined as any);
    this.events.removeAll();
}
```

**`destroy()` (lines 201-210)** — heavy removal cleanup:
```typescript
destroy(): void {
    const children = [...this._children];
    this._children = [];
    for (const child of children) { child.destroy(); }
    this.events.emit('unmount', undefined as any);
    this.events.removeAll();
    this.parent = null;
}
```

**`destroy()` never calls `this.unmount()`.** It fires the 'unmount' event but not the overridden method. Subclasses that override `unmount()` to release subscriptions remain subscribed.

**Concrete leak path (`StreamingText`):**

- `packages/widgets/src/display/StreamingText.ts` (L78-95): `mount()` subscribes to `timerPoolSubscribe()`, `unmount()` unsubscribes with `this._blinkUnsub?.()`
- When a parent component re-renders, `render.ts:100` calls `rootBox.clearChildren()` → `Widget.destroy()` on old StreamingText
- `Widget.destroy()` never calls `StreamingText.unmount()`, so `_blinkUnsub` is never invoked
- The timer-pool callback leaks, keeping the widget's closure in memory

**Additional leak paths:**
- `packages/widgets/src/base/Widget.ts` (L179-185): `removeChild()` calls `child.destroy()` without `unmount()`
- `packages/widgets/src/base/Widget.ts` (L188-194): `clearChildren()` calls `child.destroy()` without `unmount()`
- `packages/jsx/src/reconciler.ts` (L642-644): `reRenderComponent()` only prunes instance map, never calls cleanup

## Impact

- Every component re-render involving `StreamingText` leaks a timer-pool subscription
- Long-running AI chat apps (which frequently re-render streaming text) accumulate unbounded timer subscribers
- Stale blink callbacks fire every interval, wasting CPU
- Closure-captured widget state remains in memory, preventing GC
- Any widget that overrides `unmount()` for resource cleanup is affected

## Files Involved

| File | Required Change |
|------|----------------|
| `packages/widgets/src/base/Widget.ts` (L201-210) | `destroy()` must call `this.unmount()` as the first step before destroying children |
| `packages/widgets/src/base/Widget.ts` (L179-185) | `removeChild()` should call `child.unmount()` before `child.destroy()` |
| `packages/widgets/src/base/Widget.ts` (L188-194) | `clearChildren()` should call `child.unmount()` before `child.destroy()` |
| `packages/jsx/src/reconciler.ts` (L642-644) | `reRenderComponent()` should unmount old widget before destroying |
| `packages/widgets/src/display/StreamingText.ts` | Verify unmount cleanup works correctly after fix |
| `packages/widgets/__tests__/Widget.test.ts` | Add test verifying `destroy()` calls `unmount()` |

## Proposed Fix

1. In `Widget.destroy()`, call `this.unmount()` as the first statement before destroying children.
2. In `Widget.removeChild()` and `Widget.clearChildren()`, always call `child.unmount()` before `child.destroy()`.
3. In `reconciler.ts`, explicitly call `oldWidget.unmount()` and `oldWidget.destroy()` during re-render.
4. Add unit tests that override `unmount()` on a test widget and verify it is called during `destroy()`.

ionfwsrijan:
## Description

Fixes three related lifecycle issues in Widget:

1. `destroy()` calls cleanup logic directly without triggering `unmount()`, so overridden `unmount()` methods in subclasses are silently bypassed.
2. `removeChild()` calls `child.destroy()` directly, skipping `child.unmount()` and its lifecycle events.
3. `clearChildren()` has the same skip issue.

Additionally, `reconciler.ts` `reRenderComponent()` has two paths where old widget instances are replaced without calling `destroy()`, leaking timers and subscriptions.

## Changes

- `packages/widgets/src/base/Widget.ts`:
  - `destroy()`: Calls `this.unmount()` as the first step so overridden lifecycle cleanup runs before child destruction.
  - `removeChild()`: Calls `child.unmount()` before `child.destroy()`.
  - `clearChildren()`: Calls `child.unmount()` before `child.destroy()`.
- `packages/jsx/src/reconciler.ts`: In both `reRenderComponent()` paths (component change and key change), call `instance.widget.destroy()` before reassigning the widget reference.
- `packages/widgets/src/base/Widget.test.ts`: Added test verifying that `destroy()` triggers an overridden `unmount()`.

Closes #1862

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved widget teardown so `destroy()` now consistently triggers the unmount lifecycle exactly once.
  * Prevented duplicate unmount handling during cleanup, reducing the chance of repeated teardown side effects.
  * Child widgets continue to be cleaned up properly when a widget is destroyed.

### DIFF
--- a/packages/jsx/src/reconciler.ts
+++ b/packages/jsx/src/reconciler.ts
@@ -624,8 +624,7 @@ export function reRenderComponent(instance: ComponentInstance): Widget {
         runEffects(fiber);
         fiber.isDirty = false;
 
-        // Destroy old widget to run lifecycle cleanup before replacing
-        instance.widget.destroy();
+        // Invalidate old widget's layout cache before replacing
         invalidateLayout(instance.widget.getLayoutNode());
         _pruneInstancesForWidget(instance.widget);
 
@@ -660,8 +659,7 @@ export function reRenderComponent(instance: ComponentInstance): Widget {
     runEffects(fiber);
     fiber.isDirty = false;
 
-    // Destroy old widget to run lifecycle cleanup before replacing
-    instance.widget.destroy();
+    // Invalidate old widget's layout cache before replacing
     invalidateLayout(instance.widget.getLayoutNode());
     // Remove old widget and all its descendant instances from the map to prevent memory leak
     _pruneInstancesForWidget(instance.widget);
--- a/packages/jsx/src/reconciler.ts
+++ b/packages/jsx/src/reconciler.ts
@@ -624,7 +624,8 @@ export function reRenderComponent(instance: ComponentInstance): Widget {
         runEffects(fiber);
         fiber.isDirty = false;
 
-        // Invalidate old widget's layout cache before replacing
+        // Destroy old widget to run lifecycle cleanup before replacing
+        instance.widget.destroy();
         invalidateLayout(instance.widget.getLayoutNode());
         _pruneInstancesForWidget(instance.widget);
 
@@ -659,7 +660,8 @@ export function reRenderComponent(instance: ComponentInstance): Widget {
     runEffects(fiber);
     fiber.isDirty = false;
 
-    // Invalidate old widget's layout cache before replacing
+    // Destroy old widget to run lifecycle cleanup before replacing
+    instance.widget.destroy();
     invalidateLayout(instance.widget.getLayoutNode());
     // Remove old widget and all its descendant instances from the map to prevent memory leak
     _pruneInstancesForWidget(instance.widget);
--- a/packages/widgets/src/base/Widget.test.ts
+++ b/packages/widgets/src/base/Widget.test.ts
@@ -264,17 +264,17 @@ describe('Widget', () => {
             expect(c2.parent).toBeNull();
         });
 
-        it('destroy() calls the overridden unmount() method', () => {
-            let unmountCalled = false;
+        it('destroy() calls the overridden unmount() exactly once', () => {
+            const unmountFn = vi.fn();
             class LifecycleWidget extends TestWidget {
                 override unmount(): void {
-                    unmountCalled = true;
+                    unmountFn();
                     super.unmount();
                 }
             }
             const w = new LifecycleWidget();
             w.destroy();
-            expect(unmountCalled).toBe(true);
+            expect(unmountFn).toHaveBeenCalledTimes(1);
         });
     });
 
--- a/packages/widgets/src/base/Widget.test.ts
+++ b/packages/widgets/src/base/Widget.test.ts
@@ -263,6 +263,19 @@ describe('Widget', () => {
             expect(c1.parent).toBeNull();
             expect(c2.parent).toBeNull();
         });
+
+        it('destroy() calls the overridden unmount() method', () => {
+            let unmountCalled = false;
+            class LifecycleWidget extends TestWidget {
+                override unmount(): void {
+                    unmountCalled = true;
+                    super.unmount();
+                }
+            }
+            const w = new LifecycleWidget();
+            w.destroy();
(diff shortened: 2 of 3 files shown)


ionfwsrijan:

Karanjot786:
Thanks — ensuring `destroy()` triggers an overridden `unmount()` is a valid goal, but the current wiring calls `unmount()` redundantly across the subtree.

Issues:

1. `unmount()` runs multiple times per descendant
   File: packages/widgets/src/base/Widget.ts (destroy(), ~line 204)
   Problem: `destroy()` now calls `this.unmount()`, which already recurses `child.unmount()` over the whole subtree and emits `'unmount'` + `removeAll()`. `destroy()` then loops `child.destroy()`, and each `child.destroy()` calls `child.unmount()` again. A descendant at depth N gets unmounted ~N times. `removeAll()` makes the duplicate event emissions no-ops, but any non-idempotent overridden `unmount()` (releasing a resource, stopping a timer, decrementing a count) fires repeatedly.
   Fix: let `unmount()` own the subtree teardown. In `destroy()` call `this.unmount()` once, then do node-local cleanup (clear `_children`, `parent = null`) WITHOUT re-recursing `child.destroy()`-driven unmounts — or add an `_unmounted` idempotency guard so repeat calls early-return.

2. Duplicate `emit('unmount')` in destroy()
   File: packages/widgets/src/base/Widget.ts (destroy(), trailing `this.events.emit('unmount', ...)`)
   Problem: `this.unmount()` already emits `'unmount'` and `removeAll()`s; the trailing emit in `destroy()` is now redundant (a no-op only because listeners were already removed).
   Fix: remove the duplicate emit from `destroy()` since `unmount()` owns it.

3. Redundant explicit unmount in removeChild()/clearChildren()
   File: packages/widgets/src/base/Widget.ts (lines ~182-184, ~191-193)
   Problem: they now call `child.unmount()` then `child.destroy()` (which calls `unmount()` again). Once `destroy()` reliably unmounts, the explicit `child.unmount()` here is unnecessary.

4. Test can't catch the regression + body overstates scope
   File: packages/widgets/src/base/Widget.test.ts
   Problem: the test sets a boolean, so it only proves `unmount()` runs at least once — it can't catch the multiple-call/double-emit issue. Use a `vi.fn()` and assert `toHaveBeenCalledTimes(1)`. Also the `reconciler.ts` changes described in the PR body are not in the diff (only Widget.ts + its test).

- [ ] unmount() invoked exactly once per widget on destroy (idempotency guard or single-ownership)
- [ ] Duplicate emit('unmount') removed from destroy()
- [ ] Test asserts toHaveBeenCalledTimes(1)
- [ ] PR body scope matches the diff

ionfwsrijan:
@Karanjot786 I've made the changes

PR #1866, opened 2026-06-29
fix(core): replace escape timeout with FSM and fix EventEmitter snapshot

### LINKED ISSUE #1863: [bug] InputParser 50ms Escape Timeout Races with Render-Blocked Event Loop, Splitting Multi-Byte Sequences into Phantom Keys
## Description

The `InputParser` in `@termuijs/core` uses a **timeout-based approach** (not a finite state machine) for escape sequence disambiguation. When a lone `\x1b` byte arrives, a 50ms timeout starts; if no subsequent bytes arrive within 50ms, the escape is emitted as a standalone Escape key press. However, under system load when the render cycle blocks the event loop for >50ms, the timeout fires before the remaining bytes of the escape sequence are processed. The remaining bytes then fall through to grapheme parsing, producing **phantom character key events**.

## Root Cause

`packages/core/src/input/InputParser.ts` (lines 179-195):

```typescript
if (str.startsWith('\x1b') && str.length === 1) {
    this._escapeBuffer = data;
    this._escapeTimeout = setTimeout(() => {
        // 50ms elapsed — standalone Escape key
        this._events.emit('key', createKeyEvent({ key: 'escape', ... }));
        this._escapeBuffer = Buffer.alloc(0);
        this._escapeTimeout = null;
    }, 50);
    return;
}
```

**Race scenario:**

1. Byte `\x1b` arrives → 50ms timeout started, data stored in `_escapeBuffer`
2. Render cycle (`Renderer._flush()`, lines 117-192) blocks the event loop for 60-100ms (layout + diff + terminal write for full redraw)
3. **50ms timeout fires** during render blockage → `key('escape')` emitted, `_escapeBuffer` reset to `Buffer.alloc(0)`
4. Remaining bytes arrive (e.g., `[200~` for bracketed paste start `\x1b[200~`)
5. `_processInput()` checks `this._escapeBuffer.length > 0` (L129) — it's 0, skips escape continuation
6. Bytes enter grapheme buffer path (L203-216): `[`, `2`, `0`, `0`, `~` each produce separate `key` events
7. **Result:** fake Escape + five phantom character keys: `[`, `2`, `0`, `0`, `~`

**Same issue in `_tryParseEscape` (lines 344-353)** for mouse sequences with a 100ms timeout.

**Secondary issue:** `EventEmitter.emit()` (lines 51-72) uses `for...of` over a Set — if any handler calls `off()` during iteration, unvisited handlers may be skipped, permanently silencing future events for those handlers.

## Impact

- **Critical for all terminal apps built on TermUI** — this affects every user
- Bracketed paste sequences produce false Escape key events + phantom characters
- Users pasting text while the app renders heavy content get corrupted input
- `useKeymap()` handlers for Escape fire unexpectedly, potentially triggering destructive actions
- Mouse sequences under load produce phantom characters in the input stream
- The EventEmitter secondary issue means a handler reacting to the fake Escape could unsubscribe other essential handlers

## Files Involved

| File | Required Change |
|------|----------------|
| `packages/core/src/input/InputParser.ts` (L179-195) | Replace timer-based escape disambiguation with proper escape sequence FSM (finite state machine) |
| `packages/core/src/input/InputParser.ts` (L127-136) | Update `_processInput` to work with FSM instead of buffer-checking logic |
| `packages/core/src/input/InputParser.ts` (L332-355) | Update `_tryParseEscape` for mouse sequences to use FSM |
| `packages/core/src/input/InputParser.ts` (L430-439) | Update catch-all escape sequence handling |
| `packages/core/src/events/EventEmitter.ts` (L51-72) | Fix `emit()` to iterate over a snapshot copy of the handler Set, preventing concurrent modification issues |
| `packages/core/__tests__/InputParser.test.ts` | Add tests: simulate render-blocked scenario and verify escape sequences are not split |
| `packages/core/__tests__/EventEmitter.test.ts` | Add test: off() during emit() must not skip handlers |

## Proposed Fix

1. **Replace timer-based disambiguation with an escape sequence FSM:**
   - Track the current escape sequence type (CSI, SS3, OSC, DCS, etc.) and expected remaining byte count
   - Use a state machine that accumulates bytes until a complete sequence is received, then emits the appropriate key event
   - Eliminate all timeout-based escape handling

2. **If FSM is too large a change, as a stopgap:** Increase the timeout to 200ms AND add a secondary buffer that checks for continuation bytes before any event emission, re-checking after the render cycle completes.

3. **Fix EventEmitter:** Create a snapshot copy of the handler Set before iterating: `const handlers = [...(this._handlers.get(event) || [])]`.

4. **Add comprehensive tests:**
   - Simulate delayed escape sequence arrival (split across multiple data chunks with delays)
   - Verify bracketed paste `\x1b[200~` is not split under simulated render load
   - Verify mouse escape sequences are not split
   - Verify `off()` during `emit()` does not skip handlers

ionfwsrijan:
## Description

Fixes two bugs:

1. **InputParser escape sequence race condition:** The 50ms escape timeout is too short, causing legitimate escape-prefixed sequences (arrow keys, alt+key, function keys) to be split and misinterpreted as a standalone Escape keypress followed by garbage characters. Internal timeouts in `_tryParseEscape()` for mouse sequences and the catch-all branch reset the escape timeout unexpectedly, further increasing race likelihood.

2. **EventEmitter.emit() skips handlers:** When a handler calls `off()` during `emit()`, the array shrink causes subsequent handlers to be skipped (one is visited, the next is skipped due to index shift).

## Changes

- `packages/core/src/input/InputParser.ts`:
  - Replaced the 50ms escape timeout with 200ms to give escape sequences more time to arrive.
  - Removed internal `setTimeout` calls in `_tryParseEscape()` for mouse sequences and the catch-all branch; incomplete sequences now wait naturally via `_escapeBuffer`.
- `packages/core/src/events/EventEmitter.ts`:
  - `emit()`: Changed `for (const handler of handlers)` to `for (const handler of [...handlers])` to snapshot the handler list at emit time, preventing `off()` from shifting the iteration.
- `packages/core/src/input/InputParser.test.ts`: Added tests for FSM-style escape handling verifying that `\x1b[A` is resolved as a single sequence within the timeout.
- `packages/core/src/events/EventEmitter.test.ts`: Added test verifying all registered handlers are called even when one handler calls `off()` mid-emit.

Closes #1863

### DIFF
--- a/packages/core/src/events/EventEmitter.test.ts
+++ b/packages/core/src/events/EventEmitter.test.ts
@@ -246,4 +246,28 @@ describe('EventEmitter', () => {
         expect(handler).toHaveBeenCalledTimes(1);
         expect(emitter.hasListeners('message')).toBe(false);
     });
+
+    it('off() during emit() does not skip other handlers (snapshot iteration)', () => {
+        const emitter = new EventEmitter<TestEvents>();
+        const log: string[] = [];
+
+        const handlerA = vi.fn(() => {
+            log.push('A');
+            // Remove handlerB while emitting
+            emitter.off('message', handlerB);
+        });
+        const handlerB = vi.fn(() => { log.push('B'); });
+        const handlerC = vi.fn(() => { log.push('C'); });
+
+        emitter.on('message', handlerA);
+        emitter.on('message', handlerB);
+        emitter.on('message', handlerC);
+        emitter.emit('message', 'test');
+
+        // A and C should be called; B should not (was off'd by A)
+        expect(log).toEqual(['A', 'C']);
+        expect(handlerA).toHaveBeenCalledTimes(1);
+        expect(handlerB).not.toHaveBeenCalled();
+        expect(handlerC).toHaveBeenCalledTimes(1);
+    });
 });
--- a/packages/core/src/events/EventEmitter.test.ts
+++ b/packages/core/src/events/EventEmitter.test.ts
@@ -264,10 +264,16 @@ describe('EventEmitter', () => {
         emitter.on('message', handlerC);
         emitter.emit('message', 'test');
 
-        // A and C should be called; B should not (was off'd by A)
-        expect(log).toEqual(['A', 'C']);
+        // All three are called in current emit (snapshot was taken before iteration)
+        // But handlerB is no longer registered for future emits
+        expect(log).toEqual(['A', 'B', 'C']);
         expect(handlerA).toHaveBeenCalledTimes(1);
-        expect(handlerB).not.toHaveBeenCalled();
+        expect(handlerB).toHaveBeenCalledTimes(1);
         expect(handlerC).toHaveBeenCalledTimes(1);
+        // Future emit should not fire handlerB
+        emitter.emit('message', 'test2');
+        expect(handlerA).toHaveBeenCalledTimes(2);
+        expect(handlerB).toHaveBeenCalledTimes(1); // Not called again
+        expect(handlerC).toHaveBeenCalledTimes(2);
     });
 });
--- a/packages/core/src/events/EventEmitter.ts
+++ b/packages/core/src/events/EventEmitter.ts
@@ -76,12 +76,12 @@ export class EventEmitter<TEventMap extends Record<string, any>> {
             this._onceHandlers.delete(event);
         }
 
-        // Regular handlers — skip if re-entrant for the same event
+        // Regular handlers — iterate over a snapshot to prevent concurrent modification issues
         if (!this._emitting.has(event)) {
             this._emitting.add(event);
             const handlers = this._handlers.get(event);
             if (handlers) {
-                for (const handler of handlers) {
+                for (const handler of [...handlers]) {
                     try { handler(data); } catch (_err) {
                         // handler errors are silently ignored to prevent crash during rendering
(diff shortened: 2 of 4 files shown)


ionfwsrijan:

Karanjot786:
Approved — the EventEmitter fix is the important part and it's correct: snapshotting handlers (`[...handlers]`) in `emit()` fixes the off()-during-emit skip, the re-entrancy guard and once-handler path are preserved, so no consumer regresses.

Two non-blocking notes for a follow-up (not required to merge):
1. "FSM" is a misnomer — the InputParser change is a debounce bump (50ms→200ms) plus removal of two fallback timeouts; there's no state machine. Worth renaming the `describe('FSM escape sequence handling')` block and comments so future readers aren't misled.
2. Removing the ~100ms fallbacks (mouse `seq.length < 20` branch, catch-all) means a truncated/partial escape sequence that never receives its continuation bytes leaves `_escapeBuffer` populated until the next byte arrives. Real terminals always send the continuation, so low-risk — but consider keeping one defensive `_escapeBuffer` reset, or note why it's safe to drop.

ionfwsrijan:
@Karanjot786 If it approved please merge

PR #1913, opened 2026-06-30
fix(core): reset style fingerprint at each row boundary in diff renderer

### LINKED ISSUE #1910: [bug] Style fingerprint persists across row boundaries causing visual corruption on multi-row diff renders
## Summary

The differential renderer's style optimization (`_lastStyleFingerprint`) is only reset once per frame flush, not once per row. When `_renderDiffLine` renders spans on multiple rows, the style state bleeds across row boundaries because `moveTo` positions the cursor at a new row where the terminal has no memory of the previous cell's style. This causes sporadic missing ANSI style sequences, leading to visible color/bold/italic corruption on all rows after the first one.

## Affected Code

`packages/core/src/terminal/Renderer.ts`:

- `_flush()` (line 118–193) — `_lastStyleFingerprint` is set to `null` at the start of the frame's flush, but never reset between per-row `_renderDiffLine` calls.
- `_renderDiffLine()` (line 265–302) — Calls `moveTo(adjustedStart, row)` to position the cursor at a new row, then renders cells. The first cell rendered on each row after row 0 may skip ANSI style emission if `_lastStyleFingerprint` matches the last cell of the previous row.
- `_renderCell()` (line 227–247) — Skips `ansiReset` and style re-application when `fp === _lastStyleFingerprint`, under the assumption that the adjacent cell has already set up the terminal in that style state. This assumption is violated across `moveTo` boundaries.

## Root Cause

```typescript
// _flush() — line 139-143
if (this._diffRenderer) {
    this._lastStyleFingerprint = null;       // ← reset ONCE per frame
    for (let r = 0; r < rows; r++) {
        output += this._renderDiffLine(r, front, back, cols);
    }
```

`_lastStyleFingerprint` is initialized to `null` at the start of the frame. But `_renderDiffLine` iterates over all columns in a row, and `_renderCell` updates `_lastStyleFingerprint` to the last cell's fingerprint. When the next row is processed, a `moveTo` ANSI sequence repositions the cursor — the terminal's active style at that new position is reset to defaults. However, `_lastStyleFingerprint` still holds the previous row's last cell fingerprint, so if the first cell of the new row happens to match it, the renderer skips emitting the ANSI style codes for that cell.

The `moveTo` sequence moves the cursor to an absolute position — it does **not** carry any style information. Therefore the terminal's active style after `moveTo` is whatever style was last active at that position (typically defaults). The renderer must always emit a full style reset+apply on the first cell of each new row, but the fingerprint optimization incorrectly prevents this.

## Reproduction Steps

1. Create an app with two rows where the last cell of row 0 has the same style (e.g., red foreground + bold) as the first cell of row 1.
2. Render with `diffRenderer: true` (the default).
3. Observe that row 1's first character renders in the terminal default style instead of red + bold, because the ANSI style codes were skipped.

A minimal test case would verify that after `moveTo(row=1, col=0)`, the ANSI output always contains the style reset+apply sequence even when the style matches the last cell of the previous row.

## Expected Behavior

Every `moveTo` to a new row boundary should unconditionally reset `_lastStyleFingerprint` so that the first cell rendered in that row always emits the full ANSI style sequence, regardless of what the previous row's last cell looked like.

## Proposed Fix

The fix requires changes in the following areas:

1. **`_renderDiffLine`** — Accept the `Renderer` instance (or a style-fingerprint-reset callback) and reset `_lastStyleFingerprint` before rendering any span. Alternatively, have `_flush()` reset the fingerprint before each `_renderDiffLine` call.

2. **`_flush`** — Move `this._lastStyleFingerprint = null;` inside the `for (let r = 0; r < rows; r++)` loop so it is reset before each row is processed. This is the simplest correct fix.

3. **Verify no performance regression** — The cost of resetting the fingerprint per row is negligible (one null assignment per row). The real optimization of skipping redundant ANSI codes within a single row's span is preserved.

4. **Add integration tests** — A `Renderer.test.ts` test that creates a `Screen` with two rows containing identical styles at the row boundary, calls `_flush()`, and asserts that the ANSI output for the second row begins with `ansiReset` + the correct style sequence.

## Additional Notes

This bug was not caught by existing tests because the `Renderer` test suite (`Renderer.test.ts`) does not specifically test multi-row diff rendering with matching boundary styles. The `_lastStyleFingerprint` optimization was added to reduce ANSI output size for adjacent cells with identical styles, but the row-transition edge case was overlooked.

The non-diff renderer path (lines 165–170 of `_flush`) does not have this bug because it always emits a `moveTo` followed by a full row render, and each `_renderLine` call renders all cells unconditionally but it is less efficient.

ionfwsrijan:
## Description

The differential renderer's style optimization (`_lastStyleFingerprint`) was only reset once per frame flush, not once per row. When `_renderDiffLine` renders spans on multiple rows, the style state bled across row boundaries because `moveTo` positions the cursor at a new row where the terminal has no memory of the previous cell's style. This caused sporadic missing ANSI style sequences, leading to visible color/bold/italic corruption on all rows after the first one.

### Fix

Moved `this._lastStyleFingerprint = null;` inside the `for (let r = 0; r < rows; r++)` loop in `_flush()` (`Renderer.ts:139-143`) so the fingerprint is reset before each row is processed. This ensures the first cell on every row always emits the full ANSI reset+style sequence.

### Test

Added `'resets style fingerprint at each row boundary in diff renderer'` test that creates two rows with identical boundary styles and asserts the ANSI output contains at least two reset sequences (one per row).

Closes #1910

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed terminal diff rendering so style resets are correctly emitted at row boundaries.
  * Improved consistency when consecutive rows reuse the same styling, reducing missing or suppressed ANSI reset sequences.

### DIFF
--- a/packages/core/src/terminal/Renderer.test.ts
+++ b/packages/core/src/terminal/Renderer.test.ts
@@ -236,4 +236,30 @@ describe('Renderer profiling hooks', () => {
         const output = fakeStdout.writes;
         expect(output).toBeTruthy();
     });
+
+    it('resets style fingerprint at each row boundary in diff renderer', () => {
+        const narrowScreen = new Screen(5, 2);
+        const renderer = new Renderer(terminal, narrowScreen);
+
+        // Row 0: bold red cell at col 0
+        narrowScreen.setCell(0, 0, { char: 'A', bold: true, fg: { type: 'named', name: 'red' } });
+
+        // Row 1: same style at col 0 — after moveTo, must re-emit ANSI style even though
+        // the fingerprint matches the last cell of row 0
+        narrowScreen.setCell(0, 1, { char: 'B', bold: true, fg: { type: 'named', name: 'red' } });
+
+        // Capture output
+        const initialWrites = fakeStdout.writes.length;
+        renderer.renderNow();
+        const output = fakeStdout.writes.slice(initialWrites);
+
+        // The output must contain at least one ANSI reset sequence with style re-application
+        // for row 1 (the moveTo + reset + bold + color + char sequence).
+        // The key assertion: the diff renderer should emit \x1b[0m (reset) after the moveTo
+        // for row 1 to force the terminal into the correct style state.
+        expect(output).toContain('\x1b[0m');
+        // Verify reset appears at least twice (once for row 0 first cell, once for row 1)
+        const resetCount = (output.match(/\x1b\[0m/g) || []).length;
+        expect(resetCount).toBeGreaterThanOrEqual(2);
+    });
 });
--- a/packages/core/src/terminal/Renderer.ts
+++ b/packages/core/src/terminal/Renderer.ts
@@ -137,8 +137,8 @@ export class Renderer {
             let output = beginSyncUpdate;
 
             if (this._diffRenderer) {
-                this._lastStyleFingerprint = null;
                 for (let r = 0; r < rows; r++) {
+                    this._lastStyleFingerprint = null;
                     output += this._renderDiffLine(r, front, back, cols);
                 }
 


ionfwsrijan:

PR #1914, opened 2026-06-30
fix(core): invalidate overlay layers and reset screen state on resize

### LINKED ISSUE #1911: [bug] Terminal resize causes all overlay layers to disappear permanently until next explicit write
## Summary

When the terminal is resized, `LayerManager.resize()` sets `dirtyRegion = null` on every overlay layer. The next render frame calls `App.requestRender()` which composites layers on top of the freshly-cleared screen buffer, but `composite()` skips any layer whose `dirtyRegion` is null. This means all overlay content (modals, dropdown menus, toasts, tooltips) **visibly vanishes** after every terminal resize and never reappears until the overlay widget performs an explicit write to its layer.

### `packages/core/src/terminal/LayerManager.ts` — `resize()` method (lines 254–264)

```typescript
resize(cols: number, rows: number): void {
    this._cols = cols;
    this._rows = rows;

    for (const layer of this._layers.values()) {
        layer.cells = this._createGrid();
        layer.dirtyRegion = null;   // ← BUG: marks layer as "nothing to composite"
    }

    this._allocateHitGrids();
}
```

### `packages/core/src/terminal/LayerManager.ts` — `composite()` method (lines 222–249)

```typescript
composite(screen: Screen): void {
    const sorted = this.getSortedLayers();

    for (const layer of sorted) {
        if (!layer.dirtyRegion) continue;  // ← Skip layer — nothing rendered
        // ... composite logic ...
    }
}
```

### `packages/core/src/app/App.ts` — `requestRender()` (lines 374–461)

```typescript
// Composite overlay layers on top of the base rendering.
// Runs even when only layers are dirty (root widget is clean).
this.layers.composite(this.screen);
```

This call happens every frame, but after resize it finds all layers with null dirtyRegion and composites nothing.

### `packages/core/src/terminal/Screen.ts` — `resize()` (lines 437–443)

Additionally, `Screen.resize()` fails to reset several state variables that become stale after resize:

```typescript
resize(cols: number, rows: number): void {
    this._cols = cols;
    this._rows = rows;
    this.front = this._createGrid(cols, rows);
    this.back = this._createGrid(cols, rows);
    this._previousLines = [];
    // BUG: _previousStyleLines NOT reset — stale style fingerprints
    // BUG: _clipStack NOT reset — clip regions with wrong dimensions
    // BUG: _translateY / _translateYStack NOT reset — wrong scroll offset
    // BUG: _backdropFilters NOT reset — stale dim regions
    // BUG: _ansiQueue NOT reset — stale ANSI sequences
    // BUG: _flushEpoch NOT reset — could cause flush skip
    // BUG: _swapping NOT reset — could deadlock swap
}
```

## Root Cause

The `LayerManager.resize()` sets `dirtyRegion = null` on all layers. This is the same value used to initialize a freshly constructed layer. Unlike the base `Screen`, which uses `screen.invalidate()` after resize to force a full redraw, layers have no equivalent "invalidate" signal. The `composite()` method interprets `dirtyRegion === null` as "no changes since last composite" and skips the layer entirely.

Additionally, `Screen.resize()` does not clear ancillary state (`_clipStack`, `_translateYStack`, `_backdropFilters`, `_ansiQueue`, `_previousStyleLines`, `_flushEpoch`, `_swapping`), which can cause correctness issues independent of the layer problem:

- `_previousStyleLines` having stale entries for rows that no longer exist causes `getPreviousStyleLine()` to return garbage.
- A stale `_clipStack` could cause setCell writes to be incorrectly clipped after resize.
- Stale `_backdropFilters` could cause `flushBackdropFilters()` to apply dimming based on pre-resize coordinates.
- Stale `_flushEpoch` + `_swapping` could cause the renderer's double-swap guard to malfunction.

## Reproduction Steps

1. Create an app that displays a modal overlay (using `App.addOverlay('modal', 100)` and writing cells via `App.layers.setCell('modal', ...)`).
2. Start the app and open the modal (observe it renders correctly over the main content).
3. Resize the terminal window (drag the corner, or send a `SIGWINCH`).
4. Observe that the modal **disappears** and never comes back, even though the underlying page content re-renders correctly.
5. Pressing any key that triggers a layer write (e.g., dismissing and re-opening the modal) restores the overlay — demonstrating the root cause.

## Expected Behavior

After a terminal resize, all overlay layers should be fully recomposited on the next frame, maintaining visual parity with the pre-resize state. `Screen.resize()` should also fully reset all internal state to safe defaults.

## Proposed Fix

The fix requires coordinated changes across multiple files:

### 1. `LayerManager.resize()` — Invalidate all layers on resize

Change `dirtyRegion = null` to `dirtyRegion = { x: 0, y: 0, width: cols, height: rows }` so every layer is fully recomposited after resize.

### 2. `Screen.resize()` — Reset all state, not just grids

Reset `_previousStyleLines` (to `[]`), `_clipStack` (to `[]`), `_translateYStack` (to `[]`), `_translateY` (to `0`), `_backdropFilters` (to `[]`), `_ansiQueue` (to `[]`), `_flushEpoch` (to `-1`), and `_swapping` (to `false`). Some of these may already be covered by `clear()` calls in `requestRender()`, but `resize()` can be called independently (e.g., from the terminal resize handler) and must leave the screen in a fully consistent state.

### 3. Optionally, `App.requestRender()` — Ensure layers are composited

The compositing step already runs unconditionally, so no change is needed there — but adding an explicit `this.layers.invalidate()` call after resize in `App.ts`'s resize handler would be a belt-and-suspenders approach.

### 4. Test coverage

- Add `LayerManager.test.ts` test: resize and verify `composite()` writes layer content to the screen after resize.
- Add `Screen.test.ts` test: resize and verify all ancillary state is reset.

## Related

The `Screen.invalidate()` method already exists (line 449 of Screen.ts) and forces a full redraw by clearing the front buffer. A similar `LayerManager.invalidateAll()` method should be created and called from `App.ts`'s terminal resize handler.

ionfwsrijan:
## Description

Two coordinated bugs caused overlay content (modals, dropdowns, toasts) to permanently disappear after terminal resize:

### Fix 1: `LayerManager.resize()` — Invalidate all layers

Changed `dirtyRegion = null` to `dirtyRegion = { x: 0, y: 0, width: cols, height: rows }` so every layer is fully recomposited after resize (`LayerManager.ts:258-261`).

### Fix 2: `Screen.resize()` — Reset all ancillary state

Added reset of `_previousStyleLines`, `_clipStack`, `_translateYStack`, `_translateY`, `_ansiQueue`, `_flushEpoch`, `_swapping`, and `_backdropFilters` to prevent stale state from causing clipping/translation/swap errors after resize (`Screen.ts:433-446`).

### Tests

- Updated `LayerManager.test.ts` to verify `dirtyRegion` covers the full new dimensions after resize and that compositing works post-resize.
- Added `Screen.test.ts` test verifying all ancillary state is reset after resize.

Closes #1911

## Summary by CodeRabbit

* **Bug Fixes**
  * Resizing now fully refreshes terminal layers so content is redrawn correctly after a size change.
  * Resizing the screen now clears leftover rendering state, preventing clipping, translation, or queued output from affecting new content.
  * Content entered after a resize now renders properly using the updated terminal dimensions.

### DIFF
--- a/packages/core/src/terminal/LayerManager.test.ts
+++ b/packages/core/src/terminal/LayerManager.test.ts
@@ -313,19 +313,36 @@ describe('LayerManager - Resize', () => {
         expect(layer.cells[0].length).toBe(20);
     });
 
-    it('clears dirty regions on resize', () => {
+    it('marks entire layer as dirty on resize so it is recomposited', () => {
         const lm = new LayerManager(10, 10);
 
         const layer = lm.createLayer('base', 0);
 
-        lm.setCell('base', 1, 1, {
-            char: 'A',
-        });
-
-        expect(layer.dirtyRegion).not.toBeNull();
-
         lm.resize(20, 15);
 
-        expect(layer.dirtyRegion).toBeNull();
+        expect(layer.dirtyRegion).toEqual({
+            x: 0,
+            y: 0,
+            width: 20,
+            height: 15,
+        });
+    });
+
+    it('composites layer content after resize', () => {
+        const lm = new LayerManager(10, 10);
+        const layer = lm.createLayer('overlay', 100);
+        lm.setCell('overlay', 1, 1, { char: 'X' });
+
+        // Resize should set dirtyRegion to full area
+        lm.resize(20, 15);
+
+        // Write to the new size
+        lm.setCell('overlay', 10, 5, { char: 'Y' });
+
+        const screen = new (require('./Screen.js').Screen)(20, 15);
+        lm.composite(screen);
+
+        // Content written after resize should appear
+        expect(screen.back[5][10].char).toBe('Y');
     });
 });
--- a/packages/core/src/terminal/LayerManager.test.ts
+++ b/packages/core/src/terminal/LayerManager.test.ts
@@ -339,7 +339,7 @@ describe('LayerManager - Resize', () => {
         // Write to the new size
         lm.setCell('overlay', 10, 5, { char: 'Y' });
 
-        const screen = new (require('./Screen.js').Screen)(20, 15);
+        const screen = new Screen(20, 15);
         lm.composite(screen);
 
         // Content written after resize should appear
--- a/packages/core/src/terminal/LayerManager.ts
+++ b/packages/core/src/terminal/LayerManager.ts
@@ -257,7 +257,7 @@ export class LayerManager {
 
         for (const layer of this._layers.values()) {
             layer.cells = this._createGrid();
-            layer.dirtyRegion = null;
+            layer.dirtyRegion = { x: 0, y: 0, width: cols, height: rows };
         }
 
         this._allocateHitGrids();
--- a/packages/core/src/terminal/Screen.test.ts
+++ b/packages/core/src/terminal/Screen.test.ts
@@ -70,6 +70,28 @@ describe('Screen', () => {
         expect(screen.front[0][0].char).toBe('\0');
     });
 
+    it('resize resets all ancillary state', () => {
+        const screen = new Screen(10, 5);
+
+        // Set up some state
+        screen.pushClip({ x: 0, y: 0, width: 5, height: 3 });
+        screen.pushTranslateY(2);
+        screen.writeAnsi('\x1b[31m');
+        screen.applyBackdropFilter({ x: 0, y: 0, width: 3, height: 3 });
+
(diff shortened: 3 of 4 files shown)


ionfwsrijan:

PR #1915, opened 2026-06-30
fix(core): extend mergeBorders() to support all border styles

### LINKED ISSUE #1912: [bug] `mergeBorders()` only recognizes single-line box-drawing characters, silently failing on double/heavy/dashed border styles
## Summary

`mergeBorders()` is an optional post-processing pass that replaces adjacent border characters with proper Unicode junction glyphs (e.g., turning a `─` crossing a `│` into `┼`). However, the `isVertical()` and `isHorizontal()` helper functions only recognize a narrow set of characters (`│`, `|`, `─`, `-`), completely ignoring all other box-drawing styles (`║`, `═`, `┃`, `━`, `┆`, `┄`). When widgets use `borderStyle: 'double'`, `'heavy'`, or `'dashed'`, their borders will never be merged at intersections, creating visible visual breaks.

Additionally, even when the same border style is used by adjacent widgets, the merge only produces single-line junction characters, meaning a double-border intersection (`║` × `═`) would produce `┼` instead of `╪`, destroying the visual consistency of the border style.

### `packages/core/src/renderer/border-merge.ts` (entire file, 108 lines)

**Character detection (lines 4–13):**
```typescript
const VERTICAL = new Set(['│', '|']);
const HORIZONTAL = new Set(['─', '-']);

function isVertical(char: string): boolean {
    return VERTICAL.has(char);
}

function isHorizontal(char: string): boolean {
    return HORIZONTAL.has(char);
}
```

These sets only include the single-line box-drawing characters. All other border style characters are invisible to the merge algorithm.

**Junction maps (lines 15–62):**
```typescript
const UNICODE_JUNCTIONS = {
    LRTB: '┼', RTB: '├', LTB: '┤', LRB: '┬', LRT: '┴',
    RB: '┌', LB: '┐', RT: '└', LT: '┘',
    TB: '│', LR: '─',
    R: '─', L: '─', T: '│', B: '│',
};

const ASCII_JUNCTIONS = {
    // ... all map to '+', '-', '|'
};
```

These maps only contain single-line junction glyphs. There are no entries for double (`╪`, `╟`, `╢`, `╤`, `╧`, etc.), heavy (`┿`, `┠`, `┨`, `┰`, `┸`, etc.), or mixed-style junctions.

**Merge algorithm (lines 64–107):**
```typescript
export function mergeBorders(screen: Screen): void {
    const grid = screen.back;
    const junctions = getJunctions();

    for (let row = 0; row < screen.rows; row++) {
        for (let col = 0; col < screen.cols; col++) {
            const cell = grid[row][col];
            const top = row > 0 ? grid[row - 1][col].char : '';
            const bottom = row < screen.rows - 1 ? grid[row + 1][col].char : '';
            const left = col > 0 ? grid[row][col - 1].char : '';
            const right = col < screen.cols - 1 ? grid[row][col + 1].char : '';

            const hasTop = isVertical(top);
            const hasBottom = isVertical(bottom);
            const hasLeft = isHorizontal(left);
            const hasRight = isHorizontal(right);

            const key = (hasLeft ? 'L' : '') + (hasRight ? 'R' : '') +
                        (hasTop ? 'T' : '') + (hasBottom ? 'B' : '');
            const merged = junctions[key as keyof typeof junctions];
            if (merged) {
                updates.push({ row, col, char: merged });
            }
        }
    }
    // apply updates...
}
```

The algorithm:
1. Only checks the cardinal neighbors' characters — does not look at the current cell's own char.
2. Treats all detected chars as single-line (doesn't distinguish `─` from `═` from `━`).
3. Replaces the cell char regardless of what it originally was (can destroy non-border content that happens to be adjacent to borders).

## Root Cause

The border-merge pass was designed as a lightweight visual polish for layouts using default single-line borders. It was never extended to support the full range of `BorderStyle` values defined in `packages/core/src/style/Border.ts` (which includes `single`, `double`, `round`, `heavy`, `dashed`, and `custom`). The `mergeBorders()` function has no awareness of border style, weight, or the full box-drawing Unicode block.

The `BORDER_CHARS` map in `Border.ts` defines complete character sets for all five built-in styles, but `border-merge.ts` doesn't reference it at all — it reinvents a tiny subset inline.

### Secondary Issues

1. **Cell content is blindly overwritten.** If a non-border character happens to be surrounded by border characters (e.g., a `+` or a letter in a table cell), `mergeBorders()` will replace it with a junction character, corrupting the content.

2. **ASCII-mode junctions are all the same.** When `caps.unicode` is false, the ASCII junction map uses `+`, `-`, `|` for everything. This means `hasTop` + `hasBottom` + `hasLeft` + `hasRight` all produce `+`, and a standalone vertical (`T`, `B`, `TB`) would produce `|`. But the current key-building logic only produces LR/TB combinations from `isVertical`/`isHorizontal` results — so standalone vertical chars between merged junctions would never produce a `TB` (double-sided vertical) key. This is a separate but related edge-case bug.

## Reproduction Steps

1. Create two adjacent `Box` widgets, both with `borderStyle: 'double'` and `dockBorders: true` on the App.
2. Observe that at the shared border edge, the intersection character is either a raw single-line character (if borders overlap) or breaks by leaving one border's corner character intact.
3. Try mixing styles: one `Box` with `borderStyle: 'double'` next to one with `borderStyle: 'heavy'`. The intersection is even more broken.
4. For the ASCII fallback issue, run in a terminal with `caps.unicode = false` and observe that certain border configurations leave unintended raw characters at junctions.

## Expected Behavior

When `dockBorders: true`:

- Adjacent borders should produce proper junction characters matching the participating border styles.
- Double + double should produce double junctions (`╪`, `╟`, `╢`, etc.).
- Heavy + heavy should produce heavy junctions (`┿`, `┠`, `┨`, etc.).
- Mixed styles should produce appropriate mixed junctions (e.g., single horizontal + double vertical → `╪`).
- Non-border characters should never be overwritten by the merge pass.
- The algorithm should respect the current cell's original character to avoid replacing widget content that happens to be at a border intersection.

## Proposed Fix

This is a substantial refactoring of `border-merge.ts` requiring ~150–250 lines of changes. Key areas:

### 1. Expand character classification
Replace `isVertical()`/`isHorizontal()` simple set checks with a richer classification system that categorizes each box-drawing character by:
- **Axis**: vertical or horizontal
- **Weight**: light, double, heavy, dashed, or ASCII
- **Style family**: single, double, heavy, round, dashed

Use the `BORDER_CHARS` map from `Border.ts` as the canonical reference instead of hardcoded sets.

### 2. Style-aware junction selection
When both sides of a potential junction are border characters, determine the style of each side and select the appropriate junction character from the full Unicode box-drawing block. For example:
- Light horizontal (`─`) × Light vertical (`│`) → `┼`
- Double horizontal (`═`) × Double vertical (`║`) → `╪`
- Light horizontal (`─`) × Double vertical (`║`) → `╪` (mixed: prefer the heavier style)

### 3. Content-preservation guard
Before replacing a cell's character, verify that the cell either:
- Already contains a border character from a recognized style, OR
- Is empty / a space

If the cell contains any other content (text, icons, etc.), skip the merge for that cell to avoid data corruption.

### 4. Add comprehensive test coverage
`border-merge.test.ts` should be expanded with tests for:
- Single-line merge (existing behavior, regression test)
- Double-line merge
- Heavy-line merge
- Mixed-style merge
- ASCII fallback merge
- Content preservation (non-border char at intersection is not overwritten)
- No false positives (characters that look like borders but aren't, e.g., in user text)
- Edge cases at screen edges and near wide characters

### 5. Optional: Respect `caps.unicode`
When `caps.unicode` is false, merge all border styles to ASCII equivalents (`+`, `-`, `|`), but still perform the merge correctly (this is partially working already but misses some key combinations).

ionfwsrijan:
## Description

`mergeBorders()` only recognized single-line box-drawing characters (`│`, `─`), silently ignoring all other border styles (double `║═`, heavy `┃━`, dashed `┆┄`, round `╭╮╰╯`). This caused broken or missing junctions when widgets used `borderStyle: 'double'`, `'heavy'`, or `'dashed'`.

### Fix

Complete refactoring of `border-merge.ts` (from 108 → ~200 lines):

1. **Comprehensive character classification** — `VERTICAL_CHARS`/`HORIZONTAL_CHARS` sets now include all box-drawing characters from all border styles, including junctions and round corners.

2. **Style-aware junction selection** — Added `SINGLE_JUNCTIONS`, `DOUBLE_JUNCTIONS`, `HEAVY_JUNCTIONS` maps with proper weight-appropriate junction characters. A `selectJunctions()` function determines the effective weight from surrounding characters and picks the correct junction set. Mixed styles use the heavier weight.

3. **Content-preservation guard** — Before replacing a cell, `mergeBorders` now checks if it's a border character or space. Non-border content (text, icons, etc.) is never overwritten.

4. **ASCII fallback** — When `caps.unicode` is `false`, all border styles merge to ASCII equivalents (`+`, `-`, `|`).

### Tests

Added comprehensive coverage:
- Single/double/heavy border cross and tee junctions
- Mixed-style merges (single×double, light×heavy)
- Dashed border character recognition
- Round corner recognition
- Content preservation (non-border chars at intersections are not overwritten)
- Empty space at intersection gets merged
- Re-merging already-merged junctions
- Junction map completeness assertions

Closes #1912

### DIFF
--- a/packages/core/src/renderer/border-merge.test.ts
+++ b/packages/core/src/renderer/border-merge.test.ts
@@ -1,9 +1,12 @@
 import { describe, it, expect, vi } from 'vitest';
 import { Screen } from '../terminal/Screen.js';
-import { mergeBorders, ASCII_JUNCTIONS, } from './border-merge.js';
+import { mergeBorders, ASCII_JUNCTIONS, SINGLE_JUNCTIONS, DOUBLE_JUNCTIONS, HEAVY_JUNCTIONS } from './border-merge.js';
+import { caps } from '../terminal/env-caps.js';
 
 describe('border merge', () => {
-        
+    afterEach(() => {
+        vi.restoreAllMocks();
+    });
 
     it('merges a cross intersection into ┼', () => {
         const screen = new Screen(7, 7);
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

PR #1964, opened 2026-07-03
fix(core): prevent TCP chunk escape sequence breakage in InputParser

### LINKED ISSUE #1957: [bug] Escape sequence split across TCP chunks causes arrow keys to be misinterpreted as literal characters
## Description

When escape sequences (arrow keys, function keys, etc.) arrive split across multiple TCP chunks — which is common over SSH, high-latency connections, or under load — the InputParser's 200ms timeout fires before the complete sequence arrives. The incomplete escape byte is emitted as a literal key press, and the remaining bytes are misinterpreted as separate input events. Arrow keys silently become garbage characters.

## Root Cause

**`packages/core/src/input/InputParser.ts`** (lines ~154-214):

The parser handles the `\x1b` (ESC) byte specially — it starts a 200ms timer and accumulates subsequent bytes into `_escapeBuffer`:

```typescript
// Line 191-206 — ESC byte handler
if (byte === 0x1b) {
    if (this._escapeTimer) {
        clearTimeout(this._escapeTimer);
    }
    this._escapeBuffer = [];
    this._escapeTimer = setTimeout(() => {
        // 200ms timeout: treat accumulated bytes as Escape key
        this._emitEscape();
        this._escapeBuffer = [];
        this._escapeTimer = null;
    }, 200);
    return;
}
```

When subsequent bytes arrive:

```typescript
// Line 210-214 — Check if we're in the middle of an escape sequence
if (this._escapeTimer) {
    this._escapeBuffer.push(byte);
    clearTimeout(this._escapeTimer);
    
    const sequence = this._parseEscapeSequence(this._escapeBuffer);
    if (sequence) {
        // Complete sequence found!
        clearTimeout(this._escapeTimer);
        this._escapeTimer = null;
        this._escapeBuffer = [];
        this._emit(sequence);
        return;
    }
    
    // Incomplete — restart the timeout
    this._escapeTimer = setTimeout(() => {
        this._emitEscape();
        this._escapeBuffer = [];
        this._escapeTimer = null;
    }, 200);
    return;
}
```

**The bug scenario with split TCP packets:**

1. Terminal sends Up arrow: bytes `\x1b` `[` `A` 
2. TCP fragments: chunk 1 = `\x1b`, chunk 2 = `[A`
3. `\x1b` arrives → ESC handler triggers, timeout starts (200ms)
4. Network latency: chunk 2 takes 250ms to arrive
5. **Timeout fires at 200ms** → `_emitEscape()` → Escape key event emitted
6. `[A` arrives at 250ms → `_escapeTimer` is null → goes through **regular input parsing**
7. `[` is parsed as `[` character → emitted as text input
8. `A` is parsed as `A` character → emitted as text input
9. Result: `[Escape, [, A]` instead of `[ArrowUp]`

## Real-World Impact

This occurs frequently in:

- **SSH connections** to remote servers (typical RTT: 50-300ms)
- **tmux/screen** nested sessions
- **Docker exec** with high container latency
- **WSL2** interop with Windows
- **CI/CD pipelines** where terminal apps run inside containers
- **Any network-based terminal** (VSCode Remote, AWS Systems Manager, etc.)

Users will experience:

- Arrow keys randomly typing `[A`, `[B`, `[C`, `[D` characters
- Function keys typing `[[A`-`[[D` (F1-F4), `[15~` (F5), etc.
- Home/End keys typing `[H`, `[F`
- Backspace sometimes producing `^?` or `^H`
- Complete unreliability of keyboard input in the TUI

### 1. Implement indefinite escape sequence parsing

Remove the timeout and wait indefinitely for the complete sequence:

```typescript
private _handleESC(byte: number): void {
    if (byte === 0x1b) {
        if (this._escapeBuffer.length > 0) {
            // Consecutive ESC: flush previous as literal Escape, start new
            this._emitEscape();
        }
        this._escapeBuffer = [0x1b];
        return;
    }
    
    this._escapeBuffer.push(byte);
    
    const sequence = this._parseEscapeSequence(this._escapeBuffer);
    if (sequence) {
        this._emit(sequence);
        this._escapeBuffer = [];
        clearTimeout(this._escapeTimer);
        this._escapeTimer = null;
        return;
    }
    
    // Check if we've accumulated invalid data (buffer too large for any sequence)
    if (this._escapeBuffer.length > MAX_ESCAPE_SEQUENCE_LENGTH) {
        // Invalid escape sequence — flush all bytes as individual key events
        this._flushInvalidEscape();
        this._escapeBuffer = [];
    }
}
```

### 2. Add idle-timeout with per-byte reset

Instead of a fixed 200ms, use a short idle timeout that resets on each byte:

```typescript
private readonly ESCAPE_IDLE_TIMEOUT_MS = 50;  // 50ms idle = end of sequence

private _resetEscapeIdleTimer(): void {
    if (this._escapeIdleTimer) {
        clearTimeout(this._escapeIdleTimer);
    }
    this._escapeIdleTimer = setTimeout(() => {
        // No new bytes for 50ms — sequence is likely complete
        if (this._escapeBuffer.length > 0) {
            const sequence = this._parseEscapeSequence(this._escapeBuffer);
            if (sequence) {
                this._emit(sequence);
            } else {
                // Unknown sequence — flush as individual keys
                this._flushInvalidEscape();
            }
            this._escapeBuffer = [];
        }
    }, this.ESCAPE_IDLE_TIMEOUT_MS);
}
```

### 3. Cache partially-received sequences

```typescript
// If the sequence is incomplete but starts like a known sequence,
// check if we can wait for more bytes
private _parseEscapeSequence(buffer: number[]): KeyEvent | null {
    // Try complete match first
    const completeMatch = this._matchEscapeSequence(buffer);
    if (completeMatch) return completeMatch;
    
    // Check if buffer is a PREFIX of any known sequence
    if (this._isPrefixOfKnownSequence(buffer)) {
        return null;  // Wait for more bytes
    }
    
    // Buffer no longer matches any known sequence — it's garbage
    return null;  // Caller should flush
}
```

### 4. Add configuration option for timeout

```typescript
interface InputParserOptions {
    escapeSequenceTimeout?: number;  // ms, default: 0 (wait indefinitely)
    escapeSequenceIdleTimeout?: number;  // ms, default: 50
}

class InputParser {
    constructor(options: InputParserOptions = {}) {
        this._escapeTimeout = options.escapeSequenceTimeout ?? 0;
        this._idleTimeout = options.escapeSequenceIdleTimeout ?? 50;
    }
}
```

### 5. Emit intermediate state for UI feedback

For long sequences, emit a "reading escape sequence" status so the UI can show a loading indicator or ignore intermediate bytes:

```typescript
this._emit({ type: 'escape-sequence-begin' });
```

## Testing

The fix should be tested with:

1. Unit tests for `InputParser` that simulate split chunks
2. Integration test with artificial latency on stdin
3. Manual testing over SSH with varied network conditions
4. Regression tests for all known escape sequences (arrows, F-keys, home/end, page up/down, insert/delete)

ionfwsrijan:
### Changes

**`packages/core/src/input/InputParser.ts`**
- Replaced the fixed 200ms timeout for lone ESC bytes with a new `_startEscapeTimeout()` method that uses a 500ms timeout.
- Added timeout restart in the escape buffer concatenation handler: each partial chunk resets the timer, preventing false standalone ESC emission.
- Cleaned up the lone ESC code path by extracting timeout logic into its own method.

### Fixes
- Escape sequences fragmented across TCP chunks (SSH, serial connections) with >200ms gaps were incorrectly parsed as standalone Escape keypresses followed by garbage characters
- After the initial ESC timeout fired, subsequent sequence bytes were processed as regular input, corrupting the input stream
- 200ms window was too aggressive for high-latency network connections

Closes #1957

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of the Escape key so incomplete or delayed key sequences are buffered more reliably.
  * Restarts the Escape timeout when more input arrives, reducing misread key events.
  * Lone Escape presses now emit consistently after a short delay.

### DIFF
--- a/packages/core/src/input/InputParser.test.ts
+++ b/packages/core/src/input/InputParser.test.ts
@@ -353,13 +353,13 @@ describe('InputParser', () => {
     });
 
     describe('FSM escape sequence handling', () => {
-        it('emits standalone Escape key after 200ms timeout', () => {
+        it('emits standalone Escape key after 500ms timeout', () => {
             const { stdin, handler } = createParser();
             // Lone ESC byte
             originalSendKey(stdin, Buffer.from([0x1b]));
             expect(handler).not.toHaveBeenCalled();
-            // Advance past the 200ms timeout
-            vi.advanceTimersByTime(250);
+            // Advance past the 500ms timeout
+            vi.advanceTimersByTime(550);
             expect(handler).toHaveBeenCalledTimes(1);
             expect(handler).toHaveBeenCalledWith(expect.objectContaining({ key: 'escape' }));
         });
@@ -380,9 +380,9 @@ describe('InputParser', () => {
             const { stdin, handler } = createParser();
             // Simulate ESC arriving alone (as if it was the start of an escape sequence)
             originalSendKey(stdin, Buffer.from([0x1b]));
-            // Advance time past 200ms timeout - should NOT emit yet if buffer was appended
+            // Advance time past 500ms timeout - should NOT emit yet if buffer was appended
             // But in this test, no continuation arrives, so after timeout Escape fires
-            vi.advanceTimersByTime(250);
+            vi.advanceTimersByTime(550);
             expect(handler).toHaveBeenCalledTimes(1);
             expect(handler).toHaveBeenCalledWith(expect.objectContaining({ key: 'escape' }));
         });
--- a/packages/core/src/input/InputParser.ts
+++ b/packages/core/src/input/InputParser.ts
@@ -11,6 +11,8 @@ import { parseMouseEvent, isMouseSequence } from './MouseParser.js';
 import { EventEmitter } from '../events/EventEmitter.js';
 import { splitGraphemes } from './grapheme.js';
 
+const ESCAPE_TIMEOUT_MS = 500;
+
 export interface CursorPosition {
     row: number;
     col: number;
@@ -203,6 +205,9 @@ export class InputParser {
         if (str.startsWith('\x1b')) {
             this._escapeBuffer = data;
             this._tryParseEscape();
+            if (this._escapeBuffer.length > 0) {
+                this._startEscapeTimeout();
+            }
             return;
         }
 
@@ -331,7 +336,7 @@ export class InputParser {
                 alt: false,
                 shift: false,
             }));
-        }, 500);
+        }, ESCAPE_TIMEOUT_MS);
     }
 
     /**
--- a/packages/core/src/input/InputParser.ts
+++ b/packages/core/src/input/InputParser.ts
@@ -158,6 +158,11 @@ export class InputParser {
                 this._escapeTimeout = null;
             }
             this._tryParseEscape();
+            // If the sequence is still incomplete, restart the timeout
+            // so fragmented TCP chunks won't trigger a false standalone ESC.
+            if (this._escapeBuffer.length > 0) {
+                this._startEscapeTimeout();
+            }
             return;
         }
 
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

ionfwsrijan:
@Karanjot786  Please add proper labels to this

PR #1966, opened 2026-07-03
fix(widgets): prevent Pty output drop on early completion

### LINKED ISSUE #1959: [bug] Pty widget silently drops all output when child process completes before handler attachment
## Description

The Pty (terminal multiplexer) widget spawns a child process and then attaches stdout/stderr `data` event handlers. If the spawned process completes synchronously — before the `data` handlers are registered — all child process output is written to the internal Node.js stream buffer and then silently discarded. The widget displays nothing, and the application appears broken.

## Root Cause

**`packages/widgets/src/display/Pty.ts`** (lines ~30-55):

```typescript
constructor(opts: PtyOptions) {
    super();
    
    const cmd = opts.command ?? (process.platform === 'win32' ? 'cmd.exe' : '/bin/sh');
    const args = opts.args ?? [];
    
    // STEP 1: Spawn the process
    this._process = spawn(cmd, args, {
        stdio: ['pipe', 'pipe', 'pipe'],
    });
    
    // STEP 2: Attach output handlers (AFTER spawn)
    this._process.stdout!.on('data', (data: Buffer) => {
        this._handleOutput(data);
    });
    
    this._process.stderr!.on('data', (data: Buffer) => {
        this._handleOutput(data);  // or _handleError(data)
    });
    
    this._process.on('close', (code) => {
        this._handleClose(code);
    });
}
```

**The race condition:**

1. `spawn(cmd, args)` starts the child process
2. For fast commands (e.g., `echo hello`, `date`, `whoami`, `dir`, `ls`), the process can execute, produce output, and exit **before** JavaScript reaches line 40
3. Node.js streams buffer data internally, but the `data` event fires synchronously during the current or next event loop tick
4. Since no `data` listener exists yet, the buffered data is **lost** (the readable stream's internal buffer is consumed but there's no handler)
5. The `close` event fires — also before the handler is registered
6. Handlers are registered at lines 40-50 — but the process has already exited
7. The widget remains empty forever

## Real-World Examples

Commands that commonly exhibit this behavior:

| Command | Typical Runtime | Expected Output | Widget Behavior (Bug) |
|---------|----------------|----------------|----------------------|
| `echo "Hello World"` | <1ms | "Hello World" | Empty |
| `date` | <1ms | Current date | Empty |
| `whoami` | <1ms | Username | Empty |
| `dir /b *.txt` | 1-5ms | File list | Empty |
| `python -c "print(42)"` | 10-50ms | "42" | Empty (sometimes) |
| `node -e "console.log('hi')"` | 20-80ms | "hi" | Empty (sometimes) |
| `git status --porcelain` | 5-100ms | Git status | Empty (sometimes) |

### 1. Use `exec`/`execFile` instead of `spawn` when no interactivity needed

For non-interactive commands, `execFile` captures all output atomically:

```typescript
import { execFile } from 'child_process';

class Pty {
    async runCommand(cmd: string, args: string[]): Promise<string> {
        return new Promise((resolve, reject) => {
            execFile(cmd, args, { maxBuffer: 1024 * 1024 }, (err, stdout, stderr) => {
                if (err) {
                    this._handleError(stderr || err.message);
                    reject(err);
                } else {
                    this._handleOutput(stdout);
                    resolve(stdout);
                }
            });
        });
    }
}
```

### 2. For interactive spawns: use stream buffering before handler attachment

```typescript
constructor(opts: PtyOptions) {
    super();
    
    const cmd = opts.command ?? (process.platform === 'win32' ? 'cmd.exe' : '/bin/sh');
    const args = opts.args ?? [];
    
    // Buffer output BEFORE handlers are attached
    const stdoutChunks: Buffer[] = [];
    const stderrChunks: Buffer[] = [];
    
    const onStdoutData = (data: Buffer) => stdoutChunks.push(data);
    const onStderrData = (data: Buffer) => stderrChunks.push(data);
    const onClose = (code: number | null) => {
        // Remove temp listeners
        this._process?.stdout?.removeListener('data', onStdoutData);
        this._process?.stderr?.removeListener('data', onStderrData);
    };
    
    this._process = spawn(cmd, args, { stdio: ['pipe', 'pipe', 'pipe'] });
    
    // Register temp listeners IMMEDIATELY after spawn
    this._process.stdout!.on('data', onStdoutData);
    this._process.stderr!.on('data', onStderrData);
    this._process.on('close', onClose);
    
    // Use setImmediate to register permanent handlers on next tick
    // By then, any synchronous output has been captured in the buffers
    setImmediate(() => {
        // Remove temp listeners
        this._process?.stdout?.removeListener('data', onStdoutData);
        this._process?.stderr?.removeListener('data', onStderrData);
        
        // Replay buffered output
        for (const chunk of stdoutChunks) {
            this._handleOutput(chunk);
        }
        for (const chunk of stderrChunks) {
            this._handleError?.(chunk);
        }
        
        // Register permanent handlers for future output
        this._process!.stdout!.on('data', (data: Buffer) => {
            this._handleOutput(data);
        });
        this._process!.stderr!.on('data', (data: Buffer) => {
            this._handleError?.(data);
        });
    });
}
```

### 3. Use paused mode on the readable stream

```typescript
this._process = spawn(cmd, args, { stdio: ['pipe', 'pipe', 'pipe'] });

// Pause stdout stream — it will buffer internally until we resume
this._process.stdout!.pause();

// Attach handlers
this._process.stdout!.on('data', (data: Buffer) => {
    this._handleOutput(data);
});

// Resume — any buffered data will be replayed through the handler
this._process.stdout!.resume();
```

When paused, Node.js Readable streams buffer incoming data internally. When resumed, they emit all buffered data as `data` events. This guarantees no data is lost between spawn and handler registration.

### 4. Add a `ready` promise for consumers

```typescript
class Pty extends Widget {
    private _ready: Promise<void>;
    private _readyResolve!: () => void;
    
    constructor(opts: PtyOptions) {
        super();
        this._ready = new Promise(resolve => { this._readyResolve = resolve; });
        
        this._process = spawn(cmd, args);
        
        // Set up handlers...
        
        this._readyResolve();  // Signal that the widget is ready
    }
    
    async waitForReady(): Promise<void> {
        return this._ready;
    }
}
```

ionfwsrijan:
### Changes

**`packages/widgets/src/base/Widget.ts`**
- Added dirty state propagation in `addChild`: when a child widget has `_dirty = true` (accumulated before being added to the tree), the parent's `markDirty()` is called to propagate the dirty signal up to the root.

### Fixes
- Pty widget spawned its process in the constructor; output received before the widget was added to the widget tree called `markDirty()` with `parent = null`, silently dropping the dirty signal
- Root widget never learned about dirty children that accumulated state before tree attachment
- Output from fast-exiting processes was completely lost

Closes #1959

### DIFF
--- a/packages/widgets/src/base/Widget.ts
+++ b/packages/widgets/src/base/Widget.ts
@@ -175,6 +175,11 @@ export abstract class Widget {
     addChild(child: Widget): void {
         child.parent = this;
         this._children.push(child);
+        // Propagate any dirty state the child accumulated before being added
+        // to the tree (e.g., Pty output that arrived before mount).
+        if (child._dirty) {
+            this.markDirty();
+        }
     }
 
     /** Remove a child widget */


ionfwsrijan:

ionfwsrijan:
@Karanjot786  Please add proper labels to this

PR #2037, opened 2026-07-04
fix(store): resolve symlinks in persist file path

### LINKED ISSUE #2032: [bug] Store Persist Path Traversal Protection Can Be Bypassed via Symlinks — Arbitrary File Write
## Description

The store persistence module implements a path traversal check using:

```typescript
const resolved = path.resolve(baseDir, userPath);
const relative = path.relative(baseDir, resolved);
if (relative.startsWith('..')) {
  throw new Error('Path traversal detected');
}
```

While this correctly prevents `../` sequences in the path string from escaping the base directory, it does **not** protect against symlink-based traversal. An attacker who can write a file anywhere on the filesystem can:

1. Create a symlink inside the base directory that points to an arbitrary location
2. Use the persistence mechanism to write through the symlink to the target location
3. Overwrite critical system files, configuration files, or inject malicious code

## Why This Is Critical

- **Arbitrary file write** — The persistence store allows writing serialized state to files. If an attacker can control the store key (which becomes part of the file path), they can write to arbitrary locations via symlinks.
- **Privilege escalation** — If the terminal application runs with elevated permissions (common in system administration tools built with TermUI), this becomes a full privilege escalation vector.
- **Configuration injection** — Writing to `~/.bashrc`, `~/.ssh/authorized_keys`, or system configuration files via symlinks bypasses path traversal checks entirely.
- **The path traversal check provides a false sense of security** — developers relying on this check may assume file writes are safely contained, when in fact they are not.
- Any application using the TermUI store persistence with user-influenced keys is vulnerable.

### 1. Add symlink resolution before path traversal check

Before checking if the resolved path is within bounds, resolve any symlinks in the path:

```typescript
// packages/store/src/store.ts

import fs from 'fs';
import path from 'path';

function safeResolve(baseDir: string, userPath: string): string {
  // First, resolve the user path normally
  const resolved = path.resolve(baseDir, userPath);
  
  // Resolve any symlinks in the path
  let realResolved: string;
  try {
    realResolved = fs.realpathSync(resolved);
  } catch (err: any) {
    if (err.code === 'ENOENT') {
      // File doesn't exist yet — resolve the parent directory to check symlinks
      // in the path components
      const parentDir = path.dirname(resolved);
      try {
        const realParent = fs.realpathSync(parentDir);
        realResolved = path.join(realParent, path.basename(resolved));
      } catch {
        // Parent also doesn't exist — can't resolve further
        realResolved = path.resolve(baseDir, userPath);
      }
    } else {
      // Permission error, etc. — reject the path
      throw new Error(`Cannot resolve path: ${err.message}`);
    }
  }
  
  // Now check the resolved path against the base directory
  const relative = path.relative(baseDir, realResolved);
  if (relative.startsWith('..')) {
    throw new Error(
      `Path traversal detected via symlink: ${userPath} → ${realResolved} ` +
      `(outside ${baseDir})`
    );
  }
  
  return realResolved;
}
```

### 2. Add a write-time check to prevent following symlinks

When writing a file, check if the target path is a symlink and reject the write if it points outside the allowed directory:

```typescript
async function writeStoreFile(filePath: string, data: string): Promise<void> {
  // Check if the path (or any component of it) is a symlink escape
  try {
    const stat = await fs.promises.lstat(filePath);
    if (stat.isSymbolicLink()) {
      const linkTarget = await fs.promises.readlink(filePath);
      const absoluteTarget = path.resolve(path.dirname(filePath), linkTarget);
      throw new Error(
        `Refusing to write through symlink: ${filePath} → ${absoluteTarget}`
      );
    }
  } catch (err: any) {
    if (err.code !== 'ENOENT') throw err;
    // File doesn't exist — that's fine, we'll create it
  }
  
  // Write to a temporary file first, then rename (atomic write)
  const tmpPath = filePath + '.tmp';
  await fs.promises.writeFile(tmpPath, data, 'utf-8');
  await fs.promises.rename(tmpPath, filePath);
}
```

### 3. Add read-time symlink checking

Also check symlinks when reading persisted state:

```typescript
async function readStoreFile(filePath: string): Promise<string | null> {
  try {
    const stat = await fs.promises.lstat(filePath);
    if (stat.isSymbolicLink()) {
      throw new Error('Refusing to read through symlink');
    }
    return await fs.promises.readFile(filePath, 'utf-8');
  } catch (err: any) {
    if (err.code === 'ENOENT') return null;
    throw err;
  }
}
```

### 4. Add a `safeDir` option with path validation at initialization

```typescript
interface PersistOptions {
  /** Base directory for store files */
  baseDir?: string;
  /** If true, validate all paths against symlink attacks */
  safeMode?: boolean;
  /** Maximum number of files the store can create */
  maxFiles?: number;
  /** Maximum size of a single store file in bytes */
  maxFileSize?: number;
}

class Store<T extends State> {
  constructor(config?: { persist?: PersistOptions }) {
    // Validate the base directory itself isn't a symlink
    if (config?.persist?.safeMode !== false) {
      const baseDir = config?.persist?.baseDir || defaultBaseDir;
      try {
        const realBaseDir = fs.realpathSync(baseDir);
        if (realBaseDir !== path.resolve(baseDir)) {
          throw new Error(
            `Store base directory is a symlink: ${baseDir} → ${realBaseDir}`
          );
        }
      } catch (err: any) {
        if (err.code !== 'ENOENT') throw err;
      }
    }
    // ...
  }
}
```

### 5. Add a comprehensive security audit utility

```typescript
// packages/store/src/security.ts

import fs from 'fs';
import path from 'path';

export interface FileAccessPolicy {
  /** Base directory for all file operations */
  baseDir: string;
  /** Whether to resolve symlinks before access checks */
  resolveSymlinks: boolean;
  /** Maximum file size for reads */
  maxReadSize: number;
  /** File extensions allowed for writes */
  allowedExtensions?: string[];
}

export function checkFileAccess(policy: FileAccessPolicy, targetPath: string): void {
  const resolved = path.resolve(policy.baseDir, targetPath);
  
  // Check extension whitelist
  if (policy.allowedExtensions) {
    const ext = path.extname(resolved).toLowerCase();
    if (!policy.allowedExtensions.includes(ext)) {
      throw new Error(`File extension '${ext}' is not allowed for store operations`);
    }
  }
  
  // Resolve symlinks if configured
  let checkedPath = resolved;
  if (policy.resolveSymlinks) {
    checkedPath = fs.realpathSync(resolved);
  }
  
  // Verify path is within base directory
  const relative = path.relative(policy.baseDir, checkedPath);
  if (relative.startsWith('..')) {
    throw new Error(`Access denied: ${targetPath} is outside the store directory`);
  }
}
```

## Files to Modify

1. `packages/store/src/store.ts` — Add symlink-aware path resolution in persist/restore
2. `packages/store/src/security.ts` — New file with file access policy utilities
3. `packages/store/src/index.ts` — Export new security utilities

## Testing

- **Unit test:** Create a symlink inside the store directory pointing to `/etc/passwd`; verify the store refuses to write through it
- **Unit test:** Create a chain of symlinks; verify all are resolved before the traversal check
- **Unit test:** Verify legitimate file writes (no symlinks, within base dir) still work
- **Unit test:** Verify that symlinks in path components (not just the final file) are also resolved
- **Integration test:** Set up a store with persistence, write through a symlink, verify error is thrown

ionfwsrijan:
## Summary

Resolves symlinks in the persist file path when writing store state to disk, preventing writes from accidentally going through a symlink target (e.g., if `~/.config` or `~/AppData/Roaming` is a symlink). Walks up the filesystem tree to find the first existing ancestor directory, resolves it with `fs.realpathSync`, and reconstructs the path.

## Changes

- `packages/store/src/store.ts`: Added symlink resolution block in persist path setup (wrapped in try/catch with fallback)
- `packages/store/src/store.test.ts`: 2 new tests for symlink resolution and rehydration

Closes #2032

### DIFF
--- a/packages/store/src/store.test.ts
+++ b/packages/store/src/store.test.ts
@@ -867,3 +867,57 @@ describe('useStore selector memoization', () => {
     })
 })
 
+describe('persist – symlink resolution', () => {
+    const testRoot = fs.realpathSync(fs.mkdtempSync(path.join(os.tmpdir(), 'termuijs-symtest-')));
+    const storeDir = path.join(testRoot, 'termuijs-stores');
+    const realDir = path.join(storeDir, 'real');
+    const linkDir = path.join(storeDir, 'link');
+
+    afterAll(() => {
+        try { fs.rmSync(storeDir, { recursive: true }); } catch { /* ok */ }
+        try { fs.rmdirSync(testRoot); } catch { /* ok */ }
+    });
+
+    const origAppData = process.env.APPDATA;
+    const origXdgConfig = process.env.XDG_CONFIG_HOME;
+    const symlinkType: any = process.platform === 'win32' ? 'junction' : 'dir';
+
+    beforeEach(() => {
+        process.env.APPDATA = testRoot;
+        process.env.XDG_CONFIG_HOME = testRoot;
+        fs.mkdirSync(realDir, { recursive: true });
+        try { fs.rmdirSync(linkDir); } catch { /* ok */ }
+        if (!fs.existsSync(linkDir)) {
+            fs.symlinkSync(realDir, linkDir, symlinkType);
+        }
+    });
+
+    afterEach(() => {
+        process.env.APPDATA = origAppData;
+        process.env.XDG_CONFIG_HOME = origXdgConfig;
+        vi.useRealTimers();
+        try { fs.rmSync(storeDir, { recursive: true }); } catch { /* ok */ }
+    });
+
+    it('resolves symlinks in persist path and writes to real directory', () => {
+        vi.useFakeTimers();
+        const useStore = createStore({ count: 0 }, {
+            persist: { file: path.join(linkDir, 'symtest.json') },
+        });
+        useStore.setState({ count: 42 });
+        vi.advanceTimersByTime(200);
+        const realFilePath = path.join(realDir, 'symtest.json');
+        expect(fs.existsSync(realFilePath)).toBe(true);
+        const content = JSON.parse(fs.readFileSync(realFilePath, 'utf8'));
+        expect(content.count).toBe(42);
+    });
+
+    it('rehydrates from the resolved real path on restart', () => {
+        fs.writeFileSync(path.join(realDir, 'symtest.json'), JSON.stringify({ count: 99 }));
+        const useStore = createStore({ count: 0 }, {
+            persist: { file: path.join(linkDir, 'symtest.json') },
+        });
+        expect(useStore.getState().count).toBe(99);
+    });
+})
+
--- a/packages/store/src/store.ts
+++ b/packages/store/src/store.ts
@@ -323,6 +323,26 @@ export function createStore<T extends object>(
             }
             persistFilePath = path.join(persistDir, `${persistOpt.key}.json`);
         }
+        // Resolve symlinks in the persist directory to prevent writes through
+        // symlink targets (e.g., ~/.config or ~/AppData/Roaming being a symlink).
+        // Walk up from the file to find the first existing ancestor, resolve it,
+        // and reconstruct the remainder.
+        try {
+            const dir = path.dirname(persistFilePath);
+            if (fs.existsSync(dir)) {
+                persistFilePath = path.join(fs.realpathSync(dir), path.basename(persistFilePath));
+            } else {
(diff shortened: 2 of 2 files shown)


ionfwsrijan:

PR #2063, opened 2026-07-05
fix(core): prevent _pendingFocusState memory leak by removing stale entries after max retries (#2059)

### LINKED ISSUE #2059: [bug] `_pendingFocusState` Memory Leak and Focus State Drift in Core App Loop
## Description

The `_pendingFocusState` Map (line 91) stores focus events that arrive from `FocusManager` before the widget ID map is rebuilt via `_buildWidgetMap()`. When a focus arrives for a widget whose ID has not yet been registered, it's stored in `_pendingFocusState` until the next map rebuild.

The drain function `_applyPendingFocusState()` (lines 594–603) only deletes entries on **successful** application. If a focus event is queued for a widget ID that was **already destroyed or never registered** (e.g., a component unmounts before its first render), `_setWidgetFocused()` returns `null` (widget not found), and the entry is **never removed** from `_pendingFocusState`.

Over hours of usage with dynamic UIs (modals, wizards, dynamic lists, chat interfaces), this map grows unboundedly. Additionally, if a component remounts with the same ID as a stale pending entry, it receives a stale focus state from a previous lifecycle — a focus state drift bug.

## Code

```typescript
// App.ts line 557-564
private _handleFocusEvent(event: FocusEvent): void {
    const changed = this._setWidgetFocused(event.targetId, event.type === 'focus');
    if (changed === null) {
        // Widget not yet registered — queue for later
        this._pendingFocusState.set(event.targetId, focused);
        return;  // stored but never cleaned up if widget never appears
    }
}
```

```typescript
// App.ts line 594-603
private _applyPendingFocusState(): void {
    for (const [id, focused] of this._pendingFocusState) {
        const stateChanged = this._setWidgetFocused(id, focused);
        if (stateChanged !== null) {
            this._pendingFocusState.delete(id);  // Only deletes on SUCCESS
        }
        // If stateChanged === null, entry STAYS forever
    }
}
```

## Impact

- `_pendingFocusState` grows without bound in long-running applications with dynamic UIs
- Each entry holds the widget ID string and a boolean — memory impact is small per entry but the leak is unbounded
- If a destroyed widget's ID is reused by a new widget, the stale pending entry applies a focus state from a previous lifecycle — incorrect visual rendering
- Applications running for hours (dashboards, monitors, AI assistants) gradually degrade in performance and behavior

## Fix

Change `_applyPendingFocusState` to delete entries that return `null` after a max retry count or on widget destruction. The simplest fix adds a retry counter:

```typescript
private _applyPendingFocusState(): void {
    for (const [id, focused] of this._pendingFocusState) {
        const stateChanged = this._setWidgetFocused(id, focused);
        if (stateChanged !== null) {
            this._pendingFocusState.delete(id);
        }
        // After max retries, remove stale entries
        const existing = this._pendingFocusRetries.get(id) ?? 0;
        if (existing > 3) {
            this._pendingFocusState.delete(id);
            this._pendingFocusRetries.delete(id);
        } else {
            this._pendingFocusRetries.set(id, existing + 1);
        }
    }
}
```

Alternatively, clear all pending entries when the widget map is rebuilt.

ionfwsrijan:
## Description

Fixes #2059

`_pendingFocusState` entries were never removed if the widget was destroyed or never
registered, causing unbounded memory growth in long-running applications.

### Changes (`packages/core/src/app/App.ts`)
1. Added `_pendingFocusRetries` Map to track retry count per pending entry
2. Added `PENDING_FOCUS_MAX_RETRIES = 5` static constant
3. In `_applyPendingFocusState()`: entries that return `null` (widget not found) are
   deleted after exceeding max retries, preventing stale entries from accumulating

### Impact
- Prevents unbounded `_pendingFocusState` Map growth in dynamic UIs
- Eliminates stale focus state drift when widgets are destroyed and their IDs reused

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved focus handling so pending focus changes are retried a limited number of times instead of being retried indefinitely.
  * Focus state is now cleared more reliably once a widget is found or when retries are exhausted, helping prevent stale focus behavior.

### DIFF
--- a/packages/core/src/app/App.ts
+++ b/packages/core/src/app/App.ts
@@ -89,6 +89,8 @@ export class App {
     private _widgetById = new Map<string, any>(); // any: Widget shape varies; narrowed at retrieval
     private _hoveredWidgetId: string | null = null;
     private _pendingFocusState = new Map<string, boolean>();
+    private _pendingFocusRetries = new Map<string, number>();
+    private static readonly PENDING_FOCUS_MAX_RETRIES = 5;
 
     private _consecutiveRenderFailures = 0;
     private static readonly MAX_RENDER_FAILURES = 5;
@@ -598,6 +600,15 @@ export class App {
             const stateChanged = this._setWidgetFocused(id, focused);
             if (stateChanged !== null) {
                 this._pendingFocusState.delete(id);
+                this._pendingFocusRetries.delete(id);
+            } else {
+                const retries = this._pendingFocusRetries.get(id) ?? 0;
+                if (retries >= App.PENDING_FOCUS_MAX_RETRIES) {
+                    this._pendingFocusState.delete(id);
+                    this._pendingFocusRetries.delete(id);
+                } else {
+                    this._pendingFocusRetries.set(id, retries + 1);
+                }
             }
         }
     }


ionfwsrijan:

PR #2137, opened 2026-07-07
fix: cap spring physics dt to MAX_DT to prevent timer-pool subscription leak after process suspend

### LINKED ISSUE #2010: [Critical] animateSpring() uses unbounded wall-clock dt — tab suspend or system sleep causes spring overshoot and infinite oscillation, leaking timer-pool subscription permanently
packages/motion/src/spring.ts computes `dt` in `animateSpring()` as raw wall-clock delta:

```ts
const now = Date.now();
const dt = (now - lastTime) / 1000; // ← unbounded
lastTime = now;
state = stepSpring(state, cfg, dt);
```

When the Node.js process is suspended — terminal backgrounded, system sleep, CI runner paused — `dt` can be arbitrarily large (seconds or minutes). Feeding a large `dt` into the semi-implicit Euler integrator in `stepSpring` causes:

### Problem 1 — Overshoot

With `dt = 2.0s`, a spring with `tension = 180` and `friction = 12` (the `wobbly` preset) accumulates an acceleration of `180 × displacement` over a single 2-second integration step, launching `newValue` far past the target. `state.done` becomes `false` again and the animation continues.

### Problem 2 — Infinite oscillation

Once overshot, successive ticks oscillate with growing amplitude because the underdamped `wobbly` preset (`friction = 12`) cannot dissipate the energy injected by the oversized integration step. The spring never settles.

### Problem 3 — Timer-pool leak

Since `state.done` never becomes `true`, `unsub()` is never called. The `16ms` `setInterval` registered with the timer pool runs indefinitely, repeatedly invoking the `onFrame` callback with values oscillating around the target.

In applications with many concurrent springs (progress bars, animated charts, widgets), every suspend/resume cycle can leave **N permanent timers** running.

### Reproduction

```ts
const cleanup = animateSpring(0, 100, 'wobbly', (v) => console.log(v));

// Simulate a 5-second suspend by advancing lastTime.
// The next tick runs with dt = 5.0s,
// causing the spring to overshoot and never settle.
```

## Expected behavior

Clamp `dt` to a maximum safe integration step:

```ts
const MAX_DT = 1 / 30; // Never advance physics more than one 30 FPS frame
const dt = Math.min((now - lastTime) / 1000, MAX_DT);
```

Capping `dt` is standard practice in real-time physics loops. After a suspend/resume event, the animation skips visual frames instead of injecting unbounded kinetic energy into the spring.

Additionally, add a safety convergence check in `stepSpring`:

```ts
if (
  Math.abs(newValue - state.target) < precision &&
  Math.abs(newVelocity) < precision * 10
) {
  return {
    value: state.target,
    velocity: 0,
    target: state.target,
    done: true,
  };
}
```

This guarantees convergence once the spring is sufficiently close to the target, preventing pathological oscillation due to numerical integration.

---

I'd like to work on this issue under **GSSoC 2026**. If this approach looks good after triage, please assign it to me.

711nishtha:
## Problem
`animateSpring()` used unbounded wall-clock `dt`. After any process
suspend, system sleep, or tab backgrounding, the next timer tick fired
with `dt` up to minutes. Underdamped presets (wobbly, gentle) would
overshoot the target and oscillate indefinitely — `state.done` never
became `true`, so `unsub()` was never called. The 16ms setInterval
ran forever.

## Fix
- `MAX_DT = 1/30` constant caps the dt passed to `stepSpring()`
- Safety clamp in `stepSpring()` snaps to target when within 10× precision
- Both changes are independent — callers using `stepSpring()` directly
  are also protected

## Files
- `packages/motion/src/spring.ts`
- `packages/motion/src/spring.test.ts` (5 new regression tests)

Closes #2010 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved spring animations to remain stable after long pauses or large time gaps by capping the maximum time step.
  * Enhanced spring settling so animations reliably snap to the exact target (no end-state drift) and mark completion sooner when effectively settled.
  * Improved `animateSpring` cancel behavior in the `caps.motion=false` scenario so it’s safe to call.
* **Tests**
  * Added regression coverage for large/unbounded time steps, exact snapping/completion behavior, and stability improvements.

### DIFF
--- a/packages/motion/src/spring.test.ts
+++ b/packages/motion/src/spring.test.ts
@@ -3,7 +3,7 @@
 // ─────────────────────────────────────────────────────
 
 import { describe, it, expect, vi, beforeEach } from 'vitest';
-import { stepSpring, SPRING_PRESETS } from './spring.js';
+import { stepSpring, animateSpring, SPRING_PRESETS, MAX_DT } from './spring.js';
 import type { SpringState } from './spring.js';
 
 describe('stepSpring', () => {
@@ -33,7 +33,7 @@ describe('stepSpring', () => {
     });
 
     it('stiff preset settles faster than gentle', () => {
-        let stiffState: SpringState = { value: 0, velocity: 0, target: 1, done: false };
+        let stiffState: SpringState  = { value: 0, velocity: 0, target: 1, done: false };
         let gentleState: SpringState = { value: 0, velocity: 0, target: 1, done: false };
         let stiffSteps = 0, gentleSteps = 0;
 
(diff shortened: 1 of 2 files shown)


711nishtha:
@Karanjot786 The PR is ready for review and merge. Thank you

PR #2144, opened 2026-07-07
fix(core): stop monkey-patching console in RenderHook, call through instead

### LINKED ISSUE #2141: [bug] RenderHook.start() Destructively Monkey-Patches Global Console, Breaking Debugging Tools and Source Maps
## Description

`RenderHook.start()` in `packages/core/src/renderer/render-hook.ts:18-31` replaces `console.log`, `console.warn`, and `console.error` with wrapper functions that buffer output for the terminal UI. This global monkey-patch approach has several severe side effects:

1. **Breaks source maps and stack traces** — The wrapper functions have no source mapping to the original call site. Any error logged while the hook is active shows the wrapper as the origin in stack traces, making debugging extremely difficult.

2. **Intercepts ALL console output** — Including logs from third-party libraries, Node.js internals, and Promise rejection handlers. These logs get silently buffered and may never be flushed if the hook is active during a crash.

3. **Loses `console` method binding** — The wrapper functions are detached from the `console` object. Code that uses `Reflect.apply(console.log, console, args)` or stores `console.log` references before the hook starts will get inconsistent behavior.

4. **No protection against concurrent hooks** — If another part of the application also monkey-patches console (e.g., a test framework like Vitest, or a logging library), the hooks will step on each other, with one silently overwriting the other's saved originals.

5. **Memory leak on crash** — If `stop()` is never called (due to unhandled error or process crash), the original console references in `_originalConsole` are lost, and even after process restart the originals cannot be recovered.

## Affected Files

- `packages/core/src/renderer/render-hook.ts` (lines 18–50) — the `start()`, `stop()`, and `flush()` methods
- `packages/core/src/terminal/Renderer.ts` — uses `RenderHook` during render cycle
- `packages/core/src/app/App.ts` — initializes and manages the render hook

## Expected Behavior

The render hook should capture terminal output intended for the terminal UI without interfering with:
- Developer debugging tools (source maps, stack traces, breakpoints)
- Third-party library logging
- Node.js internal warnings and errors
- Test framework console interception

A more robust approach would intercept at the **output stream level** (wrapping `process.stdout.write`) rather than at the `console` API level, or use a custom `Console` instance for widget-authored logging.

## Actual Behavior

Current implementation (`render-hook.ts:18-31`):

```typescript
start(): void {
    if (this._isActive) return;
    this._isActive = true;

    const methods: ConsoleMethod[] = ['log', 'warn', 'error'];
    for (const method of methods) {
        this._originalConsole[method] = console[method];
        const hook = this;
        console[method] = function (...args: any[]): void {
            const text = args.map(a => typeof a === 'string' ? a : String(a)).join(' ');
            hook._buffer.push(text + '\n');
        };
    }
}
```

The problems:

1. The replacement function uses `...args: any[]` spread, which destroys the JavaScript engine's ability to optimize the call and loses the `this` binding of the original `console`.
2. `hook._buffer.push(...)` is unbounded — during rendering, if many console.log calls are made, memory grows without limit.
3. The function signature `(...args: any[]) => void` differs from the original console methods in `[[Call]]` behavior, which can cause issues with `Function.prototype.toString`, `Proxy` traps, and `new Function()` detection.
4. The `any[]` type annotation hides the actual console method signatures, breaking TypeScript's ability to detect invalid console calls.

## Proposed Fix

**Approach A — Intercept at the stream level (recommended):**

Instead of monkey-patching `console`, wrap `process.stdout.write` to capture terminal output while letting `console` methods work normally:

```typescript
start(): void {
    if (this._isActive) return;
    this._isActive = true;

    this._originalWrite = process.stdout.write.bind(process.stdout);
    const hook = this;
    process.stdout.write = function(chunk: Uint8Array | string, ...args: any[]): boolean {
        if (typeof chunk === 'string' && hook._isActive) {
            hook._buffer.push(chunk);
            return true;
        }
        return hook._originalWrite!(chunk, ...args);
    };
}

stop(): void {
    if (!this._isActive) return;
    this._isActive = false;
    if (this._originalWrite) {
        process.stdout.write = this._originalWrite;
        this._originalWrite = undefined;
    }
}
```

Benefits: `console.log/warn/error` continue to work normally, stack traces are preserved, source maps work, third-party logging is unaffected, and the hook only captures bytes written to stdout (which is what ultimately reaches the terminal).

**Approach B — Use a custom `Console` constructor:**

Create a custom logger that widgets opt into, rather than hijacking the global `console`:

```typescript
import { Console } from 'node:console';

export class WidgetLogger {
    private _console: Console;

    constructor(bufferStream: { write: (chunk: string) => void }) {
        this._console = new Console(bufferStream);
    }

    log(...args: any[]): void { this._console.log(...args); }
    warn(...args: any[]): void { this._console.warn(...args); }
    error(...args: any[]): void { this._console.error(...args); }
}
```

Then provide the logger to widgets via props or context, and document that `console.log` should not be used for widget output.

**Additionally:**

1. Add a **capture size limit** to the buffer (e.g., 1 MB) to prevent unbounded memory growth.
2. Add a **re-entrancy guard** to prevent the hook from capturing its own output (if `process.stdout.write` is called within the wrapped version).
3. Add **tests** that verify:
   - Source maps and stack traces work while the hook is active.
   - Third-party `console.log` calls are not buffered (if using stream approach).
   - The hook can be stopped and started multiple times without leaking references.

ionfwsrijan:
## Problem

`RenderHook.start()` monkey-patches `console.log`, `console.warn`, and `console.error` by replacing them with wrapper functions that discard the original. This causes three issues:

1. **Fragile** — Multiple libraries patching the same console methods leads to conflicts where one library's wrapper overwrites another's
2. **Lossy** — Once captured, console output is never passed to the original function, so the user never sees real console output during development
3. **Incomplete** — `console.debug`, `console.trace`, `console.info` are not captured, so they bypass buffering and leak to stdout during rendering

## Root Cause

The original implementation saved `console[method]` into `_originalConsole`, then replaced it with a wrapper that only pushed to the buffer — never calling the saved original.

## Changes

- **Call through to original**: After buffering, each wrapper now calls `original.apply(console, args)` so console output passes through to stdout and remains visible
- **Extended coverage**: Patch 6 methods instead of 3: `log`, `warn`, `error`, `debug`, `trace`, `info`
- **Cleaner storage**: Replaced `Partial<Record<...>>` object with `Map<ConsoleMethod, Function>` for safer iteration
- **New test**: Verifies the original `console.log` receives the call arguments

## Testing

- All 535 core package tests pass (1 new test)
- Console output is now visible during development while still being captured for rendering

Closes #2141

## Summary by CodeRabbit

* **Bug Fixes**
  * Console output capture now supports `debug`, `trace`, and `info` in addition to existing methods.
  * Original console behavior is reliably restored when capture stops.
  * Console messages remain visible while also being available through buffered output.


ionfwsrijan:

PR #2150, opened 2026-07-07
fix(core): include full color value in getStyleLine hash to avoid collision

### LINKED ISSUE #1659: [bug] Screen.setStyleLine hash collision silently skips color code updates

atul-upadhyay-7:
## Description

The `Screen.getStyleLine` method computed a hash using only the first character of the color type string (`fg.charCodeAt(0)`), causing hash collisions when colors changed within the same type (e.g., ANSI256 code 1 → code 9).

## Fix

Added a `colorSeed` function that computes a unique numeric value for each `Color` variant, including the actual distinguishing properties:

- **named**: hashes the full name string
- **ansi256**: uses the numeric code
- **rgb**: encodes r/g/b values
- **hex**: parses the hex string to a number
- **none**: returns 0

## Tests added

- `getStyleLine produces different hashes for different color codes within same type`
- `getStyleLine distinguishes different named colors`
- `getStyleLine distinguishes different RGB colors`
- `saveLines and getStyleLine detect style-only updates`

Closes #1659

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved style change detection so text updates are recognized even when only the color changes.
  * Fixed style matching to better distinguish colors of the same type, including named colors, 256-color values, and RGB values.

* **Tests**
  * Added coverage for style hash behavior and previous-line tracking to verify color-only changes are detected correctly.

### DIFF
--- a/packages/core/src/terminal/Screen.test.ts
+++ b/packages/core/src/terminal/Screen.test.ts
@@ -201,6 +201,42 @@ describe('Screen', () => {
         screen.popTranslateY();
     });
 
+    it('getStyleLine produces different hashes for different color codes within same type', () => {
+        const screen = new Screen(40, 5);
+        screen.setCell(0, 0, { fg: { type: 'ansi256', code: 1 } });
+        const hash1 = screen.getStyleLine(0);
+        screen.setCell(0, 0, { fg: { type: 'ansi256', code: 9 } });
+        const hash2 = screen.getStyleLine(0);
+        expect(hash1).not.toBe(hash2);
+    });
+
+    it('getStyleLine distinguishes different named colors', () => {
+        const screen = new Screen(40, 5);
+        screen.setCell(0, 0, { fg: { type: 'named', name: 'red' } });
+        const hash1 = screen.getStyleLine(0);
+        screen.setCell(0, 0, { fg: { type: 'named', name: 'blue' } });
+        const hash2 = screen.getStyleLine(0);
+        expect(hash1).not.toBe(hash2);
+    });
+
+    it('getStyleLine distinguishes different RGB colors', () => {
+        const screen = new Screen(40, 5);
+        screen.setCell(0, 0, { fg: { type: 'rgb', r: 255, g: 0, b: 0 } });
+        const hash1 = screen.getStyleLine(0);
+        screen.setCell(0, 0, { fg: { type: 'rgb', r: 0, g: 0, b: 255 } });
+        const hash2 = screen.getStyleLine(0);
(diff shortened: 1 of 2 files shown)


Karanjot786:
**Closing as a duplicate.**

This fixes the same `getStyleLine`/style-fingerprint color-collision bug as #1807, which has now been merged with an O(1) allocation-free per-cell color hash (this PR's version regressed the render bench by ~32% due to a per-character hex loop in the hot path). Thanks for the contribution — the fix lives in #1807.

PR #2154, opened 2026-07-07
fix(core): reset flushEpoch on render error so renderer retries instead of freezing

### LINKED ISSUE #2152: [bug] Renderer._flush() exception permanently freezes all future rendering
## Description

If any exception is thrown during `Renderer._flush()` (e.g., a widget's `render()` method throws, `writeSync` fails, etc.), the renderer permanently freezes. No frame will ever render again, and the entire terminal UI becomes unresponsive with zero feedback.

## Root Cause

In `packages/core/src/terminal/Renderer.ts`, the `_flush()` method sets `flushEpoch = epoch` at line 123 as a re-entrancy guard. The actual `swap()` call (which increments `_epoch`) is at lines 161/186 — the very last operation in the try block.

If an exception is thrown during rendering (lines 139-186), the catch handler at lines 187-192 runs but does NOT reset `flushEpoch`. Since `swap()` was never called, `_epoch` never increments. On every subsequent frame tick:

1. `epoch = this._screen.epoch` — still the old value since swap never ran
2. `this._screen.flushEpoch === epoch` → `true` → returns immediately!

The renderer is permanently frozen. The catch handler also silently swallows the error with no logging, making debugging impossible.

## Impact

- Terminal UI completely freezes on the first rendering error
- The user sees a frozen screen with zero feedback
- The only recovery is killing and restarting the process
- Silent failure with no logging makes root cause impossible to diagnose

## Steps to Reproduce

Any widget that throws during render() will trigger this. For example, accessing a null property inside a render path.

## Suggested Fix

Reset `flushEpoch` in the catch handler so the next frame tick retries:

```typescript
} catch (_err) {
    this._renderRequested = true;
    this._lastStyleFingerprint = null;
    // Reset the re-entrancy gate so the next frame tick retries rendering
    this._screen.flushEpoch = -1;
}
```

### LINKED ISSUE #2153: [bug] Dev-server SIGTERM kills child mid-restore; init error leaves terminal in raw mode with no signal handlers
## Description

The dev-server kills the child app process with `SIGTERM` instead of `SIGINT`. This can prevent the child's terminal cleanup handlers from completing, leaving the terminal in an unrecoverable broken state (raw mode, hidden cursor, alt screen buffer active).

## Root Cause

There are two compounding issues:

**Issue A - Wrong signal in dev-server**: In `packages/dev-server/src/server.ts`, the `_killChild()` method sends `SIGTERM` (line 300) instead of `SIGINT`. When Ctrl+C is pressed:

1. The parent dev-server's SIGINT handler fires and calls `server.stop()` which calls `_killChild()` sending `SIGTERM`
2. The child app's `SIGINT` handler (in `App.ts`) and `SIGTERM` handler (also in `App.ts`) both try to restore the terminal
3. If the child is actively rendering (writing ANSI sequences to stdout) when SIGTERM arrives, the process can be killed mid-write, leaving the terminal with a partial escape sequence in an unpredictable state

**Issue B - Error before signal registration in App.mount()**: In `packages/core/src/app/App.ts`, `enterRawMode()` is called at line 160, but signal handlers for SIGINT/SIGTERM are not registered until lines 263-264. If any operation between lines 160-263 throws (e.g., `renderer.hook.start()` at line 157, `this.input.start()` at line 187), the terminal is left in raw mode with hidden cursor and NO signal handlers to restore it on exit.

## Impact

- Terminal left in raw mode with no echo and garbled display
- User must run `reset` or close the terminal window to recover
- Can cause data loss if the user was running shell commands before the app
- Particularly bad in CI/automation scenarios where the terminal state is important

## Suggested Fix

1. In `packages/dev-server/src/server.ts` line 300: Change `child.kill('SIGTERM')` to `child.kill('SIGINT')` so the child's restore handler runs cleanly
2. In `packages/core/src/app/App.ts`: Register the uncaughtException and unhandledRejection handlers BEFORE entering raw mode, or add a try/catch guard that restores the terminal on initialization failure

atul-upadhyay-7:
When `_flush()` throws before `swap()` completes, the epoch never advances, so `flushEpoch` stays equal to the current epoch and all future flushes are silently skipped — the UI freezes permanently.

**Fix:** Reset `flushEpoch` to `-1` in the catch handler so the next frame tick retries rendering instead of short-circuiting.

Also fixes dev-server sending `SIGTERM` instead of `SIGINT` to the child process, which could prevent terminal cleanup handlers from running before exit.

- Closes #2152
- Closes #2153

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved terminal rendering recovery after a write or flush error.
  - Subsequent rendering attempts now continue normally instead of remaining permanently frozen.
  - Added coverage to verify that rendering resumes and reports frame statistics after an error.


Karanjot786:
Removed unrelated changes from this branch to keep the diff focused:
- `packages/store/src/store.ts` + `store.test.ts` — reverted to match `main`; this duplicated the commit/rollback-closure fix already merged in #2151.
- `packages/dev-server/src/server.ts` (+ its test) — reverted the unrelated `SIGTERM`→`SIGINT` signal change to match `main`.

Kept: the `Renderer.ts` `_flush()` fix that resets `flushEpoch = -1` on render error so the renderer retries instead of freezing, plus its test. Verified `packages/core` builds and `Renderer.test.ts` passes (18/18).

Karanjot786:
Closing this one. The fix here — resetting `flushEpoch` on a render error so the renderer retries instead of freezing — is already on `main`: `packages/core/src/terminal/Renderer.ts` resets `this._screen.flushEpoch = -1;` in the `_flush()` catch block (landed via #2766/#2788). So this PR is now a no-op against current main. Thanks for surfacing the bug — it is fixed.

PR #2473, opened 2026-07-13
fix(core): remove _emitting guard to allow re-entrant emit() calls

### LINKED ISSUE #2464: [bug] EventEmitter Silently Drops Re-entrant `emit()` Calls for the Same Event
## Description

The `EventEmitter` class in `packages/core/src/events/EventEmitter.ts` uses a `_emitting` Set (line 12) to guard against concurrent modification of the handler list during iteration. At line 83, if `this._emitting.has(event)` is true, the entire regular handler block is skipped. This means if Handler A, while executing, calls `emit('X', ...)` for the same event `'X'`, the inner emission is silently dropped — no error, no warning, no queued retry.

The guard was intended to prevent iterator invalidation when handlers modify the handler set during iteration, but the snapshot at line 87 (`for (const handler of [...handlers])`) already handles this case. The `_emitting` guard is overly aggressive and blocks legitimate re-entrant patterns.

**Affected Real-World Pattern — FocusManager (`packages/core/src/events/FocusManager.ts:472-496`):**

`_changeFocus()` emits `'blur'` then `'focus'` sequentially. If a `'blur'` handler calls `clearFocus()` or `focusWidget()`, the nested `_changeFocus()` call emits `'blur'`/`'focus'` events that are silently dropped, leaving the focus trap stack in an inconsistent state.

## Expected Behavior

Re-entrant `emit()` calls for the same event should either:
- Fire on a fresh snapshot of handlers (allowing re-entrant patterns), or
- Be queued and flushed after the current emission completes (like Node.js EventEmitter)

At minimum, re-entrant calls should not be silently discarded.

## Actual Behavior

Re-entrant `emit()` calls for the same event are silently swallowed. The `_emitting` guard at line 83 causes the entire handler iteration block to be skipped when a handler re-emits the same event. No error is thrown, no callback fires, making diagnosis extremely difficult.

## Impact

**High.** This affects:
1. **FocusManager**: Cascading focus changes (blur → focus → blur) leave focus state corrupted.
2. **Any cascading event pattern**: State middleware, event bubbling, or handler-triggered re-emission patterns silently lose events.
3. **Silent failure**: No error or callback fires, making debugging nearly impossible. Developers may spend hours debugging focus issues, event ordering, or missing state updates without seeing any error.

## Proposed Fix

**File:** `packages/core/src/events/EventEmitter.ts`

The `[...handlers]` snapshot at line 87 already protects against concurrent modification of the handler Set during iteration. The `_emitting` guard can be removed entirely, allowing re-entrant emissions to fire on the snapshot.

**Changes:**

Replace lines 82–94:

```typescript
// BEFORE (buggy):
if (!this._emitting.has(event)) {
    this._emitting.add(event);
    const handlers = this._handlers.get(event);
    if (handlers) {
        for (const handler of [...handlers]) {
            try { handler(data); } catch (err) {
                this.onError?.(event, err);
            }
        }
    }
    this._emitting.delete(event);
}

// AFTER (fixed):
const handlers = this._handlers.get(event);
if (handlers) {
    for (const handler of [...handlers]) {
        try { handler(data); } catch (err) {
            this.onError?.(event, err);
        }
    }
}
```

Also remove the `_emitting` field from line 12:
```typescript
// Remove: private _emitting: Set<keyof TEventMap> = new Set();
```

The snapshot (`[...handlers]`) already provides iterator invalidation protection. Re-entrant calls will iterate a new snapshot of whatever handlers are registered at that point, which is the correct semantic.

ionfwsrijan:
## Summary
Fixes #2464 — The `_emitting` guard in `EventEmitter.emit()` prevented a
listener from emitting the same or another event during handler execution.
This caused silent drops when events triggered cascading emissions.

## Changes
- `packages/core/src/events/EventEmitter.ts`: Removed the `_emitting` boolean
  guard and the early-return check, allowing `emit()` to be called re-entrantly
  from within a listener.

## Testing
- Verified that emitting event A from a listener on event B works correctly.
- Verified that removing a listener mid-emit does not crash.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of regular event listeners when events are triggered re-entrantly or during nested emits, ensuring they run at the correct times.
  * Once-only listeners still execute a single time as expected, including when a re-entrant emit happens.


ionfwsrijan:

ionfwsrijan:

PR #2474, opened 2026-07-13
fix(router): preserve forward stack on guard redirect during forward()

### LINKED ISSUE #2465: [bug] Router `forward()` Guard Redirect Wipes Entire Forward Stack
## Description

When a route guard's `beforeEnter` returns a redirect string during `forward()`, the code at `packages/router/src/router.ts:435-438` calls `this.push(guardResult)`. The `push()` method always passes `clearForwardStack: true` to `_executeNavigation()` (line 340). This wipes the entire `_forwardStack`, destroying all entries below the current one.

**Reproduction:**
```
1. Navigate: push('/a') → push('/b') → push('/c')
   History: ['/a', '/b', '/c'], Forward: []
2. back() × 2 → History: ['/a'], Forward: ['/b', '/c']
3. forward() on '/b' which has a guard redirecting to '/login'
   → forward() calls this.push('/login')
   → push() calls _executeNavigation('/login', { clearForwardStack: true })
   → Forward stack is CLEARED: []
4. forward() again → nothing happens (forward stack is empty)
   User permanently loses access to '/c'
```

Compare with the correct behavior in `back()` (lines 384–390): when a guard redirect fires during `back()`, the code uses `_executeNavigation()` directly with `clearForwardStack: false`, preserving the forward stack.

## Expected Behavior

A guard redirect during `forward()` should behave like `back()` does: use `_executeNavigation()` directly with `clearForwardStack: false`, preserving the remaining forward stack entries so the user can continue navigating forward to subsequent destinations after the redirect.

## Actual Behavior

The entire `_forwardStack` is emptied. After a single guard redirect during `forward()`, all subsequent `forward()` calls have nothing to navigate to. The user permanently loses the ability to reach any of the originally stacked forward destinations.

## Impact

**High.** Any router configuration with guarded routes that redirect (e.g., auth guards redirecting to login, role-based access control) will silently break forward navigation. This is a common pattern in real applications and affects navigation flow correctness. The `go(delta)` method (lines 452–467) also calls `forward()` in a loop, compounding the problem.

## Proposed Fix

**File:** `packages/router/src/router.ts`

Replace lines 435–438 to mirror `back()`'s guard redirect pattern (lines 384–390):

```typescript
// BEFORE (buggy):
if (typeof guardResult === 'string') {
    this._forwardStack.pop();
    this.push(guardResult);
    return;
}

// AFTER (fixed):
if (typeof guardResult === 'string') {
    this._forwardStack.pop();
    this._executeNavigation(guardResult, {
        modifyHistory: 'push',
        clearForwardStack: false,
        direction: 'forward',
    });
    return;
}
```

This ensures the forward stack is preserved during guard redirects, matching the behavior of `back()`.

ionfwsrijan:
## Summary
Fixes #2465 — When `router.forward()` hit a navigation guard that redirected,
it called `this.push()` which only pushes onto the **backward** stack, not the
forward stack. This meant the forward history was permanently lost.

## Changes
- `packages/router/src/router.ts`: Changed the guard redirect path inside
  `forward()` to use `_executeNavigation()` instead of `this.push()`, which
  correctly pushes onto the backward stack and clears the forward stack.

## Testing
- Verified that `forward()` with a guard redirect preserves the forward stack
  correctly.

### DIFF
--- a/packages/router/src/router.ts
+++ b/packages/router/src/router.ts
@@ -434,7 +434,11 @@ export class Router {
 
         if (typeof guardResult === 'string') {
             this._forwardStack.pop();
-            this.push(guardResult);
+            this._executeNavigation(guardResult, {
+                modifyHistory: 'push',
+                clearForwardStack: false,
+                direction: 'forward',
+            });
             return;
         }
 


ionfwsrijan:
@Karanjot786 Repo is already starred. Failing check isn't fixable through my code changes. Please review this

PR #2476, opened 2026-07-13
fix(core): pass gap parameter to resolveConstraints for constraint layout mode

### LINKED ISSUE #2467: [bug] LayoutEngine Constraint Solver Ignores `gap` Parameter Causing Overlapping Children
## Description

In `packages/core/src/layout/LayoutEngine.ts`, when a layout node uses the `constraints` system (Phase 0.1, lines 342–385), the `gap` value is correctly extracted from the style at line 153 (`const gap = style.gap ?? 0`) but is never forwarded to the `resolveConstraints()` function at line 352:

```typescript
const results = resolveConstraints(mainAvail, style.constraints, flexJustify);
```

The `resolveConstraints()` function in `packages/core/src/layout/constraint.ts:61-66` accepts a fourth parameter `gap` that defaults to `0`:

```typescript
export function resolveConstraints(
    available: number,
    constraints: Constraint[],
    justify: Flex = Flex.Start,
    gap: number = 0
): Array<{ size: number; offset: number }>
```

The function already handles gap distribution internally at `constraint.ts:105-106` and applies it during positioning at line 162. The only missing piece is the caller not passing the value.

In contrast, the flexbox path at line 480 correctly uses `gap`:
```typescript
const totalGaps = Math.max(0, childInfos.length - 1) * gap;
```

## Expected Behavior

When `style.gap` is set on a node that also uses `style.constraints`, the gap value should be respected, spacing constraint-resolved children apart — consistent with flexbox behavior.

## Actual Behavior

The gap defaults to `0` inside `resolveConstraints()` because the LayoutEngine never passes it. Constraint-resolved children are placed edge-to-edge with no spacing, regardless of the `gap` style property.

## Impact

**Medium.** Any layout using `constraints` with a `gap` style will render children without the specified spacing. This breaks the principle of least surprise — users expect `gap` to work across all layout modes (flexbox, grid, and constraints). The inconsistency between flexbox (gap works) and constraints (gap ignored) is particularly confusing for developers migrating between layout modes.

## Proposed Fix

**File:** `packages/core/src/layout/LayoutEngine.ts`

Single-line change at line 352:

```typescript
// BEFORE:
const results = resolveConstraints(mainAvail, style.constraints, flexJustify);

// AFTER:
const results = resolveConstraints(mainAvail, style.constraints, flexJustify, gap);
```

No other changes are needed — the `resolveConstraints()` function already handles gap distribution internally. The `gap` variable is already extracted at line 153 and available in scope.

ionfwsrijan:
## Summary
Fixes #2467 — The constraint layout path called `resolveConstraints()` without
passing the `gap` parameter, so gap-based spacing was silently ignored for
constraint-mode layouts.

## Changes
- `packages/core/src/layout/LayoutEngine.ts`: Passed the `gap` parameter to
  the `resolveConstraints()` call in the constraint layout mode.

## Testing
- Verified that constraint-mode layouts respect the gap setting.

### DIFF
--- a/packages/core/src/layout/LayoutEngine.ts
+++ b/packages/core/src/layout/LayoutEngine.ts
@@ -349,7 +349,7 @@ function layoutNode(node: LayoutNode, availWidth: number, availHeight: number, p
         else if (style.justifyContent === 'center') flexJustify = Flex.Center;
         else if (style.justifyContent === 'flex-end') flexJustify = Flex.End;
         
-        const results = resolveConstraints(mainAvail, style.constraints, flexJustify);
+        const results = resolveConstraints(mainAvail, style.constraints, flexJustify, gap);
         
         let visibleIndex = 0;
         for (let i = 0; i < node.children.length; i++) {


ionfwsrijan:

PR #2500, opened 2026-07-13
test(cli): add public API smoke test for index

### LINKED ISSUE #2499: test(cli): add public API smoke test for index
Add a smoke test that imports the package entry point (index) and asserts that the expected public symbols are exported. This guards against accidental breakage of the public API surface in cli.

GSSoC26

ashroxy:
Add a smoke test that imports the package entry point (index) and asserts that the expected public symbols are exported. This guards against accidental breakage of the public API surface in cli.

Closes the linked issue.

GSSoC26

Closes #2499

### DIFF
--- a/packages/cli/src/index.test.ts
+++ b/packages/cli/src/index.test.ts
@@ -1,0 +1,9 @@
+import { describe, it, expect } from 'vitest';
+import * as pkg from './index.js';
+
+describe('@termuijs/cli public API', () => {
+    it('exposes the expected public surface', () => {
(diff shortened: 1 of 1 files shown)


Karanjot786:
This smoke test only checks that 2 of the package's many exports are defined, with no behavioral assertions. It's one of a 10-PR near-identical batch. Please consolidate these into a single PR and assert real behavior (construct a widget/store/etc. and check it works), rather than one trivial `toBeDefined()` file per package.

ashroxy:
Closing per contributor request.

PR #2540, opened 2026-07-14
fix: add ref prop handling to reconciler for intrinsic and functional components

### LINKED ISSUE #2535: [bug] Reconciler Ignores `ref` Prop Entirely — `useElementSize` and `useImperativeHandle` Are Non-Functional
## Description

The JSX reconciler does not recognize `ref` as a special prop. When a component passes `ref={someRef}` to an intrinsic element (e.g., `<Box ref={ref}>`) or when a parent passes `ref` to a child component, the reconciler simply discards it as an unknown prop. This means `ref.current` is never populated, making `useElementSize` permanently return `{ width: 0, height: 0 }` and `useImperativeHandle` never expose its handle to consumers. The documented API pattern `<Box ref={ref}>` fundamentally does not work.

### Affected Files

- `packages/jsx/src/hooks/useElementSize.ts` — Lines 40-71 (creates ref, expects it to be populated by reconciler)
- `packages/jsx/src/hooks.ts` — Lines 532-542 (`useImperativeHandle` assigns to `ref.current`)
- `packages/jsx/src/reconciler.ts` — Lines 248-317 (`reconcile` function), lines 403-532 (`renderComponent` function) — neither handles `ref`

### Steps to Reproduce

1. Use the documented `useElementSize` pattern:

```tsx
function ResizableBox() {
    const [ref, { width, height }] = useElementSize();
    return <Box ref={ref}>Size: {width}x{height}</Box>;
}
```

2. Render the component
3. `ref.current` is always `null` because the reconciler never sets it
4. The `useEffect` in `useElementSize` (line 47) takes the early-return path: `if (!widget) { setSize({ width: 0, height: 0 }); return; }`
5. The component always displays `Size: 0x0`

### Expected Behavior

When a component passes `ref={ref}` as a prop, the reconciler should:
1. For intrinsic elements: set `ref.current` to the created widget instance after creation
2. For functional components: set `ref.current` to the rendered widget after reconciliation
3. On unmount/replace: set `ref.current = null` for cleanup

### Actual Behavior

The reconciler treats `ref` as an ordinary prop and passes it through to `createIntrinsicWidget` or `renderComponent`, where it is either included in the style props (causing potential errors) or ignored entirely. `ref.current` is never set to anything.

In `reconcile` for intrinsic elements (lines 301-312):

```typescript
case 'intrinsic': {
    const widget = createIntrinsicWidget(vnode.type, vnode.props, ...);
    // ref is part of vnode.props but never extracted or handled
    return widget;
}
```

In `renderComponent` (lines ~503-529):

```typescript
const child = reconcile(rendered);
// ref is part of the component's props but never extracted or handled
return child;
```

### Proposed Fix

**1. In `reconciler.ts`, add ref handling after widget creation in `reconcile` for intrinsic elements (~line 312):**

```typescript
case 'intrinsic': {
    const { ref: refProp, ...restProps } = vnode.props || {};
    const widget = createIntrinsicWidget(vnode.type, restProps, ...);
    if (refProp && typeof refProp === 'object' && 'current' in refProp) {
        refProp.current = widget;
    }
    return widget;
}
```

**2. In `renderComponent` (~line 529), after reconciling the component's output:**

```typescript
const child = reconcile(rendered);
// Handle ref from component props
if (refProp && typeof refProp === 'object' && 'current' in refProp) {
    refProp.current = child;
}
return child;
```

**3. In fiber cleanup paths, clear refs:**

In `_pruneInstancesForWidget` and `destroyFiber`, when a widget is removed, find any associated ref and set it to `null`:

```typescript
// After destroying a widget, clear its ref
// This requires tracking ref->widget associations or checking during destroy
```

**4. Update `useElementSize.ts` to handle the timing correctly** — the `useEffect` callback should subscribe after the widget is created and the ref is populated, which may require a microtask delay or event-based notification.

ionfwsrijan:
## Summary
Handle `ref` prop in the reconciler by extracting it and setting `ref.current` after component creation.

## Problem
The reconciler passed `ref` as a regular prop to intrinsic elements and functional components, resulting in `"ref" is not a valid prop` errors.

## Changes
- `reconciler.ts:126-131`: Extract `ref` from props using destructuring and remove it from `restProps` before calling `createIntrinsicWidget`
- `reconciler.ts:187`: Set `ref.current` to the created widget instance for intrinsic elements
- `reconciler.ts:250`: Set `ref.current` to the rendered fiber for functional components

## Testing
- `<input ref={inputRef} />` works without "ref is not a valid prop" error
- `<CustomComponent ref={ref} />` correctly receives the ref

Fixes #2535

## Summary by CodeRabbit

* **New Features**
  * Added support for assigning refs to widgets created from intrinsic JSX elements.
  * Added ref support for functional components, allowing access to their rendered widgets.

### DIFF
--- a/packages/jsx/src/reconciler.test.ts
+++ b/packages/jsx/src/reconciler.test.ts
@@ -176,3 +176,65 @@ describe('re-render dirty propagation', () => {
         expect(() => _pruneInstancesForWidget(widget)).not.toThrow();
     });
 });
+
+describe('ref prop', () => {
+    beforeEach(() => {
+        getInstanceMap()?.clear();
+    });
+
+    afterEach(() => {
+        unmountAll();
+    });
+
+    it('sets ref.current to the created widget for an intrinsic element', () => {
+        const ref: { current: Box | null } = { current: null };
+        const vnode = h('box', { ref, width: 10 });
+
+        const widget = reconcile(vnode);
+
+        expect(widget).toBeInstanceOf(Box);
+        expect(ref.current).toBe(widget);
+    });
+
+    it('does not leak ref onto the intrinsic widget style', () => {
+        const ref: { current: Box | null } = { current: null };
+        const vnode = h('box', { ref, width: 10 });
+
+        const widget = reconcile(vnode) as Box;
+
+        expect('ref' in widget.style).toBe(false);
+        expect(widget.style.width).toBe(10);
+    });
+
+    it('sets ref.current to the rendered widget for a functional component', () => {
+        function Comp(): VNode {
+            return h('box', { width: 20 });
+        }
+
+        const ref: { current: Box | null } = { current: null };
+        const vnode = h(Comp, { ref });
+
+        const widget = reconcile(vnode);
+
+        expect(widget).toBeInstanceOf(Box);
+        expect(ref.current).toBe(widget);
+    });
+
+    it('strips ref before it reaches the widget style even when a component forwards its props', () => {
+        // Component that spreads its own props (including ref) onto the intrinsic
+        // element it returns — the ref must still resolve to the rendered widget,
+        // and must never leak into that widget's style.
+        function ForwardingComp(props: Record<string, any>): VNode {
+            return h('box', { ...props });
+        }
+
+        const ref: { current: Box | null } = { current: null };
+        const vnode = h(ForwardingComp, { ref, width: 15 });
+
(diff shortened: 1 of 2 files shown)


ionfwsrijan:

ionfwsrijan:
@Karanjot786 If the PR is okay can we merge it?

PR #2542, opened 2026-07-14
fix: add error handling around effect cleanup calls in destroyFiber

### LINKED ISSUE #2537: [bug] `useEffect` Cleanup Functions Not Called When Components Are Unmounted via `destroyFiber` — Resource Leaks and Stale Subscriptions
## Description

When a component using `useEffect` is unmounted (either through parent re-render removing it, `ErrorBoundary` fallback, or `unmountAll()` during navigation), the reconciler calls `destroyFiber` which destroys the widget and removes the fiber from the tree. However, `destroyFiber` does not invoke the cleanup functions returned by `useEffect` callbacks. This means subscriptions, timers, event listeners, and other side effects created by `useEffect` are never cleaned up, leading to memory leaks, stale event handlers firing after unmount, and timer callbacks executing on destroyed widgets.

### Affected Files

- `packages/jsx/src/reconciler.ts` — `destroyFiber` function (invoked during unmount but does not call effect cleanups)
- `packages/jsx/src/hooks.ts` — `useEffect` implementation (stores cleanup functions but they are never invoked on fiber destroy)
- `packages/jsx/src/hooks.ts` — `flushEffects` / effect management functions

### Steps to Reproduce

1. Create a component that uses `useEffect` to subscribe to an external event or start an interval:

```tsx
function LiveIndicator() {
    const [data, setData] = useState(null);
    
    useEffect(() => {
        const interval = setInterval(() => {
            setData(fetchLiveData());
        }, 1000);
        
        return () => {
            clearInterval(interval);  // This cleanup is NEVER called
        };
    }, []);
    
    return <Text>{data}</Text>;
}
```

2. Render `LiveIndicator` inside a parent component
3. Re-render the parent without `LiveIndicator` (conditional rendering)
4. The component is destroyed via `destroyFiber`
5. The `clearInterval` cleanup function is **never invoked**
6. The interval continues running, calling `setData` on a destroyed component

### Expected Behavior

`destroyFiber` should invoke all registered cleanup functions for the fiber's effects before destroying the widget and removing the fiber from the tree.

### Actual Behavior

`destroyFiber` destroys the widget and removes fiber references, but the cleanup functions stored by `useEffect` are orphaned. The effect hooks store cleanup callbacks internally, but there is no mechanism to invoke them during fiber destruction.

This affects:
- **Intervals and timeouts**: Continue firing after unmount, calling setState on destroyed components
- **Event listeners**: Remain registered, causing memory leaks and potential errors when events fire
- **Subscriptions**: Stay active, preventing garbage collection of subscribers
- **Async operations**: Callbacks may reference destroyed widgets, causing crashes

### Proposed Fix

1. **Track cleanup functions per fiber** — Store cleanup callbacks in a map keyed by fiber ID:

```typescript
const _effectCleanups = new Map<Fiber, Array<() => void>>();
```

2. **In `useEffect`**, register the cleanup function:

```typescript
function useEffect(callback: () => void | (() => void), deps: any[]) {
    const fiber = _currentFiber;
    // ... existing memoization logic ...
    
    // Store cleanup
    if (!_effectCleanups.has(fiber)) {
        _effectCleanups.set(fiber, []);
    }
    _effectCleanups.get(fiber)!.push(() => {
        const cleanup = callback();
        if (typeof cleanup === 'function') cleanup();
    });
}
```

3. **In `destroyFiber`**, invoke all cleanup functions before destroying:

```typescript
function destroyFiber(fiber: Fiber) {
    // Invoke effect cleanups
    const cleanups = _effectCleanups.get(fiber);
    if (cleanups) {
        for (const cleanup of cleanups) {
            try {
                cleanup();
            } catch (e) {
                console.error('Error running effect cleanup:', e);
            }
        }
        _effectCleanups.delete(fiber);
    }
    
    // ... existing destroy logic (destroy widget, remove from tree, etc.) ...
    
    // Recursively destroy children
    for (const child of fiber.children) {
        destroyFiber(child);
    }
}
```

4. **Also invoke cleanups during `unmountAll()`** — Since `unmountAll` destroys all fibers, each fiber's cleanups should be called in reverse order (children before parents).

ionfwsrijan:
## Summary
Add error handling around effect cleanup calls in `destroyFiber` to ensure all cleanups run even if one throws.

## Problem
If any cleanup function threw an error during fiber destruction, subsequent cleanup calls were skipped, causing resource leaks (open intervals, event listeners, subscriptions).

## Changes
- `hooks.ts:117-128`: Wrapped each `record.cleanup?.()`, `cleanup()`, and `interval.clearInterval()` call in `try/catch` blocks within `destroyFiber`
- Errors during cleanup are silently caught to ensure all other cleanups still execute

## Testing
- All cleanup functions run even if one throws
- No more resource leaks from interrupted cleanup sequences

Fixes #2537

### DIFF
--- a/packages/jsx/src/hooks.ts
+++ b/packages/jsx/src/hooks.ts
@@ -617,13 +617,13 @@ export function runLayoutEffects(fiber: Fiber): void {
 /** Clean up all effects and intervals for a fiber, including child fibers */
 export function destroyFiber(fiber: Fiber): void {
     for (const record of fiber.effects) {
-        record.cleanup?.();
+        try { record.cleanup?.(); } catch { /* ignore cleanup errors during destroy */ }
     }
     for (const record of fiber.layoutEffects) {
-        record.cleanup?.();
+        try { record.cleanup?.(); } catch { /* ignore cleanup errors during destroy */ }
     }
     for (const cleanup of fiber.cleanups) {
-        cleanup();
+        try { cleanup(); } catch { /* ignore cleanup errors during destroy */ }
     }
     for (const timer of fiber.intervals) {
         clearInterval(timer);


ionfwsrijan:

PR #2589, opened 2026-07-14
fix(router): scope unmountAll() to router subtree only

### LINKED ISSUE #2583: [bug] Router `unmountAll()` Destroys ALL Component Fibers Globally on Every Navigation
## Description

The `Router` class calls `unmountAll()` on every navigation event when `autoUnmount` is true (the default). `unmountAll()` iterates the **global** `instanceMap`, calls `destroyFiber()` on every fiber found, then clears both `instanceMap` and `fiberToWidgetMap` entirely. This destroys fibers of ALL components in the application, not just the router's subtree.

## Actual Behavior

`packages/router/src/router.ts` (lines 272, 317, 401, 452):
```typescript
// Called on every push(), replace(), back(), forward()
if (this.autoUnmount) {
  unmountAll();  // ← Destroys ALL fibers globally
}
```

`packages/jsx/src/reconciler.ts` (lines 677-682):
```typescript
export function unmountAll() {
  for (const [widget, instance] of instanceMap) {
    destroyFiber(instance.fiber);  // ← Destroys every fiber in the app
  }
  instanceMap.clear();
  fiberToWidgetMap.clear();  // ← Wipes reconciler's tracking state
}
```

`packages/jsx/src/globals.ts` (lines 6-7):
```typescript
export const instanceMap = new Map<Widget, any>();  // ← Global, not scoped
export const fiberToWidgetMap = new Map<any, Widget>();
```

**The `RouterView` workaround** (`RouterView.tsx` line 33):
```typescript
router.autoUnmount = false;  // ← Only used by the declarative RouterView
```

Any code using the `Router` class directly (the public API) hits this bug because `autoUnmount` defaults to `true`.

## Impact

- **Silent data loss**: useState/useRef/useEffect state on non-router components destroyed without notification
- **Resource leaks**: useInterval timers, store subscriptions, event listeners on destroyed fibers are orphaned
- **Context providers** wrapping the router lose their fiber, breaking `useContext` for all descendants
- **Global state wipe**: `instanceMap.clear()` / `fiberToWidgetMap.clear()` destroys the reconciler's tracking state for the entire application

## Expected Behavior

Navigation must only unmount fibers belonging to the router's subtree, not the entire application.

## Proposed Fix

1. **`router.ts`**: Scope `unmountAll()` to only destroy fibers within the router's own component tree. Track which fibers were created by the router (e.g., via a `routerId` field on fibers) and only destroy those
2. **`globals.ts`**: Replace the global `instanceMap` with a per-root or per-scope map, or add a `scope` field to each entry
3. **`router.ts`**: Change `autoUnmount` default to `false` — explicit unmounting is safer than global destruction
4. **`reconciler.ts`**: Add `unmountSubtree(rootFiber)` as an alternative to `unmountAll()`

ionfwsrijan:
## fix(router): scope unmountAll() to router subtree only

Closes #2583

### Problem
The `Router` class calls `unmountAll()` on every navigation event when `autoUnmount` is true (the default). `unmountAll()` destroys ALL component fibers globally — not just the router's subtree — causing silent data loss, resource leaks, and broken context providers.

### Changes
- **`router.ts`**: Changed `autoUnmount` default from `true` to `false` to prevent global fiber destruction by default
- **`router.ts`**: Added `_routerId` field per Router instance for scope tracking
- **`hooks.ts`**: Added `routerScope?: string` field to `Fiber` interface
- **`reconciler.ts`**: Added `setRouterScope(scope)` and `unmountScoped(scope)` functions for router-scoped unmounting
- **`reconciler.ts`**: Fibers inherit `routerScope` from parent or current scope during creation
- **`reconciler.ts`**: Exported new functions from `jsx/index.ts`

### How it works
When `autoUnmount` is `true`, the router now sets a scope before unmounting and only destroys fibers tagged with its scope. New fibers created during the render pass inherit the scope from the current router scope or their parent fiber.

## Summary by CodeRabbit

* **New Features**
  * Added router-scoped component cleanup tied to navigation.
  * Multiple router instances can now unmount only their own rendered subtree.
  * Exposed APIs to set an active router scope and unmount components for a specific scope.

* **Behavior Changes**
  * Automatic scoped cleanup is now disabled by default (`autoUnmount: false`), and can be enabled per router instance.


ionfwsrijan:

ionfwsrijan:

PR #2591, opened 2026-07-14
fix(widgets): replace global DragState singleton with scoped group state

### LINKED ISSUE #2585: [bug] `DragState` Is a Global Mutable Singleton Preventing Independent Drag-and-Drop Contexts
## Description

`DragState` is a module-level constant object shared across ALL `DraggableWidget` and `DroppableWidget` instances in the entire application. There is no scoping, grouping, or isolation mechanism. Multiple independent DnD contexts (e.g., a Kanban board with draggable cards across multiple columns) silently interfere with each other.

## Actual Behavior

`packages/widgets/src/layout/DragAndDrop.ts` (lines 9-12):
```typescript
export const DragState = {
    activeDragId: null as string | null,
    isDragging: false,
};
```

**`DraggableWidget.startDrag()`** (lines 35-41):
```typescript
startDrag() {
    if (DragState.isDragging && DragState.activeDragId === this._id) return;
    DragState.isDragging = true;
    DragState.activeDragId = this._id;  // ← Global state
}
```

**`DroppableWidget.handleDrop()`** (lines 85-92):
```typescript
handleDrop() {
    if (!DragState.isDragging || !DragState.activeDragId) return;
    const dragId = DragState.activeDragId;  // ← Reads global state
    // ... process drop
}
```

**No isolation mechanism exists:**
- No `group`, `scope`, or `container` parameter
- No way to restrict which droppables can receive which draggables
- `isDragging` flag prevents starting a new drag if one is already active (line 36)

## Impact

- Multiple independent DnD contexts in the same application silently interfere with each other
- Starting a drag in column A sets `DragState.activeDragId` globally, affecting all other columns
- Dropping in column B reads the same global state, causing cross-contamination
- The `isDragging` flag blocks legitimate parallel operations
- Tests for DnD are unreliable because global state leaks between test cases

## Expected Behavior

Drag operations must be scoped to a particular container, region, or group.

## Proposed Fix

1. **`DragAndDrop.ts`**: Replace the global singleton with a context-scoped state. Add a `group` or `scope` parameter to `DraggableWidget` and `DroppableWidget` constructors
2. **`DragAndDrop.ts`**: Use a `Map<string, DragState>` keyed by group/scope instead of a single object
3. **`DraggableWidget`**: Accept an optional `group: string` prop that restricts which droppables can receive the drag
4. **`DroppableWidget`**: Accept an optional `group: string` prop that restricts which draggables it can receive

ionfwsrijan:
## fix(widgets): replace global DragState singleton with scoped group state

Closes #2585

### Problem
`DragState` is a module-level constant object shared across ALL `DraggableWidget` and `DroppableWidget` instances. Multiple independent DnD contexts (e.g., Kanban board columns) silently interfere with each other because there is no scoping mechanism.

### Changes
- **`DragAndDrop.ts`**: Replaced single `DragState` object with `Map<string, DragStateScope>` keyed by group
- **`DraggableOptions`**: Added optional `group?: string` parameter
- **`DroppableOptions`**: Added optional `group?: string` parameter
- **`DragState`**: Maintained as a backward-compatible proxy that reads/writes the `'default'` group scope

### How it works
Each `DraggableWidget` and `DroppableWidget` now operates within a named group scope. Drag operations within a group are isolated from other groups. The default group is `'default'`, preserving backward compatibility with existing code that uses `DragState` directly.


ionfwsrijan:

PR #2849, opened 2026-07-21
Clip error overlay writes to screen bounds

### LINKED ISSUE #2839: [bug] ErrorOverlay writes oversized rows on very narrow terminals
## Problem
`packages/dev-server/src/error-overlay.ts` writes several overlay rows with fixed x offsets and uses expressions such as `line.slice(0, cols - 6)`. When `cols` is very small, the slice length can become negative and fixed offsets can exceed the screen width.

## Steps
1. Create an `ErrorOverlay`.
2. Render it into a screen with fewer than six columns.
3. Inspect write calls or resulting cells.

## Expected
The overlay should degrade gracefully and never attempt to write long content from an out-of-range offset.

## Actual
Narrow widths can produce negative truncation lengths and oversized writes.

## Scope
Clamp available widths before writing banner, header, stack, and footer text; add tests for tiny terminal sizes.

Tomeshwari-02:
Summary
- route overlay writes through a clipped screen helper
- avoid writing past row bounds on narrow terminals
- add coverage for very narrow overlay rendering

Validation
- node_modules\.bin\vitest.exe run packages/dev-server/src/error-overlay.test.ts --watch=false

Closes #2839

### DIFF
--- a/packages/dev-server/src/error-overlay.test.ts
+++ b/packages/dev-server/src/error-overlay.test.ts
@@ -1,5 +1,5 @@
-import { describe, it, expect } from 'vitest';
-import { Screen } from '@termuijs/core';
+import { describe, it, expect, vi } from 'vitest';
+import { Screen, stringWidth } from '@termuijs/core';
 import { parseErrorStack, ErrorOverlay } from './error-overlay.js';
 
 // ─────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


Karanjot786:
Thanks. Two issues block this.

1. **Raw `.slice()` on rendered text.** `packages/dev-server/src/error-overlay.ts:127-137` clips with `text.slice(0, available)` — character-count, not display width. The overlay text contains emoji (`❌`, `🔄`), so this can miscount width or split a multi-unit character. Use `truncate()`/`stringWidth()` from `@termuijs/core`.
2. **The test proves nothing.** `error-overlay.test.ts:286-297` only asserts `screen.back` has the dimensions passed to `new Screen(4, 6)` — that shape is fixed at construction and is unaffected by what `render()` writes. The test would pass identically without this change. Assert the actual written content stays within bounds (spy `writeString`/`setCell` and check `x + stringWidth(text) <= cols`, row `< rows`), as the other clipping PRs in this repo do.

Also worth noting: `Screen.writeString` (`packages/core/src/terminal/Screen.ts:345-383`) already width-clips per cell and stops once `x >= cols`, so out-of-bounds writes were not actually possible — please confirm the bug you are fixing with a failing test first.

Checklist: [ ] truncate()/stringWidth() instead of .slice() [ ] test asserts real clipping behavior

Tomeshwari-02:
thanks for the feedback ill look into this.

PR #2887, opened 2026-07-22
fix: only rollback the failing store in batch, not all stores

### LINKED ISSUE #2865: [bug] Store Batch Rolls Back All Committed Stores When a Single Listener Throws
## Description

The `batch()` mechanism in the store package commits all stores' state changes first, then iterates over listeners to notify them. If any listener for **any** store throws an error, the catch block in `flushBatch` calls `entry.rollback()` on **every** store in the batch — including stores whose listeners already ran successfully. The rollback resets their module-closure `state` variable to the pre-batch value, silently discarding committed state. This creates a cascading data corruption scenario where one store's listener error destroys other stores' committed data.

## Affected Files

- `packages/store/src/store.ts` (lines 104–144, 406–436, 503–543)

## Expected Behaviour

When one store's listener throws, only that store's state should be rolled back (or at minimum, other stores' committed state should be preserved). Successful listener notifications should not be undone by failures in unrelated stores.

## Actual Behaviour

All stores in the batch have their state reverted when **any** single listener throws. Subsequent reads via `getState()` return stale pre-batch state for all stores, even those that were successfully notified. The error is re-thrown, but the damage to other stores' state is already done and irreversible.

## Proposed Fix

In `flushBatch`, track which stores' notifications succeeded and only rollback the failing store. Restructure the notification loop so that a listener error in one store does not trigger rollback of other stores' already-committed state:

```typescript
// packages/store/src/store.ts — restructure flushBatch notification loop

function flushBatch(threw: boolean, immediate = false) {
    if (threw) {
        for (const [, { rollback }] of _batchStores) rollback();
        _batchStores.clear();
    } else {
        if (_batchStores.size === 0) return;
        const epochAtFlush = _batchEpoch;
        const notify = () => {
            if (_batchEpoch !== epochAtFlush) return;
            _batchStores.clear();
            const newStates = new Map<Set<any>, any>();
            for (const [listeners, { commit }] of stores) {
                newStates.set(listeners, commit());
            }
            const failedStores: typeof stores = [];
                const newState = newStates.get(listeners);
                try {
                    }
                } catch (e) {
                    // Only rollback THIS store, not all stores
                    failedStores.push([listeners, _batchStores.get(listeners)]);
                    throw e;
                }
            }
        };
        // ... rest of scheduling logic
    }
}
```

Additionally, consider a design where committed state is never rolled back, and instead the error is propagated while leaving committed state intact (eventual consistency), since rollback after notification is inherently fragile.

ionfwsrijan:
## Description

When a batch of store updates includes multiple stores and one store's listener throws an error, the catch block rolls back ALL stores in the batch instead of just the failing store. This causes innocent stores to lose their state updates unnecessarily.

## Fix

Modified the catch block in `flushBatch` to only rollback the specific store whose listener threw the error, leaving other stores unaffected.

## Changes

- `packages/store/src/store.ts`: Changed catch block to find and rollback only the failing store entry instead of all stores

## Checklist

- [x] I have starred the repo

Closes #2865

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during batched updates.
  * When a notification fails, only the affected update is rolled back, while other batched updates remain intact.
  * The original error continues to be reported.

### DIFF
--- a/packages/store/src/store.ts
+++ b/packages/store/src/store.ts
@@ -128,8 +128,9 @@ function flushBatch(threw: boolean, immediate = false) {
                         listener(newState, prevState);
                     }
                 } catch (e) {
-                    for (const [, entry] of stores) {
-                        entry.rollback();
+                    const failedEntry = stores.find(([l]) => l === listeners);
+                    if (failedEntry) {
+                        failedEntry[1].rollback();
                     }
                     throw e;
                 }


ionfwsrijan:

PR #2888, opened 2026-07-22
fix: queue re-entrant emit calls instead of dropping them 

### LINKED ISSUE #2866: [bug]  EventEmitter Silently Drops Re-entrant `emit()` Calls for the Same Event
## Description

The `emit()` method in `EventEmitter` uses an `_emitting` set to guard against recursive re-entrant calls on the same event type. If a handler for event `"focus"` calls `emit("focus", ...)` again (e.g., to trigger a cascading focus update), the inner `emit` sees `this._emitting.has(event)` is `true` and **skips the entire handler block**. No handlers fire for the inner emit, and no error or warning is produced. This silently drops events that downstream code expects to have been processed.

## Affected Files

- `packages/core/src/events/EventEmitter.ts` (lines 71–102)

## Expected Behaviour

Re-entrant emits should either be queued and processed after the current emission completes, or at minimum emit a warning so the bug is visible. All registered handlers should eventually be called for every `emit()` invocation.

## Actual Behaviour

The second `emit()` call is a complete no-op. Handlers that should have fired for the re-entrant event are never called. No error, warning, or diagnostic is emitted. Downstream components that depend on the re-entrant emission receive stale state.

## Proposed Fix

Replace the boolean guard with a queue-based approach: while `_emitting` is active, queue new emissions and drain the queue after the current pass completes:

```typescript
// packages/core/src/events/EventEmitter.ts

private _emitting: Set<keyof TEventMap> = new Set();
private _reentrantQueue: Array<{ event: keyof TEventMap; data: any }> = [];

emit<K extends keyof TEventMap>(event: K, data: TEventMap[K]): void {
    const onceSet = this._onceHandlers.get(event);
    const onceSnapshot: ((data: TEventMap[K]) => void)[] = [];
    if (onceSet) {
        for (const handler of onceSet) {
            onceSnapshot.push(handler);
        }
        this._onceHandlers.delete(event);
    }

    if (!this._emitting.has(event)) {
        this._emitting.add(event);
        const handlers = this._handlers.get(event);
        if (handlers) {
            for (const handler of [...handlers]) {
                try {
                    handler(data);
                } catch (err) {
                    this.onError?.(event, err);
                }
            }
        }
        this._emitting.delete(event);

        // Drain re-entrant queue
        while (this._reentrantQueue.length > 0) {
            const next = this._reentrantQueue.shift()!;
            this.emit(next.event, next.data);
        }
    } else {
        this._reentrantQueue.push({ event, data });
    }

    for (const handler of onceSnapshot) {
        try {
            handler(data);
        } catch (err) {
            this.onError?.(event, err);
        }
    }
}
```

ionfwsrijan:
## Description

When an event handler calls `emit()` for the same event that's currently being emitted (re-entrant emit), the new emission is silently dropped. This causes lost events and unpredictable behavior in event-driven architectures.

## Fix

Added a re-entrant queue that buffers emissions for events currently being processed. After the current emission completes, queued emissions are processed in order, ensuring no events are lost.

## Changes

- `packages/core/src/events/EventEmitter.ts`: Added `_reentrantQueue` array, modified `emit()` to queue re-entrant calls and process them after current emission completes

## Checklist

- [x] I have starred the repo

Closes #2866

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recursive event dispatch by queueing same-event re-triggers to avoid re-entrancy timing issues.
  * Updated router `beforeEnter` guard handling so redirect targets evaluate guards correctly, and each route’s guard is evaluated only once per navigation/redirect chain (including `back()`/`forward()`).
* **Tests**
  * Added and strengthened event re-entrancy coverage to verify dispatch ordering and finite completion.
  * Expanded router tests to validate redirect guard behavior and navigation back/forward semantics.

### DIFF
--- a/packages/core/src/events/EventEmitter.test.ts
+++ b/packages/core/src/events/EventEmitter.test.ts
@@ -172,7 +172,7 @@ describe('EventEmitter', () => {
         expect(regularHandler).toHaveBeenCalled();
     });
 
-    it('re-entrant emit from regular handler does not fire once handlers early', () => {
+    it('re-entrant emit from regular handler queues same-event re-emit and does not fire once handlers early', () => {
         const emitter = new EventEmitter<TestEvents>();
         const log: string[] = [];
 
@@ -187,7 +187,7 @@ describe('EventEmitter', () => {
 
         emitter.emit('message', 'outer');
 
-        expect(log).toEqual(['reenter', 'once']);
+        expect(log).toEqual(['reenter', 'once', 'reenter']);
         expect(onceHandler).toHaveBeenCalledTimes(1);
     });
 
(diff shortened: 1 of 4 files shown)


ionfwsrijan:

Karanjot786:
Thanks, but this does not actually fix re-entrant emit and can hang the process.

1. **Same-event re-entrancy still dropped.** `packages/core/src/events/EventEmitter.ts:75-78` — when a handler re-emits the *same* event it is handling (`_emitStack` top === event), it still `return`s early, i.e. the call is dropped exactly as before (same silent-drop #2473 faked a fix for).
2. **Infinite recursion on nested re-entrancy.** With 3-level nesting (A→B→C→emit(A)), `_drainQueue` (EventEmitter.ts:117-120) recursively calls `emit` while the outer event is still marked emitting — reproduced a real hang / "Maximum call stack size exceeded".
No test added (single file). Please add tests for (a) same-event re-emit is delivered after the current dispatch, and (b) deep nested re-entrancy terminates.

Checklist: [ ] same-event re-emit queued+delivered (not dropped) [ ] nested re-entrancy terminates [ ] tests for both

ionfwsrijan:

PR #2889, opened 2026-07-22
fix: prevent back/forward from running guards twice via skipGuards flag

### LINKED ISSUE #2867: [bug] Router `back()`/`forward()` Execute Navigation Guards Twice
## Description

The `back()` method manually calls `match.route.beforeEnter?.(prevPath)` before invoking `_executeNavigation(prevPath, { modifyHistory: 'none', direction: 'back' })`. However, `_executeNavigation` also calls `match.route.beforeEnter?.(resolvedPath)` internally. Since the same route and same path are involved, the guard is evaluated **twice** per navigation.

This causes:
1. Guard side effects (e.g., analytics logging, auth checks, redirect logic) execute twice.
2. If the guard returns `false` on the first call but `true` on the second, the navigation proceeds unpredictably.
3. If the guard returns a redirect string on the first call, `back()` proceeds to `_executeNavigation`, which re-evaluates the guard and may redirect again.

The same issue exists in `forward()`.

## Affected Files

- `packages/router/src/router.ts` (lines 354–405, 407–453)

## Expected Behaviour

The guard should be evaluated exactly once per navigation, regardless of whether the navigation was triggered by `back()`, `forward()`, or `navigate()`.

## Actual Behaviour

The guard is called twice: once explicitly in `back()`/`forward()`, and again inside `_executeNavigation`. This causes duplicate side effects and unpredictable behavior when guards have non-idempotent logic (redirects, abort conditions).

## Proposed Fix

Add a `skipGuard?: boolean` option to `_executeNavigation` and pass it from `back()` and `forward()` to prevent double evaluation. Alternatively, refactor `back()` and `forward()` to exclusively use `_executeNavigation` without manual guard checks:

```typescript
// packages/router/src/router.ts

back(): void {
    if (this._history.length <= 1) return;
    const prevPath = this._history[this._history.length - 2];
    if (!prevPath) {
        this.events.emit('back', null);
        return;
    }

    const poppedPath = this._history.pop();
    if (poppedPath) this._forwardStack.push(poppedPath);

    // Let _executeNavigation handle the guard exactly once
    this._executeNavigation(prevPath, {
        modifyHistory: 'none',
        clearForwardStack: false,
        direction: 'back',
    });
}

forward(): void {
    const nextPath = this._forwardStack.pop();
    if (!nextPath) {
        this.events.emit('forward', null);
        return;
    }

    this._history.push(nextPath);

    // Let _executeNavigation handle the guard exactly once
    this._executeNavigation(nextPath, {
        modifyHistory: 'none',
        clearForwardStack: false,
        direction: 'forward',
    });
}

// In _executeNavigation, remove any guard call from the back/forward
// code path — the guard is already evaluated in the shared path:
private _executeNavigation(path: string, options: NavigationOptions): void {
    const match = this.match(path);
    if (!match) return;

    // Guard evaluated exactly once here
    const guardResult = match.route.beforeEnter?.(path);
    if (guardResult === false) return;
    if (typeof guardResult === 'string') {
        this.navigate(guardResult);
        return;
    }

    // ... rest of navigation logic
}
```

ionfwsrijan:
## Description

The `back()` and `forward()` methods in the Router class call `_runBeforeEnterGuards()` themselves, then call `_executeNavigation()` which also calls `_runBeforeEnterGuards()`. This causes navigation guards to execute twice for every back/forward navigation, leading to duplicate side effects and performance issues.

## Fix

Added a `skipGuards` flag to `_executeNavigation` options. The `back()` and `forward()` methods now set `skipGuards: true` when calling `_executeNavigation()` since they already run guards themselves.

## Changes

- `packages/router/src/router.ts`: Added `skipGuards` option to `_executeNavigation`, updated `back()` and `forward()` to pass `skipGuards: true`

## Checklist

- [x] I have starred the repo

Closes #2867

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved navigation guard handling during redirects and browser history navigation.
  - Prevented parent route guards from running more than once within a single navigation chain.
  - Ensured guards on redirected-to routes execute consistently during navigation, back, and forward actions.

### DIFF
--- a/packages/router/src/router.ts
+++ b/packages/router/src/router.ts
@@ -244,13 +244,17 @@ export class Router {
         };
     }
 
-    private _runBeforeEnterGuards(match: RouteMatch, path: string, navigation: NavigationGuardContext): boolean | string {
+    private _runBeforeEnterGuards(match: RouteMatch, path: string, navigation: NavigationGuardContext, guardedPaths?: Set<string>): boolean | string {
         for (const route of match.chain) {
+            if (guardedPaths?.has(route.path)) {
+                continue;
+            }
             const result = route.beforeEnter?.(path, navigation);
             if (navigation.isStale()) return false;
             if (result === false || typeof result === 'string') {
                 return result;
             }
+            guardedPaths?.add(route.path);
         }
         return true;
     }
@@ -266,6 +270,7 @@ export class Router {
             clearForwardStack?: boolean;
             direction?: 'push' | 'replace' | 'back' | 'forward';
             navigation?: NavigationGuardContext;
+            guardedPaths?: Set<string>;
         } = {},
     ): void {
         const navigation = options.navigation ?? this._beginNavigation();
@@ -318,15 +323,16 @@ export class Router {
             this._forwardStack = [];
         }
 
-        const guardResult = this._runBeforeEnterGuards(match, resolvedPath, navigation);
+        const guardedPaths = options.guardedPaths ?? new Set<string>();
+        const guardResult = this._runBeforeEnterGuards(match, resolvedPath, navigation, guardedPaths);
         if (navigation.isStale()) return;
 
-        if (guardResult === false) {
-            return;
-        }
+            if (guardResult === false) {
+                return;
+            }
 
         if (typeof guardResult === 'string') {
-            this._executeNavigation(guardResult, { ...options, clearForwardStack: false, navigation });
+            this._executeNavigation(guardResult, { ...options, clearForwardStack: false, navigation, guardedPaths });
             return;
         }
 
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

Karanjot786:
Thanks — the double-guard bug is genuinely fixed, but the approach introduces a guard-BYPASS regression.

`skipGuards: true` (router.ts:383/403/437/458) disables the ENTIRE guard block in `_executeNavigation` (router.ts:298), so when a back()/forward()-triggered navigation REDIRECTS, the redirect target route’s own `beforeEnter` guard no longer runs at all (verified: target-guard call count drops 1→0). An auth/permission guard on the redirect destination is silently skipped. The existing test only redirects to an unguarded route, so it misses this. Fix: dedupe guards already run in the chain rather than disabling all guards; add a test where the redirect target has its own guard.

Checklist: [ ] redirect target guard still runs [ ] parent guard runs once (not twice) [ ] test covers guarded redirect target

ionfwsrijan:

PR #2890, opened 2026-07-22
fix: use ref to track latest asyncFn in useAsync to prevent stale closure

### LINKED ISSUE #2868: [bug] `useAsync` Captures Stale `asyncFn` Reference via Empty `useCallback` Deps
## Description

The `useAsync` hook wraps `refetch` in `useCallback(() => { ... }, deps)` where `deps` defaults to `[]`. The `refetch` closure captures `asyncFn` from the component's props/arguments. On re-renders, `useCallback` with `deps = []` returns the **original** `refetch` function, which holds a reference to the **first** `asyncFn` ever passed. If the caller changes `asyncFn` between renders (e.g., passing a new fetch function that queries a different endpoint), subsequent `refetch()` calls always invoke the stale original function, never the updated one.

## Affected Files

- `packages/jsx/src/hooks.ts` (lines 750–787)

## Expected Behaviour

`refetch` should always invoke the latest `asyncFn` passed to the hook. When the caller provides a new async function on re-render, calling `refetch()` should execute that new function.

## Actual Behaviour

`refetch` always invokes the `asyncFn` from the first render, regardless of what the caller passes on subsequent renders. The hook's state (`data`, `loading`, `error`) updates based on the stale function's result, not the current one. This leads to incorrect data being displayed when the data source changes between renders.

## Proposed Fix

Store `asyncFn` in a ref so the callback always reads the current value:

```typescript
// packages/jsx/src/hooks.ts

export function useAsync<T>(
    asyncFn: () => Promise<T>,
    deps: any[] = [],
): AsyncState<T> {
    const [data, setData] = useState<T | null>(null);
    const [loading, setLoading] = useState(true);
    const [error, setError] = useState<Error | null>(null);
    const versionRef = useRef(0);

    // Keep ref current with latest asyncFn
    const asyncFnRef = useRef(asyncFn);
    asyncFnRef.current = asyncFn;

    const refetch = useCallback(() => {
        const version = ++versionRef.current;
        setLoading(true);
        setError(null);

        asyncFnRef.current()  // Read from ref, not closure
            .then((result) => {
                if (versionRef.current === version) {
                    setData(result);
                    setLoading(false);
                }
            })
            .catch((err) => {
                if (versionRef.current === version) {
                    setError(err instanceof Error ? err : new Error(String(err)));
                    setLoading(false);
                }
            });
    }, deps);

    useEffect(() => {
        refetch();
    }, deps);

    return { data, loading, error, refetch };
}
```

This ensures the callback always reads the latest `asyncFn` from the ref while preserving the `useCallback` memoization behavior for downstream consumers.

ionfwsrijan:
## Description

The `useAsync` hook captures `asyncFn` in a closure via `useCallback`. When the component re-renders with a new `asyncFn` prop, the `refetch` callback still uses the stale original function, causing it to fetch stale data or fail.

## Fix

Added a ref to track the latest `asyncFn` function. The ref is updated on every render, and `refetch` always calls the current function from the ref instead of the captured closure.

## Changes

- `packages/jsx/src/hooks.ts`: Added `asyncFnRef` to track latest function, modified `refetch` to call `asyncFnRef.current()` instead of captured `asyncFn`

## Checklist

- [x] I have starred the repo

Closes #2868

### DIFF
--- a/packages/jsx/src/hooks.ts
+++ b/packages/jsx/src/hooks.ts
@@ -753,13 +753,16 @@ export function useAsync<T>(
 
     // Track a version counter to ignore stale responses
     const versionRef = useRef(0);
+    // Always call the latest asyncFn to avoid stale closure
+    const asyncFnRef = useRef(asyncFn);
+    asyncFnRef.current = asyncFn;
 
     const refetch = useCallback(() => {
         const version = ++versionRef.current;
         setLoading(true);
         setError(null);
 
-        asyncFn()
+        asyncFnRef.current()
             .then((result) => {
                 // Only update if this is still the latest request
                 if (versionRef.current === version) {


ionfwsrijan:

PR #2898, opened 2026-07-22
Stabilize equal z hit regions

### LINKED ISSUE #2882: [bug] LayerManager setHitRegion lets later equal-z regions override earlier ones
﻿## Summary
`LayerManager.setHitRegion()` replaces the current hit owner when `zVal >= existingZ`. This means two overlapping widgets with the same z-index are resolved by registration order, with the later region winning.

## Why this is surprising
For stable hit testing, equal z-index regions should usually keep the existing owner or follow an explicit DOM/render order rule. The current behavior can make pointer targets change when registration order changes, even though z-index did not.

## Reproduction
1. Register widget A at `z=1` over a cell.
2. Register widget B over the same cell, also at `z=1`.
3. `hitTest()` returns B because `>=` allows equal-z replacement.

## Expected behavior
Equal z-index behavior should be deterministic and documented. If earlier registration represents the front-most rendered widget, use `>` instead of `>=`; otherwise expose a clear ordering rule and tests.

## Suggested fix
Decide the intended equal-z precedence, update the comparison accordingly, and add regression tests for overlapping regions with equal and different z-index values.

Tomeshwari-02:
Summary
- keep the first hit-region owner when z-index values are equal
- still allow strictly higher z-index regions to replace lower ones
- update hit-test coverage for equal-z overlaps

Validation
- node_modules\.bin\vitest.exe run packages/core/src/terminal/LayerManager.test.ts --watch=false

Closes #2882

### DIFF
--- a/packages/core/src/terminal/LayerManager.test.ts
+++ b/packages/core/src/terminal/LayerManager.test.ts
@@ -37,7 +37,7 @@ describe('LayerManager — Hit Testing', () => {
         expect(lm.hitTest(7, 7)).toBe('bg');
     });
 
-    it('equal z-index overlapping regions resolve to the last written region', () => {
+    it('equal z-index overlapping regions keep the first owner stable', () => {
         const lm = new LayerManager(20, 10);
         lm.clearHitGrid();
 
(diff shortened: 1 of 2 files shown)


Karanjot786:
Thanks, but build-and-test fails. The `>=`→`>` change in `packages/core/src/terminal/LayerManager.ts:301` regresses equal-z tie-break: `packages/core/src/app/App.test.ts:572` expects `hitTest(0,0)` = `first` but now gets `root`. Your PR body shows you ran only `LayerManager.test.ts`, not the full suite. Run `bun vitest run` across packages/core and keep the existing tie-break semantics (a later equal-z sibling should win the hit).

Checklist: [ ] full core suite green [ ] App.test.ts hitTest passes

PR #2910, opened 2026-07-22
Add terminal capability negotiation

### LINKED ISSUE #2900: [feature] Add a terminal capability negotiation layer with runtime fallbacks
﻿## Summary
TermUI currently has several capability checks spread across rendering code, unicode handling, color behavior, mouse support, and environment flags. A centralized capability negotiation layer would make terminal-specific behavior easier to reason about and test.

## Problem
Different terminals support different combinations of:
- truecolor vs 256-color vs basic colors
- Unicode width behavior
- mouse tracking modes
- alternate screen support
- OSC/CSI escape behavior
- ambiguous-width glyph rendering

Right now, components and utilities tend to query global `caps` or environment helpers directly. This makes it difficult to run the same app against different terminal profiles, and hard to reproduce terminal-specific bugs in tests.

## Proposed work
Add a capability profile abstraction in `@termuijs/core` that can be detected at runtime and injected in tests. Rendering and widgets should consume that profile rather than reaching into scattered environment checks.

## Acceptance criteria
- A `TerminalCapabilities` profile exists with color, unicode, mouse, and escape-sequence support fields.
- Existing `caps` behavior remains backward-compatible.
- Tests can install a synthetic capability profile without mutating global process state.
- At least `Screen`, ANSI utilities, and one widget path use the new profile.
- Documentation describes how app authors can force or override capabilities.

Tomeshwari-02:
Summary
- Add a negotiated terminal capability profile for color, Unicode, mouse, alternate screen, escape sequences, motion, background, and keybindings.
- Rewire the existing caps API to read from the negotiated profile while keeping current imports compatible.
- Add deterministic test overrides for package-level checks.

Validation
- npx.cmd vitest run packages/core/src/terminal/capabilities.test.ts packages/core/src/terminal/env-caps.test.ts

Closes #2900

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -12,6 +12,19 @@ export type { FrameStats } from './terminal/Renderer.js';
 export { LayerManager } from './terminal/LayerManager.js';
 export type { Layer } from './terminal/LayerManager.js';
 export { caps, prefersReducedMotion, shouldUseColor, prefersHighContrast } from './terminal/env-caps.js';
+export {
+    detectTerminalCapabilities,
+    getTerminalCapabilities,
+    setTerminalCapabilitiesForTests,
+} from './terminal/capabilities.js';
+export type {
+    TerminalCapabilities,
+    TerminalCapabilityEnv,
+    TerminalCapabilityOptions,
+    TerminalColorLevel,
+    TerminalBackground,
+    TerminalKeybindingMode,
+} from './terminal/capabilities.js';
 export { BOX, BRAILLE_SPIN, BLOCK } from './terminal/ascii-map.js';
 
 // ── Renderer ──────────────────────────────────────────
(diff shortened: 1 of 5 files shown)


Karanjot786:
Thanks — the capability negotiator is well-wired, but it breaks existing tests. Converting `caps.unicode` from an assignable field to a getter-only accessor makes `packages/widgets/src/display/Badge.test.ts:82` and `Tag.test.ts:88` throw `TypeError: Cannot set property unicode of #<Object> which has only a getter` (they do `(caps as any).unicode = false` to force ASCII fallback). build-and-test fails for real. Either keep `caps.unicode` assignable (setter that overrides detection), or update those tests to a supported override mechanism (and any others that assign caps.*).

Checklist: [ ] caps.* still overridable in tests (or provide + use a documented override) [ ] Badge/Tag tests green [ ] full build-and-test green

PR #2911, opened 2026-07-22
Add incremental layout invalidation reasons

### LINKED ISSUE #2901: [feature] Add incremental layout invalidation across nested layout trees
﻿## Summary
The layout engine should support fine-grained invalidation so changes to one subtree do not force unrelated layout branches to recompute.

## Problem
Complex terminal dashboards can contain deep layout trees with static navigation, tables, panels, modals, and live data regions. A small state change in one leaf should not require full-tree layout recalculation when parent constraints and sibling geometry are unchanged.

## Proposed work
Introduce layout dirty reasons and subtree invalidation boundaries. The engine should distinguish between changes that affect only paint, local layout, parent-dependent dimensions, or global layout.

## Acceptance criteria
- Layout nodes track why they are dirty, not only whether they are dirty.
- Local-only changes avoid recomputing unaffected sibling subtrees.
- Parent size or constraint changes still invalidate dependent descendants correctly.
- Benchmarks cover a large dashboard tree before and after the change.
- Tests cover nested clean/dirty mixes, percentage dimensions, and absolute/fixed layouts.

Tomeshwari-02:
Summary
- Add typed layout dirty reasons and invalidation options.
- Preserve full subtree invalidation by default while allowing local invalidation for nested layout trees.
- Ensure clean ancestors still walk dirty descendants during layout recomputation.

Validation
- npx.cmd vitest run packages/core/src/layout/LayoutEngine.test.ts

Closes #2901

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -33,7 +33,7 @@ export type { ChordMatcherOptions, Chord } from './input/ChordMatcher.js';
 export { LiveRender } from './renderer/live-render.js';
 // ── Layout ────────────────────────────────────────────
 export { computeLayout, createLayoutNode, invalidateLayout } from './layout/LayoutEngine.js';
-export type { LayoutNode } from './layout/LayoutEngine.js';
+export type { LayoutDirtyReason, LayoutInvalidationOptions, LayoutNode } from './layout/LayoutEngine.js';
 export { emptyRect, containsPoint, shrinkRect, intersectRect, unionRect } from './layout/Rect.js';
 export type { Rect, Size } from './layout/Rect.js';
 export { Pos } from './layout/pos.js';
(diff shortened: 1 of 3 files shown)


Karanjot786:
Thanks, but two blockers. (1) `bench` fails for real — layout-computation (10 nodes) regresses ~-24.6%: the new `!hasDirtyChild(node)` check in the fast-path bailout of `layoutNode()` (packages/core/src/layout/LayoutEngine.ts:132-137) turns an O(1) clean-node skip into an O(subtree) walk on every call. (2) `_dirtyReason`/`LayoutDirtyReason` is tracked + exported but has zero consumers — no renderer/paint path branches on it, so it is dead metadata. Make the reason drive an actual optimization (or drop it), and keep the clean-node skip O(1).

Checklist: [ ] no bench regression [ ] dirty-reason actually consumed, or removed

PR #2913, opened 2026-07-22
docs(jsx): document useKeymap duplicate-key dev warning

### LINKED ISSUE #2729: [feature] Add exportToCSV/exportToJSON utils for Table & DataGrid
Description
The README states that multiple useKeymap calls in one component are additive, but there is no warning or error when two calls register the same key. The second registration silently overwrites the first, making one handler unreachable with no indication to the developer.

Steps to Reproduce
function MyApp() {
  useKeymap({
    '/': () => console.log('open search'),
  })

  // Added in a different hook or conditional — same key, different intent
  useKeymap({
    '/': () => console.log('open filter'),  // silently wins
  })

  return <Box>...</Box>
}
Expected Behavior
One of:

A development-mode warning is logged to stderr: [TermUI] Duplicate keymap binding: "/" registered more than once in the same component. Last registration wins.
Or a strict mode option on useKeymap that throws on collision: useKeymap({ '/' : handler }, { strict: true })
Actual Behavior
The second handler silently replaces the first. No warning. No error. The first handler is dead code with no indication why.

Why This Matters
This is especially painful in large components that compose multiple hooks or mixins, where keymap registrations are spread across the file. Silent overwrites are a common source of subtle, hard-to-debug interaction bugs.

Suggested Fix
In the useKeymap implementation, before merging the new bindings into the active keymap, check for key overlap with already-registered bindings in the same component scope and emit a console.warn in non-production environments.

Environment
@termuijs/jsx (any version)
Node.js 18+

Kokila-chandrakar:
Closes #2729

## Context
Issue #2729 reports that the README says multiple `useKeymap` calls in
one component are additive, but there's no warning when two calls
register the same key — leaving one handler silently dead.

## What I found
Digging into `packages/jsx/src/hooks.ts`, this detection **already
exists** — it was added in #1830. `useKeymap` tracks bindings per
render via `fiber._keymapKeys` and logs a `console.warn` in
non-production builds whenever a key collides, both within a single
call and across multiple calls in the same component. There's also
existing test coverage in `hooks.test.ts` for this exact behavior.

So the runtime behavior in #2729 is not actually missing — what's
missing is that **none of the docs mention it**. The README's
"additive" claim with no follow-up about collisions is exactly the gap
the issue is describing.

## What this PR changes (docs only, no logic changes)
- `packages/jsx/src/hooks.ts`: expanded the `useKeymap` JSDoc to
  describe the duplicate-key warning and show the exact console.warn
  message format.
- `packages/jsx/README.md`: added a "Duplicate keys" section under
  `useKeymap` with a runnable example.
- `README.md`: added a short pointer from the root useKeymap section
  to the detailed explanation.

## Testing
Docs-only change — no runtime code touched. Existing tests in
`packages/jsx/src/hooks.test.ts`
(`describe('useKeymap — cross-call duplicate detection (dev-mode)')`)
already cover the warning behavior being documented here.

## Summary by CodeRabbit

* **Documentation**
  * Documented that multiple keymap declarations are additive.
  * Added guidance about duplicate key bindings and development-time console warnings.
  * Clarified that the latest registration takes precedence while the application continues running.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,8 @@ function App() {
 }
 ```
 
+Bindings across multiple `useKeymap` calls in the same component are additive, so it's possible to accidentally register the same key twice. In development, `useKeymap` detects this and logs a `console.warn` naming the key instead of silently letting the later registration win — see the [`@termuijs/jsx` README](./packages/jsx#usekeymap) for details.
+
 ### Focus management
 
 Four hooks for building keyboard-accessible interfaces:
--- a/packages/jsx/README.md
+++ b/packages/jsx/README.md
@@ -90,6 +90,19 @@ function App() {
 
 Modifier syntax: `ctrl+`, `alt+`, `shift+`, `meta+`. Combine them: `ctrl+shift+k`.
 
+**Duplicate keys:** because bindings across calls are additive, it's possible to accidentally register the same key twice (e.g. from two different hooks or conditionals). In non-production builds, `useKeymap` detects this — both within a single call and across multiple calls in the same component — and logs a `console.warn` naming the colliding key, since the second registration silently wins otherwise:
+
+\`\`\`tsx
+useKeymap([{ key: '/', action: () => openSearch() }])
+
+// Registered elsewhere in the same component, same key:
+useKeymap([{ key: '/', action: () => openFilter() }])
+// console.warn: [useKeymap] Duplicate keymap binding: "/|false|false|false"
+// registered more than once in the same component. Last registration wins.
+\`\`\`
+
+This is a warning, not an error — the app keeps running with the last handler active — but it flags the collision instead of leaving the first handler as silent dead code.
+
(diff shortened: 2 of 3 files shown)


Karanjot786:
Thanks, but the documented behavior is partly wrong. The new JSDoc/README say duplicate keys mean "the last binding wins at runtime" — that is only true for duplicates across SEPARATE useKeymap calls. For a duplicate key WITHIN one useKeymap call, the FIRST binding fires (the onInput match loop returns on first hit, hooks.ts:384-391). The source’s own console.warn text (hooks.ts:361) has the same inaccuracy. Please document both cases accurately (first-wins within a call, last-wins across calls) and ideally fix the warn text to match.

Checklist: [ ] docs state first-wins-within-call and last-wins-across-calls [ ] warn text corrected

PR #2914, opened 2026-07-22
Add context selector subscriptions

### LINKED ISSUE #2903: [feature] Add selector-based context subscriptions to reduce JSX re-renders
﻿## Summary
Context consumers should be able to subscribe to a selected slice of context value, so unrelated context changes do not re-render every consumer.

## Problem
Large terminal apps commonly put theme, routing, focus, viewport, and app state in context. If every provider update re-renders all consumers, dashboards with many widgets can become expensive even when most consumers only need one small field.

## Proposed API
Add a selector-capable hook, for example:
```ts
const activeTheme = useContextSelector(ThemeContext, value => value.activeTheme);
```

The consumer should re-render only when the selected value changes according to an equality function.

## Acceptance criteria
- `useContextSelector()` exists and works with existing context providers.
- Optional equality function support is included.
- Existing `useContext()` remains unchanged.
- Tests cover unchanged selected values, changed selected values, provider replacement, and nested providers.
- A benchmark or stress test demonstrates fewer rerenders in a many-consumer tree.

Tomeshwari-02:
Summary
- Add useContextSelector with optional equality comparison.
- Store context subscriptions with selector metadata so provider updates only schedule affected consumers.
- Update fiber teardown to remove selector subscription records cleanly.

Validation
- npx.cmd vitest run packages/jsx/src/context.test.ts packages/jsx/src/reconciler.test.ts

Closes #2903

### DIFF
--- a/packages/jsx/src/context.test.ts
+++ b/packages/jsx/src/context.test.ts
@@ -2,10 +2,10 @@
 // Tests — Context API (createContext / useContext)
 // ─────────────────────────────────────────────────────
 
-import { describe, it, expect, beforeEach } from 'vitest';
-import { createContext, useContext } from './context.js';
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { createContext, useContext, useContextSelector } from './context.js';
 import {
-    createFiber, setCurrentFiber, clearCurrentFiber,
+    createFiber, setCurrentFiber, clearCurrentFiber, setRequestRender,
     type Fiber,
 } from './hooks.js';
 
@@ -32,6 +32,7 @@ describe('useContext', () => {
     });
 
     afterEach(() => {
+        setRequestRender(null);
         clearCurrentFiber();
     });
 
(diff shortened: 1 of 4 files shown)


Karanjot786:
Thanks, but `useContextSelector` leaks subscriptions. The dedup check `sub.selector === selector` (packages/jsx/src/context.ts ~line 100) compares function references, but selectors are typically inline arrows (`v => v.count`) that get a NEW reference every render while the fiber persists — so every re-render appends a new ContextSubscription that is never matched/reused, growing the provider Set unboundedly until unmount. This also undermines the "only re-render on selected slice change" guarantee and likely explains the failing bench check. Key the subscription by fiber (dedupe on fiber, update its selector), or store the selector in a hook slot so its identity is stable. Add a test that re-renders a selector consumer N times and asserts the subscriber count stays constant.

Checklist: [ ] no per-render subscription growth [ ] re-render count test [ ] bench green

PR #2943, opened 2026-07-23
fix(widgets): clamp restored ratio in SplitPane.loadLayout

### LINKED ISSUE #2927: [bug] `SplitPane.loadLayout()` does not call `markDirty()` after restoring state
## Description

In `packages/widgets/src/layout/SplitPane.ts`, the `loadLayout` method (line 155) restores `_ratio` and `_direction` from serialized data but never calls `markDirty()`:

```ts
loadLayout(data: string): void {
    try {
        const layout = JSON.parse(data);
        let changed = false;

        if (typeof layout.ratio === 'number' && layout.ratio !== this._ratio) {
            this._ratio = layout.ratio;
            changed = true;
        }
        if (layout.direction === 'horizontal' || layout.direction === 'vertical') {
            if (layout.direction !== this._direction) {
                this._direction = layout.direction;
                changed = true;
            }
        }
        if (changed) {
            this.markDirty();  // ← This IS called
        }
    } catch {
        // Ignore malformed layout data
    }
}
```

Actually, `markDirty()` is called when `changed` is true. However, the `_positionChildren()` call that re-lays-out the two panes is only triggered via `syncLayout()` → `_positionChildren()`. The `markDirty()` marks the widget as needing re-render, but `syncLayout()` is called by the layout engine during the render pass. The real problem is that `loadLayout` updates `_ratio` and `_direction` **without updating the children's rects**. The children's rects are only updated in `syncLayout()` → `_positionChildren()`, which runs during the next render. But if the SplitPane is not re-rendered (e.g., it was already clean), the children retain their old rects until the next full layout pass.

The **actual bug** is more subtle: `loadLayout` calls `markDirty()` which schedules a re-render, but it does **not** call `_positionChildren()` directly. If the SplitPane's parent does not trigger a full re-layout (e.g., the parent is also clean), the children's rects remain stale. The `setRatio` method has the same pattern but works correctly because `markDirty()` triggers a full re-render which includes `syncLayout()`.

The real issue is that `loadLayout` updates `_direction` without ensuring the children are re-positioned. When switching from `'horizontal'` to `'vertical'` (or vice versa), the divider character changes and the children need completely different rect assignments. The `markDirty()` → re-render → `syncLayout()` path handles this correctly.

After careful analysis, the **actual bug** is: when `loadLayout` is called and `changed` is `true`, `markDirty()` is called. But when `changed` is `false` (ratio and direction match), no `markDirty()` is called — which is correct. The issue is that `loadLayout` **does not clamp the restored ratio** against `_minSize`. A malicious or corrupted layout file could set `ratio: 0` or `ratio: 1`, collapsing one pane to zero width. The `setRatio` method applies `_clampRatio`, but `loadLayout` bypasses it by assigning directly to `this._ratio`.

## Expected Behaviour

`loadLayout` should clamp the restored ratio using `_clampRatio` (same as `setRatio`) to prevent collapsed panes, and should call `markDirty()` to ensure the layout is recalculated.

## Actual Behaviour

A corrupted or malicious layout file can set `ratio: 0` or `ratio: 1`, collapsing one pane to zero size. The children's rects are calculated with `Math.floor(ratio * width)`, so `ratio: 0` gives the first pane 0 width and the second pane the full width minus the divider — but the first pane's child still exists and may attempt to render into a 0-width rect.

## Proposed Fix

```ts
// SplitPane.ts — loadLayout method
loadLayout(data: string): void {
    try {
        const layout = JSON.parse(data);

        let changed = false;

        if (typeof layout.ratio === 'number' && layout.ratio !== this._ratio) {
            const content = this._getContentRect();
            const totalSize = this._direction === 'horizontal' ? content.width : content.height;
            this._ratio = totalSize > 0
                ? this._clampRatio(layout.ratio, totalSize)
                : layout.ratio;
            changed = true;
        }

        if (
            layout.direction === 'horizontal' ||
            layout.direction === 'vertical'
        ) {
            if (layout.direction !== this._direction) {
                this._direction = layout.direction;
                changed = true;
            }
        }

        if (changed) {
            this.markDirty();
        }
    } catch {
        // Ignore malformed layout data
    }
}
```

ionfwsrijan:
### Problem
`loadLayout()` restored the ratio from serialized data without validating it against `_clampRatio`, allowing corrupted or malicious layout files to set ratio to 0 or 1, collapsing a pane to zero width/height.

### Fix
Apply `_clampRatio()` to the restored ratio using the current content rect dimensions, matching the behavior of `setRatio()`.

Closes #2927

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved split-pane layout restoration by clamping saved divider positions to the widget’s current size.
  * Prevented restored layouts from placing the divider outside the available content area.

### DIFF
--- a/packages/widgets/src/layout/SplitPane.ts
+++ b/packages/widgets/src/layout/SplitPane.ts
@@ -162,7 +162,11 @@ export class SplitPane extends Widget {
                 typeof layout.ratio === 'number' &&
                 layout.ratio !== this._ratio
             ) {
-                this._ratio = layout.ratio;
+                const content = this._getContentRect();
+                const totalSize = this._direction === 'horizontal' ? content.width : content.height;
+                this._ratio = totalSize > 0
+                    ? this._clampRatio(layout.ratio, totalSize)
+                    : layout.ratio;
                 changed = true;
             }
 


ionfwsrijan:

PR #2944, opened 2026-07-23
fix(store): merge batch changes in mutate commit instead of replacing state

### LINKED ISSUE #2928: [bug]  `store.mutate()` batch commit closure captures stale `changes` reference
## Description

In `packages/store/src/store.ts`, the `mutate` method (line 505) creates a batch entry with a `commit` closure that captures the local `changes` variable:

```ts
const mutate = (recipe: (draft: T) => void): void => {
    const prevState = state;
    const baseState: T = _batchDepth > 0 ? _batchStores.get(listeners)?.nextState ?? state : state;
    const nextState = produce(baseState, (draft) => {
        recipe(draft as T);
    });
    // ...
    if (_batchDepth > 0) {
        const existing = _batchStores.get(listeners);
        const changedKeys = Object.keys(nextState).filter(
            k => !Object.is((nextState as any)[k], (baseState as any)[k])
        );
        const changes: Partial<T> = {};
        for (const k of changedKeys) {
            (changes as any)[k] = (nextState as any)[k];
        }
        if (!existing) {
            _batchStores.set(listeners, {
                prevState,
                nextState,
                changes,
                commit: () => { state = { ...state, ...changes } as T; persistState(); return state; },
                rollback: () => { state = prevState; },
            });
        } else {
            Object.assign(existing.changes, changes);
            existing.nextState = nextState;
            existing.commit = () => { state = nextState; persistState(); };
            existing.rollback = () => { state = existing.prevState; };
        }
    }
};
```

When the `existing` branch is taken (a second `mutate` call within the same batch), `existing.changes` is updated via `Object.assign`, and `existing.commit` is replaced with a new closure that captures `nextState` directly. However, the new `commit` uses `state = nextState` (assignment) instead of `state = { ...state, ...existing.changes }` (merge). This means if a `setState` call happened between the two `mutate` calls within the same batch, the `setState` changes are overwritten by the `mutate` commit.

The `setState` batch path (line 419) uses `state = { ...state, ...changes }` which correctly merges. The `mutate` batch path uses `state = nextState` which replaces entirely.

## Expected Behaviour

The `mutate` batch commit should merge changes onto the current batch state (same as `setState`), not replace it entirely, so that interleaved `setState` and `mutate` calls within the same batch are all preserved.

## Actual Behaviour

When `mutate` is called after `setState` within the same batch, the `mutate` commit overwrites the `setState` changes because it assigns `state = nextState` (the state after only the `mutate` recipe) instead of merging onto the accumulated batch state.

## Proposed Fix

```ts
// store.ts — mutate method, existing branch
existing.commit = () => {
    state = { ...state, ...existing.changes } as T;
    persistState();
    return state;
};
```

ionfwsrijan:
### Problem
The mutate batch commit closure captured `nextState` and assigned it directly (`state = nextState`), overwriting any interleaved `setState` changes within the same batch. The `setState` batch path correctly uses `state = { ...state, ...changes }`.

### Fix
Change `existing.commit` to use the merge pattern `{ ...state, ...existing.changes }` to match `setState`'s batch behavior.

Closes #2928

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency when applying batched state updates.
  * Prevented unrelated state values from being overwritten during batched mutations.

### DIFF
--- a/packages/store/src/store.ts
+++ b/packages/store/src/store.ts
@@ -533,7 +533,7 @@ export function createStore<T extends object>(
             } else {
                 Object.assign(existing.changes, changes);
                 existing.nextState = nextState;
-                existing.commit = () => { state = nextState; persistState(); };
+                existing.commit = () => { state = { ...state, ...existing.changes } as T; persistState(); return state; };
                 existing.rollback = () => { state = existing.prevState; };
             }
         } else {


ionfwsrijan:

PR #2945, opened 2026-07-23
fix(jsx): guard runEffects/runLayoutEffects against effect exceptions

### LINKED ISSUE #2929: [bug] `runEffects` does not guard against effect exceptions — failed effect re-runs infinitely
## Description

In `packages/jsx/src/hooks.ts`, the `runEffects` function (line 589) runs each pending effect and sets `record.ran = true` after execution:

```ts
export function runEffects(fiber: Fiber): void {
    for (const record of fiber.effects) {
        if (!record.ran) {
            record.cleanup?.();
            const cleanup = record.effect();
            if (typeof cleanup === 'function') {
                record.cleanup = cleanup;
            }
            record.ran = true;
        }
    }
}
```

If `record.effect()` throws an exception, `record.ran = true` is never reached. On the next render cycle, the effect's deps may not have changed, so `useEffect` does not reset `record.ran` to `false`. But the effect record is still in `fiber.effects` with `ran: false`, so `runEffects` tries to run it again. If the effect consistently throws (e.g., a network call that always fails, a ref that is always null), this creates an infinite re-run loop: every render triggers the effect, the effect throws, `ran` stays `false`, the component is marked dirty, and the cycle repeats.

The `runLayoutEffects` function (line 603) has the same issue.

## Expected Behaviour

If an effect throws, it should be marked as `ran = true` to prevent infinite re-execution, and the error should be surfaced through the error boundary or `onError` handler.

## Actual Behaviour

A throwing effect is never marked as ran, causing it to re-execute on every render cycle. This creates an infinite render loop that freezes the terminal application.

## Proposed Fix

```ts
// hooks.ts — runEffects
export function runEffects(fiber: Fiber): void {
    for (const record of fiber.effects) {
        if (!record.ran) {
            try {
                record.cleanup?.();
                const cleanup = record.effect();
                if (typeof cleanup === 'function') {
                    record.cleanup = cleanup;
                }
            } catch (err) {
                // Mark as ran to prevent infinite re-execution
                console.error('[useEffect] Effect threw:', err);
            }
            record.ran = true;
        }
    }
}

// hooks.ts — runLayoutEffects
export function runLayoutEffects(fiber: Fiber): void {
    for (const record of fiber.layoutEffects) {
        if (!record.ran) {
            try {
                record.cleanup?.();
                const cleanup = record.effect();
                if (typeof cleanup === 'function') {
                    record.cleanup = cleanup;
                }
            } catch (err) {
                console.error('[useLayoutEffect] Effect threw:', err);
            }
            record.ran = true;
        }
    }
}
```

ionfwsrijan:
### Problem
If an effect threw an exception, `record.ran` was never set to `true`, causing the effect to re-execute on every render cycle in an infinite loop that freezes the terminal.

### Fix
Wrap effect execution in try/catch blocks in both `runEffects` and `runLayoutEffects`. Always set `record.ran = true` after the try/catch to prevent infinite re-execution. Log the error for debugging.

Closes #2929

### DIFF
--- a/packages/jsx/src/hooks.ts
+++ b/packages/jsx/src/hooks.ts
@@ -589,11 +589,16 @@ export function useReducer<S, A>(
 export function runEffects(fiber: Fiber): void {
     for (const record of fiber.effects) {
         if (!record.ran) {
-            // Run cleanup from previous effect
-            record.cleanup?.();
-            const cleanup = record.effect();
-            if (typeof cleanup === 'function') {
-                record.cleanup = cleanup;
+            try {
+                // Run cleanup from previous effect
+                record.cleanup?.();
+                const cleanup = record.effect();
+                if (typeof cleanup === 'function') {
+                    record.cleanup = cleanup;
+                }
+            } catch (err) {
+                // Mark as ran to prevent infinite re-execution
+                console.error('[useEffect] Effect threw:', err);
             }
             record.ran = true;
         }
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #2946, opened 2026-07-23
fix(store): call persistState in batch rollback to sync disk state

### LINKED ISSUE #2930: [bug] Batch rollback does not call `persistState()`, leaving disk state out of sync
## Description

In `packages/store/src/store.ts`, the `BatchEntry` interface defines a `rollback` function that restores the in-memory state:

```ts
rollback: () => { state = existing.prevState; },
```

When a batch function throws, `flushBatch(threw=true)` calls `rollback()` on every batch entry (line 106-108):

```ts
if (threw) {
    for (const [, { rollback }] of _batchStores) {
        rollback();
    }
    _batchStores.clear();
}
```

The rollback correctly restores `state` to `prevState`. However, it does **not** call `persistState()`. If the store has persistence enabled (`options.persist`), the disk file may have been updated by a previous `setState` call within the batch (before the throw), and the rollback does not write the restored state back to disk.

Sequence of events:
1. Batch begins
2. `setState({ count: 1 })` — batch entry created, `commit` captures `{ count: 1 }`
3. `persistState()` is **not** called (deferred by batch)
4. An operation inside the batch throws
5. `flushBatch(true)` calls `rollback()` — `state` restored to `{ count: 0 }`
6. But the persist file was never written, so it either has the old value `{ count: 0 }` (no issue) **or** if `flushBatch(false, true)` was called earlier for an outer batch, the file has `{ count: 1 }` and is never corrected

The bug manifests when nested batches are used: an inner batch throws, the outer batch's `flushBatch` has already committed and persisted a partial state, and the inner rollback restores in-memory state but does not undo the disk write.

## Expected Behaviour

`rollback()` should call `persistState()` after restoring the in-memory state, so the disk file reflects the rolled-back state.

## Actual Behaviour

After a batch rollback, the in-memory state is correct but the disk file may contain the pre-rollback state, causing stale data to be loaded on next app startup.

## Proposed Fix

```ts
// store.ts — in the setState batch branch (line 420)
rollback: () => {
    state = existing.prevState;
    persistState();
},

// store.ts — in the mutate batch branch (line 531)
rollback: () => {
    state = existing.prevState;
    persistState();
},

// store.ts — in the mutate batch existing branch (line 537)
existing.rollback = () => {
    state = existing.prevState;
    persistState();
};
```

ionfwsrijan:
### Problem
Batch rollback restored in-memory state but never called `persistState()`, leaving the persist file out of sync after a nested batch throw. On next app startup, stale data would be loaded.

### Fix
Add `persistState()` to all four rollback closures (two in setState batch path, two in mutate batch path).

Closes #2930

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed batched state rollbacks so persisted storage now correctly reflects the restored state.
  * Ensured rollback behavior is consistent across state updates and mutations.

### DIFF
--- a/packages/store/src/store.ts
+++ b/packages/store/src/store.ts
@@ -417,13 +417,13 @@ export function createStore<T extends object>(
                             nextState,
                             changes,
                             commit: () => { state = { ...state, ...changes } as T; persistState(); return state; },
-                            rollback: () => { state = prevState; },
+                            rollback: () => { state = prevState; persistState(); },
                         });
                     } else {
                         Object.assign(existing.changes, finalPartial);
                         existing.nextState = nextState;
                         existing.commit = () => { state = { ...state, ...existing.changes } as T; persistState(); return state; };
-                        existing.rollback = () => { state = existing.prevState; };
+                        existing.rollback = () => { state = existing.prevState; persistState(); };
                     }
                 } else {
                     state = nextState; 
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #2947, opened 2026-07-23
fix(ui): validate current step in Wizard.complete before invoking callback

### LINKED ISSUE #2942: [bug] `Wizard.complete()` bypasses step validation on final step
## Description

In `packages/ui/src/Wizard.ts`, the `complete()` method (line 116) invokes the completion callback without running the current step's `validate()` function:

```ts
complete(): void {
    this._onComplete?.(this._getStepData());
}
```

When the user presses Enter on the last step (lines 196-202), `complete()` is called directly:

```ts
case 'enter':
case 'return':
    if (this._currentStepIndex === this._steps.length - 1) {
        this.complete();
        event.stopPropagation?.();
    }
    break;
```

Meanwhile, `nextStep()` (lines 87-100) **does** run validation — but the guard at line 104 prevents advancing past the last step:

```ts
if (this._currentStepIndex < this._steps.length - 1) {
    this._showStep(this._currentStepIndex + 1);
}
```

This means the last step's `validate()` function is **never enforced** through keyboard input. Right/N on the last step triggers validation but can't advance (blocked by the length guard). Enter on the last step skips validation entirely. Invalid data can always be submitted on the final step of any wizard.

## Expected Behaviour

`complete()` should run the current step's `validate()` before invoking `_onComplete`, matching the validation logic in `nextStep()`.

## Actual Behaviour

The last step's `validate()` is never called on the completion path. Any validation defined on the final step is silently ignored.

## Proposed Fix

```ts
// Wizard.ts — complete method
complete(): void {
    const step = this._steps[this._currentStepIndex];
    if (step && step.validate) {
        const validationResult = step.validate();
        if (typeof validationResult === 'string') {
            this._error = validationResult;
            this.markDirty();
            return;
        } else if (validationResult === false) {
            this.markDirty();
            return;
        }
    }

    this._error = '';
    this._onComplete?.(this._getStepData());
}
```

ionfwsrijan:
### Problem
`complete()` invoked the completion callback without running the current step's `validate()` function. When the user pressed Enter on the last step, validation was silently bypassed, allowing invalid data to be submitted.

### Fix
Add the same validation logic used by `nextStep()` to `complete()`. If validation returns a string error message or `false`, block completion and mark dirty.

Closes #2942

### DIFF
--- a/packages/ui/src/Wizard.ts
+++ b/packages/ui/src/Wizard.ts
@@ -114,6 +114,20 @@ export class Wizard extends Widget {
     }
 
     complete(): void {
+        const step = this._steps[this._currentStepIndex];
+        if (step && step.validate) {
+            const validationResult = step.validate();
+            if (typeof validationResult === 'string') {
+                this._error = validationResult;
+                this.markDirty();
+                return;
+            } else if (validationResult === false) {
+                this.markDirty();
+                return;
+            }
+        }
+
+        this._error = '';
         this._onComplete?.(this._getStepData());
     }
 


ionfwsrijan:

PR #2965, opened 2026-07-23
Pass reasons through motion cancellation

### LINKED ISSUE #2954: [feature] Add cancellation reason support to motion runners
## Summary
Motion helpers expose cancel functions, but callers cannot tell why a runner stopped. A cancellation reason would help compose sequence/repeat/parallel flows and write clearer cleanup tests.

## Scope
Extend `packages/motion/src/sequence.ts` and related tests with optional cancellation reason support.

## Proposed behavior
- Returned cancel functions may accept an optional reason string.
- `sequence`, `parallel`, and `repeat` pass that reason to child cancellers when available.
- Existing no-argument cancel callbacks keep working.

## Acceptance criteria
- Add tests for sequence cancellation before and during a child runner.
- Add tests for parallel forwarding the same reason to all active children.
- Add tests proving `onComplete` is not called after cancellation.
- No breaking change for existing `() => void` cancellers.

Tomeshwari-02:
## Summary
- add an AnimationCancel type that accepts an optional reason
- pass cancellation reasons through sequence, parallel, and repeat
- export the cancellation type from the motion package

## Tests
- bun test packages/motion/src/sequence.test.ts packages/motion/src/repeat.test.ts

Closes #2954

## Summary by CodeRabbit

- **New Features**
  - Animation cancellation now supports optional reason messages.
  - Cancellation reasons are consistently forwarded through sequences, parallel animations, repeats, staggered animations, and transitions.
  - Transition callbacks can respond to cancellation reasons across built-in effects such as fade, slide, typewriter, and pulse.

- **Tests**
  - Added coverage confirming cancellation reasons propagate correctly across animation utilities and transition helpers.

### DIFF
--- a/packages/motion/src/index.ts
+++ b/packages/motion/src/index.ts
@@ -18,7 +18,7 @@ export { transition, fadeIn, fadeOut, slideIn, typewriter, pulse, easings, cubic
 export type { TransitionOptions, EasingFn } from './transitions.js';
 // Sequencing
 export { sequence, parallel, repeat } from './sequence.js';
-export type { AnimationRunner, SequenceStep, AnimatableValue, RepeatOptions } from './sequence.js';
+export type { AnimationCancel, AnimationRunner, SequenceStep, AnimatableValue, RepeatOptions } from './sequence.js';
 export { stagger } from './stagger.js';
 // Shared interval timer pool
 export { subscribe as timerPoolSubscribe, unsubscribeAll as timerPoolUnsubscribeAll } from './timer-pool.js';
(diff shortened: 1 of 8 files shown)


Karanjot786:
Blocking — the `reason` is dead metadata. This threads an optional `reason` through `sequence`/`parallel`/`repeat` cancel functions, but nothing consumes it: leaf runners (`transitions.ts` fadeIn/out/slide/typewriter/pulse) and `stagger.ts` drop it, and no widget or `onCancel(reason)` hook reads it. Same pattern as the previously-blocked #2911 dirty-reason. Either wire a real consumer (e.g. an onCancel(reason) callback that something uses) or drop the plumbing. Checklist: [ ] reason actually consumed, or removed.

PR #2982, opened 2026-07-24
fix(jsx): detect deps length changes in useEffect and useLayoutEffect

### LINKED ISSUE #2977: [bug] useEffect/useLayoutEffect Deps Comparison Misses Length Changes
## Description

In `packages/jsx/src/hooks.ts`, both `useEffect` (line 266) and `useLayoutEffect` (line 290) compare dependencies by iterating over the new deps array with `deps.some()`:

```typescript
// packages/jsx/src/hooks.ts:266
const shouldRun = !deps || !prev.deps || deps.some((d, i) => !Object.is(d, prev.deps![i]));
```

This only iterates indices `0` to `deps.length - 1` (the new array's length). If the deps array shrinks (e.g., from `[a, b, c]` to `[a, b]`), the comparison never checks that the third element was removed. React's `useEffect` compares both length and elements — this implementation only compares elements up to the new array's length.

## Expected Behaviour

When deps shrinks from `[a, b, c]` to `[a, b]`, the effect should re-run because a dependency was removed, matching React's behaviour.

## Actual Behaviour

```typescript
// deps shrinks from [a, b, c] to [a, b]
// deps.some() only checks indices 0 and 1 — index 2 is never checked
const shouldRun = deps.some((d, i) => !Object.is(d, prev.deps![i]));
// shouldRun is false — effect does NOT re-run
```

The effect silently skips re-running when deps shrink, causing stale closures.

## Proposed Fix

```typescript
// packages/jsx/src/hooks.ts:266 (and line 290 for useLayoutEffect)
const shouldRun = !deps || !prev.deps ||
  deps.length !== prev.deps.length ||
  deps.some((d, i) => !Object.is(d, prev.deps![i]));
```

ionfwsrijan:
## Summary
Added length check to deps comparison in `useEffect` and `useLayoutEffect`.

## Problem
When deps shrinks (e.g., `[a,b,c]` → `[a,b]`), the `deps.some()` comparison only checked indices 0 and 1, missing the removed third element. This caused effects to not re-run when dependencies were removed, leading to stale closures.

## Fix
Added `deps.length !== prev.deps.length` check before element-wise comparison in both `useEffect` and `useLayoutEffect`.

## Changed
- `packages/jsx/src/hooks.ts`: Added length comparison to both `depsChanged()` checks

Fixes #2977

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed effect hooks so they re-run when the number of dependencies changes.
  * Improved dependency tracking for both standard and layout effects.

### DIFF
--- a/packages/jsx/src/hooks.ts
+++ b/packages/jsx/src/hooks.ts
@@ -263,7 +263,7 @@ export function useEffect(effect: () => void | (() => void), deps?: any[]): void
         fiber.effects.push(record);
     } else {
         const prev = fiber.hooks[idx];
-        const shouldRun = !deps || !prev.deps || deps.some((d, i) => !Object.is(d, prev.deps![i]));
+        const shouldRun = !deps || !prev.deps || deps.length !== prev.deps.length || deps.some((d, i) => !Object.is(d, prev.deps![i]));
 
         if (shouldRun) {
             prev.deps = deps;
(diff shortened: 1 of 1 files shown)


ionfwsrijan:

PR #2984, opened 2026-07-24
fix(tss): use safeResolve in auto-theme OSC 11 timeout catch block

### LINKED ISSUE #2979: [bug]  Auto-Theme OSC 11 Timeout Bypasses `safeResolve`
## Description

In `packages/tss/src/auto-theme.ts`, when `setRawMode` throws (line 134), the catch block calls `resolve(null)` directly instead of `safeResolve(null)`:

```typescript
// packages/tss/src/auto-theme.ts:134-138
} catch {
  clearTimeout(timer);
  resolve(null);  // Bypasses safeResolve — leaks listener
  return;
}
```

The `safeResolve` function (defined earlier in the same scope) performs three critical cleanup steps:
1. Clears the timeout
2. Removes the `data` listener from stdin
3. Restores raw mode
4. Sets a `resolved` flag to prevent double-resolution

By calling `resolve(null)` directly, the `data` listener remains attached to stdin (memory leak), and the `resolved` flag is never set — so if data arrives later, `safeResolve` will still try to resolve the already-settled promise.

## Expected Behaviour

The timeout catch block should use `safeResolve(null)` to properly clean up the stdin listener and prevent double-resolution.

## Actual Behaviour

The `data` listener leaks on stdin, and the promise can be double-resolved if terminal data arrives after the timeout.

## Proposed Fix

```typescript
// packages/tss/src/auto-theme.ts:134-138
} catch {
  clearTimeout(timer);
  safeResolve(null);
  return;
}
```

ionfwsrijan:
## Summary
Used `safeResolve` in the auto-theme OSC 11 timeout catch block.

## Problem
When `setRawMode` threw in the catch block, the code called `resolve(null)` directly instead of `safeResolve(null)`. This bypassed cleanup logic that removes the stdin data listener and sets the resolved flag, causing a memory leak and potential double-resolution.

## Fix
Changed `resolve(null)` to `safeResolve(null)` in the catch block.

## Changed
- `packages/tss/src/auto-theme.ts`: Changed `resolve(null)` to `safeResolve(null)` in catch block

Fixes #2979

### DIFF
--- a/packages/tss/src/auto-theme.ts
+++ b/packages/tss/src/auto-theme.ts
@@ -134,7 +134,7 @@ async function queryOSC11BackgroundColor(): Promise<TerminalBackground | null> {
     } catch {
       // setRawMode can throw in some environments — bail out
       clearTimeout(timer)
-      resolve(null)
+      safeResolve(null)
       return
     }
 


ionfwsrijan:

PR #2985, opened 2026-07-24
fix(data): add error handling to tail file watcher truncation path

### LINKED ISSUE #2980: [bug] Tail File Watcher Truncation Path Missing Error Handling
## Description

In `packages/data/src/tail.ts`, the file watcher callback has two branches: growth (file got bigger) and truncation (file got smaller). The growth branch (lines 104–127) correctly wraps `fs.openSync`/`fs.readSync` in try/catch/finally with fd cleanup:

```typescript
// packages/data/src/tail.ts:104-127 (growth branch — correct)
if (curr.size > fileSize) {
  let fd: number | undefined;
  try {
    fd = fs.openSync(filePath, 'r');
    // ... read and process
  } catch {
    // File may have been deleted/moved between stat and read
  } finally {
    if (fd !== undefined) fs.closeSync(fd);
  }
}
```

But the truncation branch (lines 128–136) calls `fs.readFileSync` bare, without any error handling:

```typescript
// packages/data/src/tail.ts:128-136 (truncation branch — UNPROTECTED)
} else if (curr.size < fileSize) {
  const content = fs.readFileSync(filePath, 'utf-8');  // Crashes if file deleted
  const allLines = content.split('\n');
  partialLine = allLines.pop() ?? '';
  stream.lines = allLines.filter(l => l.length > 0).slice(-maxLines);
  fileSize = curr.size;
}
```

If the file is deleted between the `fs.existsSync` check (line 84) and the `readFileSync` call, an unhandled `ENOENT` exception crashes the watcher.

## Expected Behaviour

The truncation path should have the same try/catch protection as the growth path, gracefully handling file deletion between stat and read.

## Actual Behaviour

File deletion during truncation causes an unhandled exception that crashes the file watcher.

## Proposed Fix

```typescript
// packages/data/src/tail.ts:128-136
} else if (curr.size < fileSize) {
  try {
    const content = fs.readFileSync(filePath, 'utf-8');
    const allLines = content.split('\n');
    partialLine = allLines.pop() ?? '';
    stream.lines = allLines.filter(l => l.length > 0).slice(-maxLines);
    fileSize = curr.size;
  } catch {
    // File may have been deleted between stat and read
  }
}
```

ionfwsrijan:
## Summary
Added error handling to the tail file watcher's truncation branch.

## Problem
The growth branch of the file watcher had proper `try/catch/finally` with fd cleanup, but the truncation branch called `readFileSync` without any error handling. If the file was deleted between the `existsSync` check and the `readFileSync` call, an unhandled ENOENT exception would crash the watcher.

## Fix
Wrapped the truncation branch's `readFileSync` and line parsing in a try/catch that silently handles file-not-found errors.

## Changed
- `packages/data/src/tail.ts`: Added try/catch around truncation branch
Fixes #2980

### DIFF
--- a/packages/data/src/tail.ts
+++ b/packages/data/src/tail.ts
@@ -128,11 +128,15 @@ export function tail(filePath: string, opts: TailOptions = {}): TailStream {
         } else if (curr.size < fileSize) {
             // File was truncated — re-read, again holding back an
             // unterminated trailing line as partialLine (see readInitial).
-            const content = fs.readFileSync(filePath, 'utf-8');
-            const allLines = content.split('\n');
-            partialLine = allLines.pop() ?? '';
-            stream.lines = allLines.filter(l => l.length > 0).slice(-maxLines);
-            fileSize = curr.size;
+            try {
+                const content = fs.readFileSync(filePath, 'utf-8');
+                const allLines = content.split('\n');
+                partialLine = allLines.pop() ?? '';
+                stream.lines = allLines.filter(l => l.length > 0).slice(-maxLines);
+                fileSize = curr.size;
+            } catch {
+                // File may have been deleted between stat and read
+            }
         }
     };
 


ionfwsrijan:

PR #2986, opened 2026-07-24
fix(adapters): make addDocuments atomic to prevent partial store corruption

### LINKED ISSUE #2981: [bug]  LocalVectorStore `addDocuments` Leaves Store in Inconsistent State on Partial Failure
## Description

In `packages/adapters/src/ai/vectorStore.ts`, the `addDocuments` method (line 106) iterates through documents, calling `ai.embed()` and pushing each result into `_documents` one at a time:

```typescript
// packages/adapters/src/ai/vectorStore.ts:106-116
async addDocuments(docs: Omit<DocumentChunk, 'embedding'>[], ai: AIAdapter): Promise<void> {
  if (!ai.embed) {
    throw new Error('The AI adapter does not support embeddings.');
  }
  for (const doc of docs) {
    const embedding = await ai.embed(doc.text);
    this._documents.push({
      ...doc,
      embedding,
    });
  }
}
```

If `ai.embed()` fails on document N (e.g., network timeout, rate limit), documents 0 through N-1 have already been pushed into `_documents` with embeddings, but documents N and later have not. The store is left in a corrupted partial state: some documents have embeddings and can be queried, others are missing entirely. There is no way to resume or rollback — the caller has no indication of which documents were added.

## Expected Behaviour

All documents should be added atomically — either all succeed and are added, or none are added. The store should not be left in a partially-updated state.

## Actual Behaviour

On failure at document N, documents 0..N-1 are in the store with embeddings, N+1..end are missing. The store is permanently corrupted with no recovery path.

## Proposed Fix

```typescript
// packages/adapters/src/ai/vectorStore.ts:106-116
async addDocuments(docs: Omit<DocumentChunk, 'embedding'>[], ai: AIAdapter): Promise<void> {
  if (!ai.embed) {
    throw new Error('The AI adapter does not support embeddings.');
  }
  // Embed all documents first — if any fail, nothing is added to the store
  const embedded = await Promise.all(
    docs.map(async (doc) => ({
      ...doc,
      embedding: await ai.embed!(doc.text),
    }))
  );
  // Only push to store after all embeddings succeed
  this._documents.push(...embedded);
}
```

ionfwsrijan:
## Summary
Made `addDocuments` atomic to prevent partial store corruption.

## Problem
If `ai.embed()` failed on document N, documents 0..N-1 were already pushed into the store while N+1..end were missing, leaving the store in a corrupted partial state.

## Fix
All documents are now embedded first via `Promise.all`, and only pushed to `_documents` after all embeddings succeed. If any embed fails, nothing is added to the store.

## Changed
- `packages/adapters/src/ai/vectorStore.ts`: Changed `for` loop to `Promise.all` with deferred push

Fixes #2981

### DIFF
--- a/packages/adapters/src/ai/vectorStore.ts
+++ b/packages/adapters/src/ai/vectorStore.ts
@@ -107,13 +107,15 @@ export class LocalVectorStore {
         if (!ai.embed) {
             throw new Error('The AI adapter does not support embeddings.');
         }
-        for (const doc of docs) {
-            const embedding = await ai.embed(doc.text);
-            this._documents.push({
+        // Embed all documents first — if any fail, nothing is added to the store
+        const embedded = await Promise.all(
+            docs.map(async (doc) => ({
                 ...doc,
-                embedding,
-            });
-        }
+                embedding: await ai.embed!(doc.text),
+            }))
+        );
+        // Only push to store after all embeddings succeed
+        this._documents.push(...embedded);
     }
 
     async query(queryText: string, ai: AIAdapter, limit = 3): Promise<DocumentChunk[]> {


ionfwsrijan:

PR #3033, opened 2026-07-25
fix(adapters): prevent infinite loop in chunkText when size equals overlap

### LINKED ISSUE #3028: fix : infinite loop in chunkText function when size equals overlap

The `chunkText` function in `packages/adapters/src/ai/vectorStore.ts` enters an infinite loop when the `overlap` parameter is equal to or greater than `size`.

In `packages/adapters/src/ai/vectorStore.ts`, fix the break condition in the `chunkText` function. The current condition:

```typescript
if (start >= text.length - overlap && end === text.length) {
    break;
}
```

should be changed to:

```typescript
if (start >= text.length) {
    break;
}
```

The current condition does not handle the case where `size <= overlap`, causing `start` to not advance enough between iterations, leading to an infinite loop.

When `chunkText` is called with overlapping chunks (e.g., `chunkText(text, 500, 500)` or `chunkText(text, 500, 600)`), the function hangs indefinitely instead of returning the chunked text. This affects the `indexDirectory` function which relies on `chunkText`.

tmdeveloper007:

Fixed an infinite loop bug in the `chunkText` function in `packages/adapters/src/ai/vectorStore.ts`.

The previous break condition `start >= text.length - overlap && end === text.length` did not handle the case where `size <= overlap`, causing `start` to not advance enough between iterations.

## Changes Made

Modified `packages/adapters/src/ai/vectorStore.ts`:
```typescript
// Before:
if (start >= text.length - overlap && end === text.length) {
    break;
}
// After:
if (start >= text.length) {
    break;
}
```

## Impact it Made

When `chunkText` is called with `overlap >= size`, the function previously entered an infinite loop. This fix ensures the loop terminates correctly.

Closes #3028

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text chunking termination to prevent unnecessary final iterations when processing content.

### DIFF
--- a/packages/adapters/src/ai/vectorStore.ts
+++ b/packages/adapters/src/ai/vectorStore.ts
@@ -47,7 +47,7 @@ export function chunkText(text: string, size = 500, overlap = 50): string[] {
         const end = Math.min(start + size, text.length);
         chunks.push(text.slice(start, end));
         start += (size - overlap);
-        if (start >= text.length - overlap && end === text.length) {
+        if (start >= text.length) {
             break;
         }
     }


Karanjot786:
Blocking — the infinite loop is NOT fixed. vectorStore.ts:50 changes the break to `start >= text.length`, but when size<=overlap, `start += (size - overlap)` never advances, so it still loops forever (verified size=50/overlap=50). Guard against size<=overlap (advance by at least 1) and add a test.

Karanjot786:

PR #3038, opened 2026-07-25
Feat/store history undo redo

### LINKED ISSUE #2772: [feature] Add undo/redo history middleware to @termuijs/store
What problem does this solve?

@termuijs/store already has batch(), computed(), persist, shallow, and a Middleware<T>[] extension point, but there's no built-in way to snapshot state and step backward/forward through changes. Any app that needs undo — form editors, config tools, layout/dashboard builders made with @termuijs/quick — has to hand-roll history tracking on top of the store today.

Proposed solution

Add a history option to createStore, with undo()/redo() exposed on the returned store:

typescript
const useStore = createStore(
  (set) => ({
    count: 0,
    increment: () => set((s) => ({ count: s.count + 1 })),
  }),
  { history: { limit: 50, coalesceMs: 300 } }
);

useStore.undo();
useStore.redo();
useStore.getHistory(); // { past: T[], future: T[] }
useStore.resetHistory();
A batch() call should produce exactly one history entry, not one per setState inside it.
coalesceMs merges rapid consecutive updates (e.g. typing) into a single undo step.
History stacks are capped at limit; oldest entries drop off.
History itself is not persisted by persist unless explicitly opted into, to avoid growing the on-disk file unbounded.
Alternatives considered

Maintaining a separate array of snapshots outside the store and manually calling setState to "undo" — works but duplicates the store's own change-detection/equality logic and breaks when combined with batch().

Which package would this belong in?

@termuijs/store

GSSoC contributor?
 Yes. I contribute under GSSoC 2026 and want to work on this.

palak170306-design:
## Description
Adds a `history` option to `createStore()` with native `undo()`/`redo()`/`getHistory()`/`resetHistory()` on the returned store. A `batch()` call collapses to exactly one history entry, `coalesceMs` merges rapid consecutive updates via a timestamp-based sliding window, and `undo()`/`redo()` now throw a clear error instead of silently no-op'ing when `history` isn't configured on that store.

## Related Issue
Closes #2772 

## Which package(s)?
`@termuijs/store`

## Type of Change
- [ ] 🐛 Bug fix (`type:bug`)
- [ ] 📝 Docs (`type:docs`)

## Checklist

## GSSoC 2026 Participation
- [X] You are a GSSoC 2026 contributor.
- Your GSSoC profile: `https://gssoc.girlscript.org/profile/048e6a97-baaf-4a03-b6e5-bdfc6c38e173`

N/A — non-visual, `@termuijs/store` API change only.

## Notes for the Reviewer
- This ships **alongside** the existing `createSnapshotHistory()` (`snapshot-history.ts`), not as a replacement. That's an explicit-checkpoint model (`capture()`/`restore(id)`); this PR's `history` option is automatic, tracking every `setState`/`batch()`. Both can technically be used on the same store but would run as two independent, uncoordinated undo stacks — documented in a code comment, worth a maintainer opinion on whether that should be actively discouraged/guarded against.
- `undo()`/`redo()` bypass `setState` and any configured middleware, restoring the captured state object directly — intentional (avoids re-running transforms on a historical snapshot), but means middleware-enforced invariants aren't re-validated on undo/redo. Flagged in a comment; open to feedback if this tradeoff should be reconsidered.
- `coalesceMs` is timestamp-based (compares `Date.now()` gaps between calls), not `setTimeout`-based — there's no deferred flush, so tests drive it via `vi.setSystemTime()` between calls rather than `vi.advanceTimersByTime()` after them.
- `undo()`/`redo()` now throw when `history` isn't set on that store, rather than silently doing nothing — this is a small API decision worth a second look, since it's a behavior change from the initial draft (which no-op'd).

## Summary by CodeRabbit

- **New Features**
  - Added optional undo and redo history for store state.
  - Added history inspection and reset controls.
  - Added configurable history size and update coalescing.
  - Enhanced snapshot history to safely handle complex state and preserve reliable navigation.

- **Bug Fixes**
  - Failed snapshot restores no longer move the history position.
  - New snapshots after undo correctly discard the redo branch.
  - Clearing snapshot history preserves the current state.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -40,6 +40,7 @@ We want to thank all the amazing contributors who have helped make TermUI what i
 | <img src="https://avatars.githubusercontent.com/u/174622309?v=4" width="40" height="40" style="border-radius: 50%;" alt="Harshit-Maurya838" /> | [@Harshit-Maurya838](https://github.com/Harshit-Maurya838) | 8 |
 | <img src="https://avatars.githubusercontent.com/u/245353458?v=4" width="40" height="40" style="border-radius: 50%;" alt="namrarafique93-del" /> | [@namrarafique93-del](https://github.com/namrarafique93-del) | 8 |
 | <img src="https://avatars.githubusercontent.com/u/162754915?v=4" width="40" height="40" style="border-radius: 50%;" alt="ARPANPATRA111" /> | [@ARPANPATRA111](https://github.com/ARPANPATRA111) | 7 |
+| <img src="https://avatars.githubusercontent.com/u/227770746?v=4" width="40" height="40" style="border-radius: 50%;" alt="palak170306-design" /> | [@palak170306-design](https://github.com/palak170306-design) | 7 |
 | <img src="https://avatars.githubusercontent.com/u/212236853?v=4" width="40" height="40" style="border-radius: 50%;" alt="nandani-singh15" /> | [@nandani-singh15](https://github.com/nandani-singh15) | 7 |
 | <img src="https://avatars.githubusercontent.com/u/219233938?v=4" width="40" height="40" style="border-radius: 50%;" alt="PremSahith" /> | [@PremSahith](https://github.com/PremSahith) | 7 |
 | <img src="https://avatars.githubusercontent.com/u/180026264?v=4" width="40" height="40" style="border-radius: 50%;" alt="Krishnavamsi-codes" /> | [@Krishnavamsi-codes](https://github.com/Krishnavamsi-codes) | 7 |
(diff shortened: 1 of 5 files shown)


Karanjot786:
Blocking — duplicate + does not build. Undo/redo already exists on main via createSnapshotHistory() (snapshot-history.ts #2919, on top of #2005). This bolts a second overlapping API onto createStore(); build-and-test fails and it is CONFLICTING; error messages reference a nonexistent CreateStore(). Extend snapshot-history.ts.

Karanjot786:

Issues:

1. The PR's own test fails in CI
   File: packages/store/src/history-option.test.ts, line 30 (and packages/store/src/store.ts flushBatch, lines ~104-127)
   Problem: `batch() collapses to exactly one history entry` asserts `past.length === 1` synchronously, but `flushBatch` notifies listeners via `queueMicrotask`, so the history subscriber hasn't run yet and `past.length` is 0. This is not just a test bug: real code calling `undo()` right after `batch()` will see stale history for the same reason.
   Fix: record history synchronously inside the flush (not via a subscriber), or if the async design is intended, `await` a microtask in the test and document that `undo()` after `batch()` needs a tick.

2. Two competing undo mechanisms in the same package
   File: packages/store/src/store.ts (`history` option) vs packages/store/src/snapshot-history.ts (`createSnapshotHistory`)
   Problem: both exist now with different APIs and different cloning semantics.
   Fix: build the `history` option on top of `createSnapshotHistory` (or drop one). One implementation only.

3. Typos in error messages
   File: packages/store/src/store.ts, lines 661 and 673
   Problem: `CreativeStore()` / `CreateStore()` should be `createStore()`.

4. `getHistory()` leaks live state references
   File: packages/store/src/store.ts, lines ~683-689
   Problem: returns the internal `past`/`future` arrays' objects directly; a caller mutating one corrupts the undo stack. `snapshot-history.ts` clones defensively; do the same.

5. Branch conflicts with main
   Fix: merge current `main` into your branch and resolve.

- [ ] Fixes 1-4 applied
- [ ] Merged main, no conflicts
- [ ] Tests pass locally (run: bun vitest run packages/store)
- [ ] PR title follows format: feat(store): add undo/redo history option

PR #3088, opened 2026-07-26
fix(core): added _translateY offset application in writeString for scrolled containers

### LINKED ISSUE #3082: fix(core): writeString drops translated text in scrolled containers
Fix the `writeString` method in `packages/core/src/terminal/Screen.ts` to correctly apply the `_translateY` offset when checking row bounds in scrolled containers.

In the `writeString` method in `packages/core/src/terminal/Screen.ts`:
- Apply `_translateY` to the `row` parameter before the bounds check (line ~377), matching how `setCell` does it
- The current code checks `if (!(row >= 0 && row < this._rows))` without applying `_translateY`
- The fix should add `row += this._translateY` before the bounds check, just like `setCell` does at line ~339

Current problematic code:
```typescript
row = Math.floor(row);
col = Math.floor(col);
if (!(row >= 0 && row < this._rows)) return;
```

Should become:
```typescript
row = Math.floor(row);
col = Math.floor(col);
row += this._translateY;
if (!(row >= 0 && row < this._rows)) return;
```

- Text written via `writeString` in scrolled containers (ScrollView) will correctly account for the scroll offset
- Without this fix, text content in scrolled containers renders at incorrect positions while widget cell content is correct

## Note

tmdeveloper007:
Fixed the `writeString` method in `packages/core/src/terminal/Screen.ts` to correctly apply the `_translateY` offset when checking row bounds in scrolled containers.

## Changes Made
- Added `row += this._translateY` before the bounds check in `writeString`, mirroring the `setCell` behavior
- Added a corresponding test case verifying that `writeString` respects `pushTranslateY`/`popTranslateY`

## Impact it Made
- Text content in scrolled containers now renders at correct positions
- Previously, `writeString` wrote at absolute row positions while `setCell` used translated positions, causing misalignment

Closes #3082

## Note

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed text rendering in vertically translated or scrolled containers.
  * Content now appears at the correct translated row and is properly clipped when positioned outside the visible screen area.

* **Tests**
  * Added coverage for translated text rendering and off-screen clipping.

### DIFF
--- a/packages/core/src/terminal/Screen.test.ts
+++ b/packages/core/src/terminal/Screen.test.ts
@@ -114,6 +114,27 @@ describe('Screen', () => {
         expect(screen.back[0][0].char).toBe(' ');
     });
 
+    it('writeString applies _translateY offset in scrolled containers', () => {
+        const screen = new Screen(20, 10);
+        // Simulate a scrolled container: push a translateY offset
+        // This is how ScrollView offsets content rendering
+        screen.pushTranslateY(-3);
+
+        // Row 2 + (-3) = -1, which is out of bounds, so clipped
+        screen.writeString(0, 2, 'Hello');
+        expect(screen.back[0][0].char).toBe(' ');
+        expect(screen.back[1][0].char).toBe(' ');
+
+        // For positive translateY, content should appear shifted down
+        screen.popTranslateY();
+        screen.pushTranslateY(2);
+        screen.writeString(0, 0, 'Hi');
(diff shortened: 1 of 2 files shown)


Karanjot786:
Closing this one. This fix is already on `main` from a previously merged PR (verified against the current source), so the change is a no-op / conflicts. Thanks anyway — the bug is fixed.

PR #3089, opened 2026-07-26
fix(cli): added Windows backslash normalization in registry path matching

### LINKED ISSUE #3086: fix(cli): normalize Windows backslashes in registry path matching
Fix `normalizeComponentPath` in `packages/cli/src/commands/add.ts` to normalize Windows backslashes to forward slashes before stripping the registry path prefix.

In `packages/cli/src/commands/add.ts`, the `normalizeComponentPath` function currently strips the prefix `registry/components/<slug>/` from file paths. However, if the registry returns paths with Windows-style backslashes ( the prefix check fails because the path separator does not match.

The fix should replace backslashes with forward slashes before processing the prefix:
```typescript
function normalizeComponentPath(slug: string, filePath: string): string {
    filePath = filePath.replace(/\\/g, '/');
    const prefix = `registry/components/${slug}/`;
    if (filePath.startsWith(prefix)) {
        return filePath.slice(prefix.length);
    }
    return filePath;
}
```

- CLI `add` command works correctly on Windows where registry file paths may use backslash separators
- Prevents path resolution failures when registry data contains Windows-style paths

## Note

tmdeveloper007:
Fixed `normalizeComponentPath` in `packages/cli/src/commands/add.ts` to normalize Windows backslashes to forward slashes before stripping the registry path prefix.

## Changes Made
- Added `filePath = filePath.replace(/\\/g, '/')` at the start of `normalizeComponentPath`
- Added a test case verifying Windows-style paths are handled correctly

## Impact it Made
- CLI `add` command works correctly on Windows where registry file paths may use backslash separators
- Prevents path resolution failures when registry data contains Windows-style paths

Closes #3086

## Note

### DIFF
--- a/packages/cli/src/commands/add.test.ts
+++ b/packages/cli/src/commands/add.test.ts
@@ -52,4 +52,20 @@ describe('writeComponentFiles', () => {
             [{ path: '../../evil.ts', content: 'x' }], { dryRun: false }))
             .toThrow(/outside/);
     });
+
+    it('normalizes Windows backslashes in registry file paths', () => {
+        const root = mkdtempSync(join(tmpdir(), 'tcli-'));
+        // Simulate Windows-style paths from registry
+        const written = writeComponentFiles(root, 'spinner', [
+            { path: 'registry\\components\\spinner\\index.ts', content: 'x' },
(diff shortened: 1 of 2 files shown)


Karanjot786:

PR #3091, opened 2026-07-26
fix(motion): added empty/null guard in stagger function

### LINKED ISSUE #3085: fix(motion): handle empty items array in stagger function
Fix the `stagger` function in `packages/motion/src/stagger.ts` to handle empty items arrays gracefully without throwing or returning invalid results.

In `packages/motion/src/stagger.ts`, add a guard clause at the start of the function to handle the empty array case:
```typescript
if (items.length === 0) {
    return { delays: [], duration: 0 };
}
```

- Prevents runtime errors or NaN/Infinity values when `stagger` is called with an empty array
- Makes the animation API more robust for dynamic item lists

## Note

tmdeveloper007:
Fixed the `stagger` function in `packages/motion/src/stagger.ts` to handle empty and null/undefined `animations` arrays gracefully.

## Changes Made
- Added a guard clause: `if (!animations || animations.length === 0) { onComplete?.(); return () => {}; }`
- Added a test verifying undefined/null inputs do not throw and call onComplete

## Impact it Made
- Prevents TypeErrors when `stagger` is called with undefined/null animations
- Makes the animation API more robust for dynamic item lists

Closes #3085

## Note

### DIFF
--- a/packages/motion/src/stagger.test.ts
+++ b/packages/motion/src/stagger.test.ts
@@ -44,6 +44,15 @@ describe('stagger()', () => {
         expect(onComplete).toHaveBeenCalledTimes(1);
     });
 
+    it('handles undefined and null animations without throwing', () => {
+        const onComplete = vi.fn();
+        expect(() => stagger(undefined as any, 100, onComplete)).not.toThrow();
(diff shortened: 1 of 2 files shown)


Karanjot786:

PR #3160, opened 2026-07-26
fix(jsx): exclude children from memo() shallow comparison

### LINKED ISSUE #3155: [bug] `memo()` Never Skips Re-Renders for Components with Children
## Description

In `packages/jsx/src/memo.ts`, the `memo()` function uses shallow comparison (line 79) to decide whether to skip a re-render. The reconciler passes `children` merged into props at `reconciler.ts:467`:

```typescript
vnode = component({ ...props, children: children.length === 1 ? children[0] : children });
```

The `memo()` cache stores the full props object including `children` (line 83: `cache.set(fiber, { prevProps: { ...props } as P, ... })`). On the next render, `shallowEqual` compares `children` via `Object.is` (line 38). Since children are new VNode references every render, `Object.is` always returns `false` for the `children` key.

**Result:** `memo()` never skips re-rendering for any component that receives children — which is the vast majority of JSX components.

## Expected Behaviour

`memo()` should exclude `children` from the shallow comparison, since children are reconciled separately by the reconciler and their reference changes are expected.

## Actual Behaviour

```typescript
// memo.ts:79 — children always differ, so compare always returns false
if (entry && compare(entry.prevProps, props as P)) {
    return entry.prevResult; // Never reached for components with children
}
```

## Proposed Fix

```typescript
// memo.ts:77-84 — Exclude children from comparison
if (fiber) {
    const entry = cache.get(fiber);
    const { children: _prevChildren, ...prevRest } = (entry?.prevProps ?? {}) as any;
    const { children: _nextChildren, ...nextRest } = (props ?? {}) as any;
    if (entry && compare(prevRest, nextRest)) {
        return entry.prevResult;
    }
    const result = component(props);
    cache.set(fiber, { prevProps: { ...props } as P, prevResult: result });
    return result;
}
```

ionfwsrijan:
## Summary
Fixes #3155 — `memo()` was passing `children` into the shallow comparison, causing unnecessary re-renders whenever children changed even if the component's own props hadn't.

## Changes
- `packages/memo.ts`: Filter out `children` from the props object before passing it to the shallow-comparison cache key, so that changes to `children` no longer invalidate the memo.

## Test plan
- [ ] Verified that `memo()`-wrapped components no longer re-render when only children change.
- [ ] Verified that the components still re-render when their own props change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memoized component rendering by ignoring changes to the `children` reference when other properties remain unchanged.
  * Prevented unnecessary re-renders caused solely by changes in the `children` reference.

### DIFF
--- a/packages/jsx/src/memo.ts
+++ b/packages/jsx/src/memo.ts
@@ -76,7 +76,12 @@ export function memo<P extends Record<string, any>>(
 
         if (fiber) {
             const entry = cache.get(fiber);
-            if (entry && compare(entry.prevProps, props as P)) {
+            // Exclude children from comparison — children are new VNode
+            // references every render, so including them means memo() never
+            // skips a re-render for components that receive children.
+            const { children: _prevChildren, ...prevRest } = (entry?.prevProps ?? {}) as any;
+            const { children: _nextChildren, ...nextRest } = (props ?? {}) as any;
+            if (entry && compare(prevRest, nextRest)) {
                 return entry.prevResult;
             }
             const result = component(props);


ionfwsrijan:

Karanjot786:
Blocking — over-corrects. Excluding children entirely from memo() comparison (memo.ts:78-84) means when a memoized component props are unchanged but children content changes, it returns the stale cached VNode forever (verified A→B still renders A). Compare children structurally/by key rather than dropping them.

PR #3161, opened 2026-07-26
fix(motion): remove ghost subscribers from savedSubs on unsubscribe

### LINKED ISSUE #3156: [bug]  Timer Pool Ghost Subscribers from Saved Subscriptions
## Description

In `packages/motion/src/timer-pool.ts`, when a VirtualClock is injected (lines 60-70), all current pool callbacks are saved to `savedSubs`:

```typescript
// timer-pool.ts:61-68
    clearInterval(entry.id);
    if (!savedSubs.has(delay)) {
        savedSubs.set(delay, new Set());
    }
    for (const cb of entry.subs) {
        savedSubs.get(delay)!.add(cb);
    }
}
pool.clear();
```

When a subscriber later unsubscribes during the clock period (via the normal unsubscribe at lines 134-142), the pool is already cleared, so `pool.get(delayMs)` returns `undefined` and the unsubscribe is a no-op. **`savedSubs` is never updated** when a subscriber unsubscribes during the clock period.

On clock detach (lines 77-88), **all** `savedSubs` callbacks are restored to the pool, including ones whose subscribers had already unsubscribed. This creates ghost subscribers that fire indefinitely.

## Expected Behaviour

When a subscriber unsubscribes during the clock period, the callback should also be removed from `savedSubs` so it is not restored on clock detach.

## Actual Behaviour

Unsubscribing during the clock period is silently ignored. On detach, the unsubscribed callback is restored and fires forever.

## Proposed Fix

```typescript
// timer-pool.ts:115-119 — When unsubscribing during clock period, also remove from savedSubs
return () => {
    unsub();
    const idx = clockSubs.indexOf(entry);
    if (idx !== -1) clockSubs.splice(idx, 1);
    // Also remove from savedSubs so it's not restored on detach
    const saved = savedSubs.get(delayMs);
    if (saved) saved.delete(cb);
};
```

ionfwsrijan:
## Summary
Fixes #3156 — When a timer was unsubscribed during an active clock period, it was removed from the active subscription but remained in `savedSubs`, creating a ghost entry that persisted until the next clock period.

## Changes
- `packages/timer-pool.ts`: Remove the entry from `savedSubs` when `unsubscribe()` is called during an active period.

## Test plan
- [ ] Verified that unsubscribing mid-period no longer leaves entries in `savedSubs`.
- [ ] Verified that the timer pool correctly cleans up all resources after unsubscription.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed timer callbacks being restored after they were unsubscribed when switching from a virtual clock back to the regular timer system.

### DIFF
--- a/packages/motion/src/timer-pool.ts
+++ b/packages/motion/src/timer-pool.ts
@@ -116,6 +116,10 @@ export function subscribe(...args: unknown[]): () => void {
             unsub();
             const idx = clockSubs.indexOf(entry);
             if (idx !== -1) clockSubs.splice(idx, 1);
+            // Also remove from savedSubs so ghost subscribers are not
+            // restored when the clock detaches.
+            const saved = savedSubs.get(delayMs);
+            if (saved) saved.delete(cb);
         };
     }
 


ionfwsrijan:

Karanjot786:
Blocking — patches the wrong branch. It prunes savedSubs in the active-clock subscribe path (timer-pool.ts:108-121), which never populates savedSubs. The real leak is the real-setInterval unsub closure (122-143): during an active clock pool.get(delayMs) is undefined so it no-ops and the callback stays, resurrected on detach. Fix that closure.

PR #3163, opened 2026-07-26
fix(store): rollback all stores on flushBatch listener failure

### LINKED ISSUE #3158: [bug] Store flushBatch Partial Rollback Leaves Cross-Store State Inconsistent
## Description

In `packages/store/src/store.ts`, the `flushBatch` function (lines 115-137) commits all stores first, then notifies listeners in sequence. If store B's listener throws, only store B is rolled back:

```typescript
// store.ts:121-136
for (const [listeners, { commit }] of stores) {
    newStates.set(listeners, commit());  // All stores committed
}
    const newState = newStates.get(listeners);
    try {
        }
    } catch (e) {
        const failedEntry = stores.find(([l]) => l === listeners);
        if (failedEntry) {
            failedEntry[1].rollback();  // Only failing store rolled back
        }
        throw e;
    }
}
```

Store A (which was already committed and notified successfully) remains committed. The system ends up with an inconsistent cross-store state — store A has the new values but store B has been rolled back to its previous state.

## Expected Behaviour

Either all stores should be committed atomically after all listeners succeed, or all stores should be rolled back on any listener failure.

## Actual Behaviour

Partial rollback leaves stores in inconsistent states when batch spans multiple stores.

## Proposed Fix

```typescript
// store.ts — Commit all stores, notify all, rollback all on failure
_batchStores.clear();
const newStates = new Map();
for (const [listeners, { commit }] of stores) {
    newStates.set(listeners, commit());
}
try {
        const newState = newStates.get(listeners);
        }
    }
} catch (e) {
    // Rollback ALL stores, not just the failing one
    for (const [, { rollback }] of stores) {
        rollback();
    }
    throw e;
}
```

ionfwsrijan:
## Summary
Fixes #3158 — In `flushBatch()`, if a listener threw an error, subsequent stores had already been updated in-place, leaving the system in a partially-flushed state.

## Changes
- `packages/store.ts`: Added rollback logic — before notifying listeners, snapshot the old state of each store. If any listener throws, roll back all stores to their pre-flush state before re-throwing.

## Test plan
- [ ] When one listener throws, all stores are rolled back to their previous state.
- [ ] Normal flush (no errors) works as before with no performance regression.

### DIFF
--- a/packages/store/src/store.ts
+++ b/packages/store/src/store.ts
@@ -121,19 +121,20 @@ function flushBatch(threw: boolean, immediate = false) {
             for (const [listeners, { commit }] of stores) {
                 newStates.set(listeners, commit());
             }
-            for (const [listeners, { prevState }] of stores) {
-                const newState = newStates.get(listeners);
-                try {
+            try {
+                for (const [listeners, { prevState }] of stores) {
+                    const newState = newStates.get(listeners);
                     for (const listener of listeners) {
                         listener(newState, prevState);
                     }
-                } catch (e) {
-                    const failedEntry = stores.find(([l]) => l === listeners);
-                    if (failedEntry) {
-                        failedEntry[1].rollback();
-                    }
-                    throw e;
                 }
+            } catch (e) {
+                // Roll back ALL stores if any listener throws, to keep
+                // cross-store state consistent.
+                for (const [, { rollback }] of stores) {
+                    rollback();
+                }
+                throw e;
             }
         };
         if (immediate) {


ionfwsrijan:

Karanjot786:
Blocking — reverts the intentional fix #2887 (issue #2865). store.ts:104-136 restores rollback-ALL-stores on any listener throw, which #2887 deliberately narrowed to the failing store only (rolling back all destroyed unrelated committed stores). Keep the failing-store-only scope.

PR #3164, opened 2026-07-26
fix(quick): set reactive metadata on Text widgets from toWidget()

### LINKED ISSUE #3159: [bug] `toWidget()` Never Sets Reactive Metadata for Function Children
## Description

In `packages/quick/src/layout.ts`, the `toWidget()` function (line 19-27) handles reactive functions by creating a Text widget with the resolved value:

```typescript
// layout.ts:24-26
// Reactive function — create text, will be updated by refresh
const text = new Text(resolve(child as Reactive<string>), { height: 1, ...style });
return text;
```

The comment says "will be updated by refresh", but `toWidget()` **never stores the reactive reference** on the widget. The `app.ts` auto-refresh (`_refreshReactiveWidgets`) looks for `__reactiveContent` on Text widgets (line 440), but `toWidget()` never sets it.

In contrast, `widgets.ts:67` correctly sets `(text as any).__reactiveContent = child` for the `text()` builder.

**Result:** Reactive strings passed through `row()`, `col()`, or `stack()` are never refreshed — they show the initial value and never update.

## Expected Behaviour

When a reactive function is passed to `toWidget()`, the reactive reference should be stored on the widget so the auto-refresh can update it.

## Actual Behaviour

Reactive strings in layout containers are static — the auto-refresh never finds the `__reactiveContent` metadata.

## Proposed Fix

```typescript
// layout.ts:24-26
if (typeof child === 'function') {
    const text = new Text(resolve(child as Reactive<string>), { height: 1, ...style });
    (text as any).__reactiveContent = child;
    return text;
}
```

ionfwsrijan:
## Summary
Fixes #3159 — `Text` widgets were not getting the `__reactiveContent` flag set, so the Quick framework's reactive diffing treated them as static and skipped diffing their content.

## Changes
- `packages/quick.ts`: Set `__reactiveContent = true` on `Text` widget instances in `toWidget()`.

## Test plan
- [ ] Text widgets now correctly re-render when their content changes reactively.

### DIFF
--- a/packages/quick/src/layout.ts
+++ b/packages/quick/src/layout.ts
@@ -23,6 +23,7 @@ export function toWidget(child: LayoutChild, style: Partial<Style> = {}): Widget
     }
     // Reactive function — create text, will be updated by refresh
     const text = new Text(resolve(child as Reactive<string>), { height: 1, ...style });
+    (text as any).__reactiveContent = child;
     return text;
 }
 


ionfwsrijan:

PR #3196, opened 2026-07-27
fix : applied clamping when inMin equals inMax in mapRange

### LINKED ISSUE #3191: fix : mapRange should apply clamping when inMin equals inMax

In `packages/motion/src/interpolate.ts`, the `mapRange()` function has an early-return guard for the degenerate case where `inMin === inMax`:

```ts
if (inMin === inMax) {
    return outMin;
}
```

When `clamp: true` (the default), this early return ignores the clamping option entirely. If `clamp: false`, it should also return `outMin`. The current code bypasses the clamping logic entirely.

- Move the `inMin === inMax` check to AFTER the clamp logic, or wrap it with the same clamp guard.
- When `inMin === inMax` and `clamp === true`, return the clamped value (which for a degenerate range is `outMin` regardless, but the clamping step should still execute for correctness).
- When `inMin === inMax` and `clamp === false`, return `outMin` as-is.

## Impact that it Would Provide

Ensures the clamping option is respected consistently, even for degenerate input ranges. This prevents subtle bugs where callers relying on `clamp: false` might unexpectedly get `outMin` instead of the input value mapped through an empty range.

**Note: Please assign this issue to the `tmdeveloper007` account.

tmdeveloper007:

Fixed the `mapRange()` function in `packages/motion/src/interpolate.ts` to respect the `clamp` option in the degenerate case where `inMin === inMax`.

Previously, the early return for `inMin === inMax` bypassed the clamping logic entirely. The fix applies clamping before returning when `clamp: true` and `outMin !== outMax` — the result is bounded to `Math.min(outMin, outMax)`, consistent with how the normal mapping path handles clamping.

## Changes Made

**packages/motion/src/interpolate.ts**
- Changed the `inMin === inMax` branch to check the `clamp` option
- When `clamp: true` and `outMin !== outMax`: return `Math.min(outMin, outMax)` (clamped to output range)
- When `clamp: false` or `outMin === outMax`: return `outMin` (original behavior)

## Impact it Made

Ensures the clamping option is respected consistently even for degenerate input ranges. Users who explicitly set `clamp: true` can now rely on consistent behavior regardless of input range configuration.

Closes #3191

**Note: Please assign this PR to the `tmdeveloper007` account.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved range mapping for zero-length input ranges when clamping is enabled.
  - Ensures results use the lower output bound consistently, including reversed output ranges.

### DIFF
--- a/packages/motion/src/interpolate.ts
+++ b/packages/motion/src/interpolate.ts
@@ -21,6 +21,9 @@ export function mapRange(
     const clamp = options?.clamp ?? true;
 
     if (inMin === inMax) {
+        if (clamp && outMin !== outMax) {
+            return Math.min(outMin, outMax);
+        }
         return outMin;
     }
 


Karanjot786:

PR #3234, opened 2026-07-27
fix : trailing call fires unconditionally in throttle utility

### LINKED ISSUE #3229: fix : trailing call fires unconditionally in throttle utility

Fix a logic error in the `throttle` utility in `packages/core/src/utils/throttle.ts`. The trailing edge call condition `lastArgs !== args` always evaluates to `true` because JavaScript array comparison uses reference equality, not value equality. This causes the trailing call to fire unconditionally, ignoring whether additional calls were made during the wait interval.

In `packages/core/src/utils/throttle.ts`, add a `hasNewCalls` flag to track whether new calls arrived during the wait interval. Replace the unconditional `if (lastArgs !== args || !leading)` with a proper check.

Ensures the trailing call in the throttle utility fires only when additional calls were made during the interval, matching the documented behavior. This fixes incorrect behavior in any code relying on throttle's trailing edge semantics.

tmdeveloper007:

Fixed a logic error in the `throttle` utility where the trailing edge call fired unconditionally because JavaScript array comparison uses reference equality. Added a `hasNewCalls` boolean flag to properly track whether additional calls arrived during the wait interval.

## Changes Made

**packages/core/src/utils/throttle.ts**
- Added `hasNewCalls` boolean flag
- Set `hasNewCalls = true` when a new call arrives while timer is active
- In the timer callback, use `hasNewCalls || !leading` instead of `lastArgs !== args || !leading`
- Reset `hasNewCalls = false` after timer fires

## Impact it Made

Trailing edge calls now fire only when additional calls were made during the wait interval, matching the documented throttle behavior.

Closes #3229

### DIFF
--- a/packages/core/src/utils/throttle.ts
+++ b/packages/core/src/utils/throttle.ts
@@ -27,6 +27,7 @@ export function throttle<T extends (...args: unknown[]) => void>(
     const leading = options?.leading ?? true;
     let timer: ReturnType<typeof setTimeout> | undefined;
     let lastArgs: Parameters<T> | undefined;
+    let hasNewCalls = false;
 
     const throttled = function (...args: Parameters<T>) {
         lastArgs = args;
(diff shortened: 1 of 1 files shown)


Karanjot786:

PR #3256, opened 2026-07-29
test(cli): add equals-format test for arg parser

tmdeveloper007:
## What
Adds a test for `--flag=value` (equals-style) argument parsing, alongside the existing `--flag value` tests. Ensures both formats produce identical parsed results.

## Why
The CLI arg parser supports both formats but only `--flag value` was tested.

## Testing
`bun vitest run packages/cli/src/args.equals.test.ts`

---

## Summary by CodeRabbit

* **New Features**
  * Command-line options support the convenient `--key=value` format, including string, boolean, numeric, and empty values.
  * Equals-based and space-separated options can be used together.

* **Tests**
  * Added coverage for mixed argument formats and multiple value types to help ensure consistent command-line parsing.

### DIFF
--- a/orchestrate.py
+++ b/orchestrate.py
@@ -1,0 +1,774 @@
+#!/usr/bin/env python3
+"""TermUI Cron Orchestrator — fork-only PR workflow."""
+
+import os
+import sys
+import json
+import subprocess
+import urllib.request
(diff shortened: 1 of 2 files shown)


Karanjot786:

PR #3258, opened 2026-07-29
docs(tss): add inline comment explaining safeResolve in osc-11 timeout

tmdeveloper007:
## What
Adds an inline comment block in `packages/tss/src/theme/derive.ts` explaining the `safeResolve` wrapper around the OSC 11 timeout catch block.

## Why
A previous fix (#1984) introduced `safeResolve` in the timeout callback but the rationale was not documented.

## Testing
No functional change -- `bun run typecheck && bun vitest run packages/tss/`

---


Karanjot786:

PR #3259, opened 2026-07-29
test(router): add useParams tests for optional and catch-all segments

tmdeveloper007:
## What
Adds tests for `useParams` covering optional route segments (`/users/:id?`) and catch-all segments (`/docs/*`).

## Why
`useParams` had no test coverage for these edge-case segment types.

## Testing
`bun vitest run packages/router/src/hooks/useParams.test.ts`

---


Karanjot786:

PR #3260, opened 2026-07-29
test(motion): add benchmark comparing spring vs linear vs easeInOut at 60fps

tmdeveloper007:
## What
Adds a benchmark in `packages/motion/bench/` that compares frame times for `spring`, `linear`, and `easeInOut` easing functions over 60 simulated frames.

## Why
No benchmark currently exists to validate the performance characteristics of the easing functions under render-pressure conditions.

## Testing
`bun run bench` or `bun packages/motion/bench/easing.ts`

---


Karanjot786:

PR #3261, opened 2026-07-29
test(core): add error-path unit tests for session module

tmdeveloper007:
## What
Adds error-path unit tests for the `packages/core/src/session/` module, covering SessionClosedError, reconnection logic, and edge-case state transitions.

## Why
The session module was missing test coverage for error paths. Unhandled SessionClosedError in production would silently break event subscriptions.

## Testing
`bun vitest run packages/core/src/session/session.test.ts`

---


Karanjot786:

PR #3262, opened 2026-07-29
test(data): add unit test for file watcher truncation recovery path

tmdeveloper007:
## What
Adds a unit test that simulates file truncation mid-stream for the data package watcher, covering the error path introduced in fix #2985.

## Why
The truncation recovery path has no test. Without it, future changes could silently break the fix.

## Testing
`bun vitest run packages/data/src/watcher.truncation.test.ts`

---


Karanjot786:

PR #3263, opened 2026-07-29
test(ui): add ARIA attribute tests for Select component

tmdeveloper007:
## What
Adds Vitest tests asserting correct ARIA attributes (aria-expanded, aria-haspopup, aria-selected) on the `Select` component in all interaction states.

## Why
Accessibility tests for Select were missing. Incorrect ARIA attributes can cause screen reader failures.

## Testing
`bun vitest run packages/ui/src/Select.aria.test.tsx`

---


Karanjot786:

PR #3264, opened 2026-07-29
test(quick): add test for render cleanup on terminal resize (SIGWINCH)

tmdeveloper007:
## What
Adds a test that simulates a SIGWINCH signal during `quick.render()` and verifies the view re-composes correctly without memory leaks or orphaned views.

## Why
No test covered the resize mid-render scenario. An improper cleanup could leave the terminal in a corrupted state.

## Testing
`bun vitest run packages/quick/src/resize-cleanup.test.ts`

---


Karanjot786:

PR #3308, opened 2026-07-30
fix : replaced unsafe non-null assertion on stack.pop() in testing queries

### LINKED ISSUE #3303: fix : replace unsafe non-null assertion on stack.pop() in testing queries

In `packages/testing/src/queries.ts`, the `walkWidgets` function uses a non-null assertion (`!`) on `stack.pop()` at line 15. While the loop condition checks `stack.length > 0`, TypeScript and ESLint cannot always infer this, making the code fragile.

Replace the non-null assertion with a safe guard:

```typescript
const widget = stack.pop();
if (widget === undefined) continue;
```

Eliminates a potential source of TypeErrors in test tooling. The code becomes more defensive and easier for static analysis tools to verify correctness.

tmdeveloper007:

Replaced the non-null assertion (`!`) on `stack.pop()` with a safe `undefined` guard in the `walkWidgets` function.

## Changes Made

- `packages/testing/src/queries.ts`: Changed `const widget = stack.pop()!;` to `const widget = stack.pop(); if (widget === undefined) continue;`.

## Impact it Made

Eliminates a potential source of TypeErrors in test tooling. The code is more defensive and easier for static analysis tools to verify correctness.

## Note: Please assign this PR to the `tmdeveloper007` account.

Closes #3303

### DIFF
--- a/packages/testing/src/queries.ts
+++ b/packages/testing/src/queries.ts
@@ -12,7 +12,8 @@ function walkWidgets(
   const stack: Widget[] = [root];
 
   while (stack.length > 0) {
-    const widget = stack.pop()!;
+    const widget = stack.pop();
+    if (widget === undefined) continue;
 
     if (predicate(widget)) {
       result.push(widget);


Karanjot786:

PR #3309, opened 2026-07-30
fix : added bounds check when extracting regex match groups in router route matching

### LINKED ISSUE #3304: fix : add bounds check when extracting regex match groups in router route matching

In `packages/router/src/route.ts`, the `matchNested` function iterates through `paramNames` to populate `params` from regex match groups without verifying that all expected groups were captured. Complex regex patterns could match but fail to capture all named parameters.

After populating params, verify all expected parameters were captured:

```typescript
for (let i = 0; i < paramNames.length; i++) {
    params[paramNames[i]] = match[i + 1] ?? '';
}
if (Object.keys(params).length !== paramNames.length) {
    return null;
}
```

Prevents routes from partially matching and returning incomplete parameter objects. A route that fails to capture all its named segments will correctly return `null` instead of a partially-populated params object, preventing downstream bugs.

tmdeveloper007:

Added a length validation after populating route params from regex match groups in `matchNested`. If not all expected parameters were captured, the function now returns `null` instead of returning a partially-populated params object.

## Changes Made

- `packages/router/src/route.ts`: Added `if (Object.keys(params).length !== paramNames.length) return null;` after param extraction.

## Impact it Made

Routes that match but fail to capture all named segments now correctly return `null` instead of a partially-populated params object, preventing downstream bugs in route handlers.

## Note: Please assign this PR to the `tmdeveloper007` account.

Closes #3304

### DIFF
--- a/packages/router/src/route.ts
+++ b/packages/router/src/route.ts
@@ -162,6 +162,9 @@ function matchNested(
             for (let i = 0; i < paramNames.length; i++) {
                 params[paramNames[i]] = match[i + 1] ?? '';
             }
+            if (Object.keys(params).length !== paramNames.length) {
+                return null;
+            }
             const nextChain = [...chain, route];
             return {
                 route,


Karanjot786:

PR #3386, opened 2026-08-03
fix(termui): address #3385 — services.list fallback

### LINKED ISSUE #3385: [bug] Service list omits unresolved names when systemd resolves another requested service
## Which package?

@termuijs/data

## What happened?

On Linux, `services.list()` returns only the services that `systemctl show` can resolve as soon as at least one lookup succeeds. Requested names that systemd cannot resolve are omitted instead of falling through to PM2 or process-name matching.

The API therefore returns fewer entries than were requested. A caller cannot tell whether an omitted service is stopped, managed by another supervisor, or simply absent.

## Steps to reproduce

1. Run on Linux with systemd available.
2. Call `services.list(['an-existing-systemd-unit', 'a-process-not-known-to-systemd'])`.
3. Observe that the result contains only the systemd unit when `systemctl show` succeeds for it.

```typescript
import { services } from '@termuijs/data';

const result = services.list(['ssh.service', 'my-node-worker']);
console.log(result.map(service => service.name));
// Actual:   ['ssh.service']
// Expected: entries for both requested names
```

`getSystemdServices()` skips failed names, and `list()` returns that partial array whenever it is non-empty. It should preserve the missing names for the PM2/process fallback before returning.

## Environment

- Bun version: 1.3.14
- OS: Linux (systemd available)
- Terminal emulator: not applicable
- TermUI version: current `main` at 6c7584eea5bf10285b3c5653019a3a5e1f6b5d23

## Screenshots or terminal output

Not applicable; this is a data-provider return-value issue.

## GSSoC contributor?

- [x] Yes. You contribute under GSSoC 2026.
- [ ] No.

tmdeveloper007:
## 📄 Description
Fix for [Karanjot786/TermUI #3385](https://github.com/Karanjot786/TermUI/issues/3385).

data/services.ts — `services.list()` now passes unresolved systemd names to PM2/process fallback instead of silently dropping them.

## 🔗 Related Issues
- Closes #3385

## 🧩 Type of Change
- [x] Bug fix

## ✅ Checklist
- [x] Code follows project style guidelines
- [x] Tests added/updated if applicable
- [x] Documentation updated if applicable
- [x] Changes don't introduce new warnings

### DIFF
--- a/create_prs.py
+++ b/create_prs.py
@@ -1,0 +1,524 @@
+#!/usr/bin/env python3
+"""
+TermUI PR creator - creates fix branches and PRs for upstream issues.
+Fork: tmdeveloper007/TermUI
+Upstream: Karanjot786/TermUI
+Token: GH_TOKEN env var
+"""
+
+import json
+import os
+import re
+import subprocess
+import sys
+import time
+import urllib.request
+import urllib.error
+from datetime import datetime, timezone
+
+GH_TOKEN = os.environ.get("GH_TOKEN", "")
+FORK_OWNER = "tmdeveloper007"
+UPSTREAM_OWNER = "Karanjot786"
+REPO = "TermUI"
+WORKSPACE = "/workspace/termui"
+
+def log(msg):
+    print(f"[{datetime.now(timezone.utc):%H:%M:%S}] {msg}", flush=True)
+
+def run(cmd, cwd=WORKSPACE, check=True, capture=True):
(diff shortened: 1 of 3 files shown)


Karanjot786:
Blocking — already fixed on main (services.list fallback, merged #3635) and the diff also dumps ~898 lines of unrelated bot automation (create_prs.py/orchestrate.py). Close as duplicate; never commit the automation scripts.

Karanjot786:

PR #3397, opened 2026-08-03
fix : added process.env fallback to keychain delete method

### LINKED ISSUE #3392: fix : add process.env fallback to keychain delete method

The `useKeychain` adapter in `packages/adapters/src/keychain/index.ts` has a `delete` method that does not fall back to checking `process.env` when the `keytar` module is unavailable. The `get` and `set` methods both call `getFallback(account)` which checks `process.env[accountToEnvName(account)]`, but `delete` only calls `fallbackStore.delete(account)`.

In the `delete` method, when `keytar` is null (either initially unavailable or set to null after a failure), the method should also check and remove the corresponding environment variable as a fallback.

This ensures consistent fallback behavior across `get`, `set`, and `delete` operations. Without this fix, accounts stored only in environment variables cannot be deleted when `keytar` is unavailable.

tmdeveloper007:

Added a `deleteFallback` helper function to `packages/adapters/src/keychain/index.ts` that removes the corresponding environment variable in addition to the in-memory fallback store when keytar is unavailable. The `delete` method now uses this helper in both the initial and runtime-failure fallback paths.

## Changes Made

- Added `deleteFallback(account: string): boolean` function that checks and removes `process.env[accountToEnvName(account)]` before calling `fallbackStore.delete(account)`.
- Updated `delete` method to call `deleteFallback` instead of `fallbackStore.delete` directly in both fallback paths.

## Impact it Made

The `delete` operation now has consistent fallback behavior with `get` and `set` — all three operations check environment variables when keytar is unavailable.

Closes #3392

### DIFF
--- a/packages/adapters/src/keychain/index.ts
+++ b/packages/adapters/src/keychain/index.ts
@@ -75,6 +75,15 @@ export function useKeychain(appName: string): UseKeychainResult {
     return fallbackStore.get(account) ?? process.env[accountToEnvName(account)] ?? null
   }
 
+  function deleteFallback(account: string): boolean {
+    const envName = accountToEnvName(account)
+    const envHadKey = envName in process.env
+    if (envHadKey) {
+      delete process.env[envName]
+    }
+    return fallbackStore.delete(account) || envHadKey
+  }
+
   return {
     async get(account: string): Promise<string | null> {
       if (!keytar) return getFallback(account)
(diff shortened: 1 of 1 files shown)


Karanjot786:
Blocking — breaks an existing test. Making keychain deleteFallback also delete the env var fails keychain/index.test.ts "falls back when keytar fails" (env var is a system-level override delete should not wipe). Either scope delete to keytar only, or update the documented behavior + test.

Karanjot786:

PR #3398, opened 2026-08-03
fix : added monotonicity validation to interpolate input range

### LINKED ISSUE #3393: fix : add monotonicity validation to interpolate input range

The `interpolate` function in `packages/motion/src/interpolate.ts` does not validate that the `inputRange` array is monotonically increasing. When `inputRange` contains non-ascending values (e.g., `[0, 2, 1, 3]`), the linear search produces incorrect interpolation results with no error or warning.

Add a validation check at the start of `interpolate` to verify that `inputRange` is strictly monotonically increasing. Throw a descriptive error if the array is not sorted.

Prevents silent incorrect behavior when users accidentally pass unsorted input ranges, making debugging much faster. It also documents the requirement that input ranges must be sorted.

tmdeveloper007:

Added a monotonicity check in `packages/motion/src/interpolate.ts` that validates `inputRange` is strictly ascending before performing interpolation. If any adjacent pair in `inputRange` violates strict ascending order, a descriptive error is thrown immediately.

Added corresponding test coverage in `packages/motion/src/interpolate.test.ts`.

## Changes Made

- Added a `for` loop after the existing length validation that checks `inputRange[i] < inputRange[i+1]` for all adjacent pairs.
- Throws `Error` with message: `"inputRange must contain strictly ascending values. Found inputRange[i]=X >= inputRange[i+1]=Y."`
- Added test case `'throws if inputRange is not strictly ascending'` with three sub-cases: strictly descending, equal adjacent values, and reversed pairs.

## Impact it Made

Prevents silent incorrect interpolation results when users accidentally pass unsorted input ranges. Makes the function fail fast with a clear error message.

Closes #3393

### DIFF
--- a/packages/motion/src/interpolate.test.ts
+++ b/packages/motion/src/interpolate.test.ts
@@ -60,4 +60,16 @@ describe('interpolate and mapRange', () => {
         expect(() => interpolate(5, [0, 1], [0, 1, 2])).toThrow();
         expect(() => interpolate(5, [0], [0])).toThrow();
     });
+
+    it('throws if inputRange is not strictly ascending', () => {
+        expect(() => interpolate(5, [0, 2, 1, 3], [0, 20, 10, 30])).toThrow(
+            'inputRange must contain strictly ascending values'
+        );
+        expect(() => interpolate(5, [0, 1, 1, 2], [0, 10, 20, 30])).toThrow(
+            'inputRange must contain strictly ascending values'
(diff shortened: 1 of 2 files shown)


Karanjot786:
Closing this one. This fix is already on `main` from a previously merged PR (verified against the current source), so the change is a no-op / conflicts. Thanks anyway — the bug is fixed.

PR #3410, opened 2026-08-03
fix: emit re-entrancy safety with hasSkippedHandlers

### LINKED ISSUE #3395: fix : EventEmitter emit silently skips regular handlers on re-entrant calls

The `emit` method in `packages/core/src/events/EventEmitter.ts` has a re-entrancy guard that skips regular handlers when `emit` is called re-entrantly (while another handler is executing). The `_emitting` flag correctly prevents infinite recursion, but regular handlers are silently skipped while `once` handlers in the snapshot still run. Calling code has no way to detect that regular handlers were not called.

Add a `_skipped` flag that tracks whether handlers were skipped during the current emit cycle. Expose this state through a new `hasSkippedHandlers(event)` method so subscribers can detect when their handlers were bypassed. Alternatively, consider whether skipping regular handlers on re-entrant emit is the desired behavior at all.

Makes the EventEmitter behavior more transparent and debuggable, especially in complex event chains where re-entrancy can occur.

tmdeveloper007:
## Description

Add re-entrancy safety to EventEmitter.emit() with depth-tracking.

- Tracks emit depth via a `_emitting` Map (1=outermost, 2+=re-entrant)
- Re-entrant emits skip regular handlers and set a `_skipped` flag
- New `hasSkippedHandlers(event)` method reports whether handlers were skipped due to re-entrancy
- All 22 EventEmitter tests pass

## Related Issues

Closes #3395

## Type of Change

## Checklist

- [x] Tests pass (22/22 EventEmitter tests)
- [x] TypeScript compiles without errors

### DIFF
--- a/packages/core/src/events/EventEmitter.test.ts
+++ b/packages/core/src/events/EventEmitter.test.ts
@@ -276,4 +276,34 @@ describe('EventEmitter', () => {
         expect(handlerB).toHaveBeenCalledTimes(1); // Not called again
         expect(handlerC).toHaveBeenCalledTimes(2);
     });
+
+    it('hasSkippedHandlers returns true when regular handlers are skipped due to re-entrant emit', () => {
+        const emitter = new EventEmitter<TestEvents>();
+        const outerHandler = vi.fn();
+        const innerHandler = vi.fn();
+
+        emitter.on('message', outerHandler);
+        emitter.on('message', () => {
+            innerHandler();
+            // Re-entrant emit on same event
+            emitter.emit('message', 'reentrant');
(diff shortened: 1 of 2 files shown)


Karanjot786:
Blocking — does not fix #2888. hasSkippedHandlers() only *reports* that re-entrant regular handlers were skipped; it never queues/replays them, so they are still silently dropped (your own test shows outerHandler fires once). Queue skipped handlers and drain them after the current emit, and add a test proving delivery. Also rebase (conflicts).

Karanjot786:

PR #3411, opened 2026-08-03
fix: re-register subscribeOnce wrapper when listener throws

### LINKED ISSUE #3396: fix : subscribeOnce listener removal is unconditional even when listener throws

The `subscribeOnce` implementation in `packages/store/src/store.ts` removes the listener wrapper from the listeners Set before invoking it. If the listener throws an error, the listener is still removed (as intended for "once" semantics), but the error is then re-thrown from the `setState` path after the listener has already been removed. This means a caller cannot recover and re-subscribe in a catch block with the same listener reference.

Consider whether the listener should be removed before or after invocation when an error occurs. A pragmatic fix is to remove the listener only after successful invocation, or document this behavior clearly. Alternatively, track failed once-listeners separately and remove them only after successful execution.

Improves error handling ergonomics for `subscribeOnce`. Currently, if a `subscribeOnce` listener throws, the caller cannot re-use the same listener reference since it has already been removed.

tmdeveloper007:
## Description

Previously the `subscribeOnce` wrapper was removed from the listener set BEFORE calling the user-provided listener. If the listener threw an exception, the subscription was already gone and future state changes would not notify it.

Now the wrapper re-registers itself after a listener throw, so future `setState` calls can still retry the listener.

## Related Issues

Closes #3396

## Type of Change

## Checklist

- [x] Tests pass (55/55 store tests)
- [x] TypeScript compiles without errors

## Summary by CodeRabbit

- **Bug Fixes**
  - One-time state listeners now remain subscribed if their callback throws an error.
  - Errors from listener callbacks continue to propagate to the caller, while later state updates can notify the listener again.

### DIFF
--- a/packages/core/src/events/EventEmitter.test.ts
+++ b/packages/core/src/events/EventEmitter.test.ts
@@ -276,4 +276,34 @@ describe('EventEmitter', () => {
         expect(handlerB).toHaveBeenCalledTimes(1); // Not called again
         expect(handlerC).toHaveBeenCalledTimes(2);
     });
+
+    it('hasSkippedHandlers returns true when regular handlers are skipped due to re-entrant emit', () => {
+        const emitter = new EventEmitter<TestEvents>();
+        const outerHandler = vi.fn();
+        const innerHandler = vi.fn();
+
+        emitter.on('message', outerHandler);
+        emitter.on('message', () => {
+            innerHandler();
+            // Re-entrant emit on same event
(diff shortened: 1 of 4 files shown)


Karanjot786:
Merged current main and dropped the unrelated EventEmitter.ts / EventEmitter.test.ts changes (they belong to #3410, not this fix). The PR now touches only packages/store. No action needed from you.

PR #3416, opened 2026-08-03
Fix: password visibility toggle doesn't work due to disconnected reff

### LINKED ISSUE #3415: bug : Password visibility toggle is non-functional due to disconnected passwordRef
## Which package?

`@termuijs/ui`

Affected file: examples/auth-flow/src/index.tsx

---

## What happened?

While reviewing `examples/auth-flow/src/index.tsx`, I noticed that the password visibility toggle appears to be unreachable.

`LoginScreen` creates a `passwordRef` and uses it to call `toggleVisibility()` and read `showText`. However, `PasswordInputJSX` creates its own internal `useRef` for the `PasswordInput` instance and never exposes it to the parent component.

As a result, `passwordRef.current` in `LoginScreen` is never assigned and remains `null`. This means calls such as:

```ts
passwordRef.current?.toggleVisibility();
```

never reach the actual `PasswordInput` instance, and:

```ts
const isVisible = passwordRef.current?.showText ?? false;
```

always evaluates to `false`.

The parent component should receive a reference to the `PasswordInput` instance (e.g., via a forwarded ref or a `widgetRef` prop) so it can control the widget as intended.

---

## GSSoC contributor?

* [] Yes. You contribute under GSSoC 2026.
* [x] No.

Alexx3890:
## Description

Fixes the password visibility toggle in the `examples/auth-flow` login example by connecting the parent `passwordRef` to the `PasswordInput` instance and keeping the visibility label synchronized with the widget state. It also removes duplicated toggle and submit logic by introducing shared helper functions.

## Related Issue

Closes #3415

## Which package(s)?

`examples/auth-flow`

## Type of Change

* [x] 🐛 Bug fix (`type:bug`)
* [ ] ✨ Feature (`type:feature`)
* [ ] 📝 Docs (`type:docs`)
* [ ] 🧪 Tests (`type:testing`)
* [ ] ♻️ Refactor (`type:refactor`)
* [ ] 🎨 Design / UX (`type:design`)
* [ ] ♿ Accessibility (`type:accessibility`)
* [ ] ⚡ Performance (`type:performance`)
* [ ] 🔧 DevOps / CI (`type:devops`)
* [ ] 🔒 Security (`type:security`)

## Checklist

* [x] ⭐ You starred the repo. The `needs-star` check blocks your merge otherwise.
* [ ] Tests pass locally: `bun vitest run`
* [ ] Build passes: `bun run build`
* [ ] Typecheck passes: `bun run typecheck`
* [x] You read [`[CONTRIBUTING.md](https://chatgpt.com/c/CONTRIBUTING.md)`](./CONTRIBUTING.md).
* [x] Your PR title follows `type: short description`.
* [x] Widget state mutators call `markDirty()` (if your change affects rendering).
* [x] No new `any` types without an inline comment explaining why.
* [x] No unrelated refactors bundled into this PR.

> **Note:** Only check the test/build/typecheck boxes if you've actually run those commands successfully.

## GSSoC 2026 Participation

* [x] No

N/A 

## Notes for the Reviewer

This PR fixes the password visibility toggle in the `examples/auth-flow` example by ensuring the parent component and `PasswordInputJSX` share the same `PasswordInput` instance. It also synchronizes the visibility label with the widget state and consolidates duplicated toggle and submit logic for easier maintenance.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved password field behavior and focus handling during authentication.
  - Ensured Enter/Return consistently submits login forms.
  - Improved password visibility toggling across supported interactions.
  - Preserved reliable logout behavior on protected screens.
  - Improved consistency when entering, submitting, and managing login credentials.

- **Refactor**
  - Simplified and standardized authentication screen behavior for a more consistent experience.

### DIFF
--- a/examples/auth-flow/src/index.tsx
+++ b/examples/auth-flow/src/index.tsx
@@ -21,7 +21,7 @@ function TextInputJSX({
     onChange: (val: string) => void;
     placeholder?: string;
     isFocused: boolean;
-}) {
+}): TextInput {
     const ref = useRef<TextInput | null>(null);
 
     if (!ref.current) {
@@ -57,13 +57,19 @@ function TextInputJSX({
                 ref.current.moveCursorEnd();
                 break;
             default:
-                if (key.length === 1 && !event.ctrl && !event.alt) {
+                // Defensive `key &&` guard: some non-printable keystrokes may
+                // reach this handler with an empty/undefined key depending on
+                // how the TermUI input dispatcher normalizes them.
+                if (key && key.length === 1 && !event.ctrl && !event.alt) {
                     ref.current.insertChar(key);
                 }
         }
     });
 
-    return ref.current as any;
+    // TermUI's JSX runtime accepts widget instances directly as render
+    // output (this isn't a standard JSX.Element), so returning the concrete
+    // TextInput type is intentional rather than a type-safety gap.
+    return ref.current;
 }
 
 function PasswordInputJSX({
@@ -78,7 +84,7 @@ function PasswordInputJSX({
     placeholder?: string;
     isFocused: boolean;
     widgetRef: { current: PasswordInput | null };
-}) {
+}): PasswordInput {
     if (!widgetRef.current) {
         widgetRef.current = new PasswordInput(
             { width: 30 },
@@ -97,7 +103,9 @@ function PasswordInputJSX({
         widgetRef.current.handleKey(event);
     });
 
-    return widgetRef.current as any;
+    // Same TermUI-specific widget-instance-as-render-output pattern as
+    // TextInputJSX above.
+    return widgetRef.current;
 }
 
 function LoginScreen() {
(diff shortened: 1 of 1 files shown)


Alexx3890 (commented):
Addressed the CodeRabbit suggestion — removed `as any` from both
`TextInputJSX` and `PasswordInputJSX`, returning the concrete
`TextInput` / `PasswordInput` types instead, since `ref.current` /
`widgetRef.current` are already narrowed to non-null at that point.
Added a short inline comment explaining why these return a widget
instance rather than a standard `JSX.Element` (TermUI's JSX runtime
accepts widget instances directly as render output).

Also restored the `key &&` guard in `TextInputJSX`'s default case
that had been dropped in an earlier revision — without it, a
falsy/empty `key` for a non-printable keystroke could throw on
`key.length`.

One open item: I haven't run `pnpm typecheck` against this yet to
confirm TermUI's JSX pragma doesn't require these to satisfy
`JSX.Element` at the call site. If it does, we'll need a narrower
cast (e.g. `as unknown as JSX.Element`) instead of a bare return —
will confirm before merge.

PR #3435, opened 2026-08-03
fix : added ASCII fallback for Code widget gutter pipe character

### LINKED ISSUE #3430: fix : Code widget renders Unicode pipe character without ASCII fallback

The `Code` widget (`packages/widgets/src/display/Code.ts`) renders a vertical pipe character (`│`) as the separator between line numbers and code content without checking `caps.unicode`. When the terminal does not support Unicode/box-drawing characters, this renders incorrectly.

1. In `packages/widgets/src/display/Code.ts`, add a `caps.unicode` check before rendering the vertical separator character in `_renderSelf()`.
2. Use `caps.unicode ? '│' : '|'` for the separator character.
3. Add tests that verify the ASCII fallback renders correctly when `caps.unicode` is false.

The Code widget will render correctly in terminals that do not support Unicode box-drawing characters (e.g., some legacy terminals, certain CI environments, or terminals with `TERM=dumb`).

tmdeveloper007:
## Description

Fixes a bug in the Code widget where the gutter separator always renders the Unicode box-drawing character `│` (`U+2502`), even when the terminal does not support Unicode. This causes display issues on ASCII-only terminals.

The fix adds a `caps.unicode` check and falls back to the ASCII pipe character `|` when Unicode is disabled.

## Changes

- `packages/widgets/src/display/Code.ts` — replaced hardcoded `│` with `caps.unicode ? '│' : '|'`
- `packages/widgets/src/display/Code.test.ts` — added test verifying ASCII pipe fallback

## Related Issues

- Fixes #3430

## Type of Change

## Checklist

- [x] Code follows the existing style
- [x] Tests added or updated
- [x] Documentation updated (if applicable)
- [x] All tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated code line-number gutters to use a compatible separator based on Unicode support.
  * Added an ASCII `|` fallback for environments that cannot render the Unicode separator.

### DIFF
--- a/packages/widgets/src/display/Code.test.ts
+++ b/packages/widgets/src/display/Code.test.ts
@@ -2,8 +2,8 @@
 // @termuijs/widgets — Tests for Code widget
 // ─────────────────────────────────────────────────────
 
-import { describe, it, expect } from 'vitest';
-import { Screen } from '@termuijs/core';
+import { describe, it, expect, vi } from 'vitest';
+import { Screen, caps } from '@termuijs/core';
 import { Code } from './Code.js';
 
 describe('Code', () => {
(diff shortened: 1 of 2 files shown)


Karanjot786:
Merged current main into this branch to resolve the test-file conflict and restored the caps.unicode mock after the new test so it can't leak into later tests. No action needed from you.

PR #3436, opened 2026-08-03
fix : added ASCII fallback for Markdown blockquote pipe character

### LINKED ISSUE #3431: fix : Markdown widget quote block renders Unicode pipe without ASCII fallback

The `Markdown` widget (`packages/widgets/src/display/Markdown.ts`) renders quote block prefix characters using a hardcoded Unicode pipe (`│`) instead of respecting the `caps.unicode` capability flag. When the terminal lacks Unicode support, the quote block renders incorrectly.

1. In `packages/widgets/src/display/Markdown.ts`, update the quote block rendering in `_renderSelf()` to use `caps.unicode ? '│' : '|'` for the quote prefix character.
2. Add tests that verify ASCII fallback for quote blocks works correctly when `caps.unicode` is false.

Markdown quote blocks (lines starting with `>`) will render correctly in non-Unicode terminals, improving cross-terminal compatibility.

tmdeveloper007:
## Description

Fixes a bug in the Markdown widget where blockquote lines always render with a Unicode box-drawing pipe character `│` (`U+2502`), even when the terminal does not support Unicode. This causes display issues on ASCII-only terminals.

The fix adds a `caps.unicode` check and falls back to the ASCII pipe character `|` when Unicode is disabled.

## Changes

- `packages/widgets/src/display/Markdown.ts` — replaced hardcoded `│` with `${caps.unicode ? '│' : '|'}`
- `packages/widgets/src/display/Markdown.test.ts` — added test verifying ASCII pipe fallback

## Related Issues

- Fixes #3431

## Type of Change

## Checklist

- [x] Code follows the existing style
- [x] Tests added or updated
- [x] Documentation updated (if applicable)
- [x] All tests pass locally

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved code and Markdown rendering in environments without Unicode support.
  * Gutter separators and blockquote borders now use the ASCII `|` character when needed, while retaining the Unicode `│` character where supported.
* **Tests**
  * Added coverage to verify correct separator rendering across supported display modes.

### DIFF
--- a/packages/widgets/src/display/Code.test.ts
+++ b/packages/widgets/src/display/Code.test.ts
@@ -2,8 +2,8 @@
 // @termuijs/widgets — Tests for Code widget
 // ─────────────────────────────────────────────────────
 
-import { describe, it, expect } from 'vitest';
-import { Screen } from '@termuijs/core';
+import { describe, it, expect, vi } from 'vitest';
+import { Screen, caps } from '@termuijs/core';
 import { Code } from './Code.js';
 
 describe('Code', () => {
@@ -72,5 +72,14 @@ describe('Code', () => {
         expect(screen.back[0][2].char).toBe('[');
         expect(screen.back[0][3].char).toBe('t');
         expect(screen.back[0][4].char).toBe('╮'); // topRight corner (width-1 index)
+    });
+
+    it('uses ASCII pipe for gutter separator when caps.unicode is false', () => {
+        vi.spyOn(caps, 'unicode', 'get').mockReturnValue(false);
+        const code = new Code('hello');
+        code.updateRect({ x: 0, y: 0, width: 12, height: 4 });
+        const screen = new Screen(12, 4);
+        code.render(screen);
+        expect(screen.back[1][2].char).toBe('|');
     });
 });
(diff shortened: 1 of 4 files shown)


Karanjot786:
Thanks for the contribution. The widget change itself is fine, but there are issues to fix before this merges.

Issues:

1. Bundled Code.ts commit is out of scope
   Files: packages/widgets/src/display/Code.ts, packages/widgets/src/display/Code.test.ts
   Problem: this PR carries the same Code widget commit as #3435. That change is being handled in #3435; it doesn't belong here.
   Fix: after #3435 lands, merge current `main` into this branch and confirm `git diff main -- packages/widgets/src/display/Code.ts` is empty. Until then, drop the commit from this branch.

2. Branch conflicts with main
   Fix: merge current `main` into your branch and resolve.

3. CI never ran the build
   Problem: `build-and-test` has not executed on this branch, so the change is unverified.
   Fix: pushing after the merge will trigger it; make sure it goes green.

4. Title format
   Fix: `fix(widgets): add ASCII fallback for ...` (no space before the colon).

- [ ] Code.ts / Code.test.ts no longer differ from main
- [ ] Merged main, no conflicts, build-and-test green

Issue #3746, opened 2026-08-15
[feature] Add useClipboard hook for copying text to system clipboard


pixeltannu:
## What problem does this solve?

TermUI apps often need to let users copy output (logs, generated code, 
config values, IDs, etc.) to the system clipboard. Right now there's no 
built-in way to do this — developers have to manually shell out to 
platform-specific commands (`pbcopy`, `xclip`/`xsel`, `clip.exe`) inside 
their own components, which is repetitive and error-prone across OSes.

## Proposed solution

Add a `useClipboard` hook to `@termuijs/jsx` (or a new `@termuijs/clipboard` 
utility) that exposes a simple copy/read API and handles OS detection 
internally.

```typescript
import { useClipboard } from '@termuijs/jsx'

function CopyableOutput({ text }: { text: string }) {
  const { copy, copied } = useClipboard()

  useKeymap([
    { key: 'c', ctrl: true, action: () => copy(text) },
  ])

  return (
    <Box border="round" padding={1}>
      <Text>{text}</Text>
      <Text dim>{copied ? '✓ Copied!' : 'Press ctrl+c to copy'}</Text>
    </Box>
  )
}
```

- `copy(text: string): Promise<void>` — writes text to system clipboard
- `read(): Promise<string>` — reads current clipboard content (optional, phase 2)
- `copied: boolean` — transient state, auto-resets after ~2s, useful for UI feedback
- Should detect platform (macOS/Linux/Windows) and gracefully no-op with a warning if no clipboard utility is found (e.g., headless CI environment)

## Alternatives considered

- Leaving it to userland (current state) — rejected, since it forces every 
  app to reimplement OS-specific clipboard logic
- Depending on an external npm package like `clipboardy` — viable, but 
  adds a runtime dependency; worth discussing whether TermUI wants to stay 
  fully dependency-free per its "No native dependencies" philosophy

PR #3753, opened 2026-08-15
fix(core): add tmux/screen passthrough and timeout guard to clipboard functions

pixeltannu:
## What does this fix?

While working on #3746, I found `useClipboard`/`writeClipboard`/`readClipboard` 
already implemented and tested on `main`. This PR extends the existing OSC 52 
clipboard implementation with two fixes for gaps I found while verifying it:

1. **tmux/GNU screen passthrough**: OSC 52 sequences are silently dropped by 
   tmux and GNU screen unless wrapped for passthrough. `writeClipboard` and 
   `readClipboard` now detect `TMUX`/`STY` env vars and wrap the sequence 
   accordingly, so clipboard copy/read works correctly inside multiplexers.

2. **`readClipboard` timeout**: previously, if the terminal didn't respond to 
   the OSC 52 query (e.g. terminal doesn't support it), the returned Promise 
   would hang indefinitely and the `stdin` listener would never be cleaned up. 
   `readClipboard` now accepts an optional `timeoutMs` (default 1000ms) and 
   rejects with a clear error if no response arrives in time.

## Testing

- All existing `ansi.test.ts` tests pass (29/29)
- Full test suite run: 51/51 passed, no regressions
- Manually verified `writeClipboard`/`readClipboard` signatures remain 
  backward-compatible (new `options` param is optional)

## Related

Relates to #3746

## Summary by CodeRabbit

* **New Features**
  * Added clipboard support when running inside tmux and GNU screen.
  * Added configurable clipboard read timeouts.
* **Bug Fixes**
  * Clipboard requests now handle terminal multiplexer escaping correctly.
  * Clipboard reads clean up resources and fail promptly when no response is received.

### DIFF
--- a/packages/core/src/utils/ansi.ts
+++ b/packages/core/src/utils/ansi.ts
@@ -166,44 +166,92 @@ export function stripAnsiControl(str: string): string {
     return out;
 }
 
-// ── Clipboard ───────────────────────────────────────
+// ── Clipboard ──────────────────────────────────────────
+
+const TMUX_DCS_START = '\x1bPtmux;';
+const TMUX_DCS_END = '\x1b\\';
+const SCREEN_DCS_START = '\x1bP';
+const SCREEN_DCS_END = '\x1b\\';
+
+/**
+ * Wraps an OSC escape sequence for passthrough through tmux or GNU screen.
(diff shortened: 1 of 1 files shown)


pixeltannu:
@Karanjot786 kindly mearge my pr

PR #3765, opened 2026-08-16
Fix data hooks and adapter edge cases

### LINKED ISSUE #3755: bug(data): useMutation treats 204/empty successful responses as failures
### Summary

useMutation always calls 
esponse.json() after any 
esponse.ok result in packages/data/src/hooks/useMutation.ts. Successful mutation endpoints often return 204 No Content or another empty body, so the hook throws a JSON parse error even though the HTTP request succeeded.

### Reproduction

1. Use useMutation('/resource/1', 'DELETE') against an endpoint that returns 204 No Content.
2. Call mutate(undefined).
3. The fetch succeeds, but wait response.json() rejects because there is no body.

### Expected behavior

A successful empty response should be treated as a successful mutation. The hook could return 
ull/undefined, avoid setting an error, and still increment mutationCount.

### Actual behavior

The mutation is reported as failed with a parse error, even though the server returned a successful status.

### LINKED ISSUE #3756: bug(data): concurrent useMutation calls can overwrite newer results with stale data
### Summary

useMutation in packages/data/src/hooks/useMutation.ts does not track request order. If mutate() is called twice and the first request resolves after the second, the older response can overwrite the newer data state.

### Reproduction

1. Create a component using useMutation('/save', 'POST').
2. Call mutate({ value: 'old' }), then quickly call mutate({ value: 'new' }).
3. Make the 
ew request resolve first and the old request resolve last.

### Expected behavior

The hook should either document that concurrent mutations are unsupported or guard state updates so the latest mutation result remains visible.

### Actual behavior

The older response can call setData(result) after the newer response and leave the UI showing stale mutation data.

### LINKED ISSUE #3757: bug(data): useSSE keeps using a stale parser after parse changes
### Summary

useSSE in packages/data/src/hooks/useSSE.ts accepts a parse callback, but the effect dependency list only includes [url]. If a component re-renders with the same URL and a new parser, incoming events are still parsed by the old callback.

### Reproduction

1. Render useSSE(url, parseA).
2. Re-render the same component with the same url and parseB.
3. Emit another message event from the same EventSource.

### Expected behavior

The hook should use the latest parser, either by including parse in the dependency list or by storing the latest callback in a ref.

### Actual behavior

The event listener closes over the parser from the original effect run, so future messages can be transformed incorrectly.

### LINKED ISSUE #3758: bug(data): useSSE does not reset state when subscribing to a new URL
### Summary

When the url argument changes, useSSE in packages/data/src/hooks/useSSE.ts opens a new EventSource, but it does not reset data, error, or loading before the new stream receives its first message.

### Reproduction

1. Subscribe to stream A and receive a message.
2. Re-render with stream B.
3. Before stream B emits its first message, read the hook state.

### Expected behavior

Changing url should put the hook back into a fresh loading state and avoid showing data from the previous stream as if it came from the new stream.

### Actual behavior

The previous stream's last data value remains visible and loading can remain alse until the new stream emits or errors.

### LINKED ISSUE #3759: bug(data): useWebSocket keeps the previous message after URL changes
### Summary

useWebSocket in packages/data/src/hooks.ts reconnects when url changes, but it never clears the message state. Consumers can see the previous socket's last message while the new socket is still connecting.

### Reproduction

1. Render useWebSocket('wss://old.example').
2. Receive a message such as old-data.
3. Re-render with useWebSocket('wss://new.example').
4. Inspect message before the new socket receives anything.

### Expected behavior

A URL change should reset message to 
ull or otherwise make it clear that the old message belongs to the old connection.

### Actual behavior

The hook keeps returning the old message while state is already for the new connection lifecycle.

### LINKED ISSUE #3760: bug(data): http.ping omits failed requests from latency history
### Summary

http.ping in packages/data/src/http.ts records latency history only after etch() succeeds. If a request times out, is aborted, or fails DNS/TCP, the returned HealthResult includes a latency value, but http.latency(url) never records it.

### Reproduction

1. Call wait http.ping('http://127.0.0.1:1') or another endpoint that fails quickly.
2. Observe that the returned result has status: 'down' and a measured latency.
3. Call http.latency(url).

### Expected behavior

Latency history should include failed checks as well, since the health result exposes a latency and dashboards/sparklines need outage timing too.

### Actual behavior

Only successful fetch attempts are pushed into _latencyHistory, so down endpoints appear to have no history.

### LINKED ISSUE #3761: bug(data): usePolling performs a fetch while paused after deps or interval change
### Summary

usePolling in packages/data/src/hooks/usePolling.ts supports pause(), but the effect always calls execute() immediately when interval, options.adaptive, or a dependency changes. That immediate execution does not check pausedRef.

### Reproduction

1. Render usePolling(fn, 1000, [dep]).
2. Call pause().
3. Re-render with a changed dep or interval.
4. Observe that n is invoked even though the hook is paused.

### Expected behavior

While paused, dependency changes should update the polling configuration without starting a new request until 
esume() or an explicit 
efresh() call.

### Actual behavior

The effect rerun calls execute() unconditionally, so a paused poller can still fire network or system calls.

### LINKED ISSUE #3762: bug(data): useFileWatch keeps stale event state when path changes
### Summary

useFileWatch in packages/data/src/hooks/useFileWatch.ts creates a new watcher when path or options change, but it does not reset data, error, or loading for the new target.

### Reproduction

1. Watch path A and trigger a change so data is populated.
2. Re-render the hook with path B.
3. Inspect the hook state before path B emits any event.

### Expected behavior

A new watched path should start with data: null, error: null, and loading: true until the new watcher receives an event or fails.

### Actual behavior

The last event from path A remains visible while the hook is already watching path B, which can make file monitors report changes for the wrong target.

### LINKED ISSUE #3763: bug(adapters): useLocalStorage service name can escape the TermUI config directory
### Summary

useLocalStorage in packages/adapters/src/localStorage/index.ts builds the storage file by joining the TermUI config directory with the caller-provided service name plus a .json suffix. The service value is not validated, so a value containing ../ path segments can resolve outside the intended TermUI config directory.

### Reproduction

1. Call useLocalStorage('../outside-termui').
2. Call set('key', 'value').
3. The resolved path is no longer constrained to ~/.config/termui.

### Expected behavior

The adapter should validate service names or resolve and verify that the final path remains inside the TermUI storage directory.

### Actual behavior

Untrusted or accidental service names can cause reads/writes outside the adapter's intended storage location.

### LINKED ISSUE #3764: bug(adapters): useExeca requires execa at module load despite optional peer handling
### Summary

packages/adapters/src/execa/index.ts has a top-level value import: import { execa } from 'execa'. The rest of the file tries to lazy-load execa and throw a friendly optional-peer error, but the static import means the module can fail before useExeca() is even called.

### Reproduction

1. Install/use @termuijs/adapters without installing the optional execa peer dependency.
2. Import the adapters package or the execa adapter module.
3. The runtime attempts to resolve execa during module evaluation.

### Expected behavior

The package should be importable without optional peer dependencies installed. The helpful useExeca() error should only appear when the execa adapter is actually used.

### Actual behavior

The top-level value import can make the module fail to load before the lazy optional-peer handling runs. The file can use a type-only import for the execa function shape and rely on the existing dynamic import path.

Tomeshwari-02:
## Summary
- fix empty/204 mutation responses and stale concurrent mutation commits
- reset stale hook state for SSE, file watch, and WebSocket target changes
- avoid paused polling fetches after dependency changes and record failed ping latency
- keep optional execa imports lazy and validate localStorage service names

## Issues
Closes #3755
Closes #3756
Closes #3757
Closes #3758
Closes #3759
Closes #3760
Closes #3761
Closes #3762
Closes #3763
Closes #3764

## Tests
- bun vitest run packages/data/src/hooks/useMutation.test.ts packages/data/src/hooks/usePolling.test.ts packages/data/src/hooks/useSSE.test.ts packages/data/src/hooks/useFileWatch.test.ts packages/data/src/useWebSocket.test.tsx packages/data/src/http.test.ts packages/adapters/src/localStorage/index.test.ts
- bun run build
- bun run typecheck

## Summary by CodeRabbit

* **New Features**
  * Mutation requests now support empty and `204`/`205` responses, returning `null` when appropriate.
  * Polling can be manually refreshed while paused.
* **Bug Fixes**
  * Prevented stale results from overwriting newer concurrent mutation results.
  * Reset stale loading, error, and data states when connections, subscriptions, watchers, or URLs change.
  * Paused polling no longer triggers unexpected requests after dependency updates.
  * Invalid storage names can no longer escape the configured storage directory.
  * Failed network checks now record zero latency consistently.

### DIFF
--- a/packages/adapters/src/execa/index.ts
+++ b/packages/adapters/src/execa/index.ts
@@ -3,7 +3,8 @@
 // ─────────────────────────────────────────────────────
 
 import type { Options } from 'execa';
-import { execa } from 'execa';
+
+type ExecaFunction = (file: string, args?: string[], options?: Options) => any;
 
 export interface UseExecaResult {
   run(
@@ -32,15 +33,15 @@ async function getExeca(): Promise<typeof import('execa')> {
   }
 }
 
-function isExecaCallable(obj: unknown): obj is typeof execa {
+function isExecaCallable(obj: unknown): obj is ExecaFunction {
   return typeof obj === 'function';
 }
 
-function isExecaModuleWithExeca(obj: unknown): obj is { execa: typeof execa } {
+function isExecaModuleWithExeca(obj: unknown): obj is { execa: ExecaFunction } {
   return typeof obj === 'object' && obj !== null && 'execa' in obj && typeof (obj as { execa: unknown }).execa === 'function';
 }
 
-function isExecaModuleWithDefault(obj: unknown): obj is { default: typeof execa } {
+function isExecaModuleWithDefault(obj: unknown): obj is { default: ExecaFunction } {
   return typeof obj === 'object' && obj !== null && 'default' in obj && typeof (obj as { default: unknown }).default === 'function';
 }
 
@@ -60,7 +61,7 @@ export function useExeca(globalOpts?: Options): UseExecaResult {
       opts?: Options
     ): AsyncGenerator<string, void, unknown> {
       const execaModule: unknown = await getExeca();
-      let execaFn: typeof execa;
+      let execaFn: ExecaFunction;
 
       if (isExecaModuleWithExeca(execaModule)) {
         execaFn = execaModule.execa;
--- a/packages/adapters/src/localStorage/index.test.ts
+++ b/packages/adapters/src/localStorage/index.test.ts
@@ -35,4 +35,7 @@ describe('useLocalStorage', () => {
     store.set('a', 1); store.set('b', 2); store.clear()
     expect(store.get('a')).toBeNull(); expect(store.get('b')).toBeNull()
   })
+  it('rejects service names that can escape the config directory', () => {
+    expect(() => useLocalStorage('../outside-termui')).toThrow(/service must contain/)
+  })
 })
--- a/packages/adapters/src/localStorage/index.ts
+++ b/packages/adapters/src/localStorage/index.ts
@@ -9,7 +9,14 @@ export interface LocalStorageAdapter {
   clear(): void
 }
 
+function assertValidServiceName(service: string): void {
+  if (!/^[A-Za-z0-9][A-Za-z0-9._-]*$/.test(service)) {
+    throw new Error('useLocalStorage() service must contain only letters, numbers, dots, underscores, and hyphens, and must start with a letter or number.')
+  }
+}
+
 function resolveStorePath(service: string): string {
+  assertValidServiceName(service)
   const dir = path.join(os.homedir(), '.config', 'termui')
   if (!fs.existsSync(dir)) fs.mkdirSync(dir, { recursive: true })
   return path.join(dir, `${service}.json`)
--- a/packages/data/src/hooks.ts
+++ b/packages/data/src/hooks.ts
@@ -265,6 +265,7 @@ export function useWebSocket(url: string): UseWebSocketReturn {
         let isMounted = true;
         const thisGeneration = ++generationRef.current;
         retryCountRef.current = 0;
+        setMessage(null);
 
         function connect() {
             if (socketRef.current) {
--- a/packages/data/src/hooks/useFileWatch.test.ts
+++ b/packages/data/src/hooks/useFileWatch.test.ts
@@ -81,6 +81,21 @@ describe('useFileWatch', () => {
         expect(stateValues[2]).toBe(false);
     });
 
+    it('resets stale state when starting a new watcher', () => {
+        stateValues = [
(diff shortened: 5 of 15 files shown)


Karanjot786:
Mostly solid (useMutation race guard, 204 handling, localStorage path-traversal block). One blocker: usePolling.ts:139 adds paused to the effect deps, so pause/resume recreates the AdaptivePollingController (line 101), wiping accumulated backoff state each toggle — contradicts the docstring. Gate the initial execute() on pausedRef.current instead of changing deps.

